Issue #19, opened 2024-07-11
Update project layouts from go/v4 to traditional `pkg` way


Jeffwan:
Currently, `kubebuilder` uses `go/v4` https://github.com/kubernetes-sigs/kubebuilder/issues/932#issuecomment-1368208867. We should still follow golang pkg fashion and let me change the layout after https://github.com/aibrix/aibrix/pull/17 is merged

Issue #16, opened 2024-07-11
Create controller skeleton for autoscaler


Jeffwan:
Since @kr11 will pick up the autoscaler implementation work, let's kick off the controller skeleton. 

There're different ways to organize the codes and different controller implementations. We will still adopt `kubebuilder` style for short term. Once we have more controllers, we will extract the common parts and build the controller framework.

1. Figure out the kubebuilder/go/kubernetes version
2. Create the skeleton and document the steps
3. Update the development guidance with necessary steps to build the project

PR #21, opened 2024-07-12
Optimize project layouts by moving controllers to pkg folder

Jeffwan:
Address #16 #19 

We should exactly follow kubebuilder's layout. Moving `internal/controllers` under `pkg`

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -19,7 +19,6 @@ package main
 import (
 	"crypto/tls"
 	"flag"
-	autoscalingcontroller "github.com/aibrix/aibrix/pkg/controller/podautoscaler"
 	"os"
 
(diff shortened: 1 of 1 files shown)


cmd/main.go
```diff
@@ -118,15 +118,15 @@ func main() {
 		// LeaderElectionReleaseOnCancel: true,
 	})
 	if err != nil {
-		setupLog.Error(err, "unable to start manager")
+		setupLog.Error(err, "unable to create manager")
```
Jeffwan, line 121: @varungup90 address the PR review from last round

Jeffwan: I need this change (folder structure) for the coming ones. I will merge it now. Feel free to leave comments if anything doesn't look good to you. I will address them later

Issue #22, opened 2024-07-13
[RFC] Support Lora management in kubernetes native way


Jeffwan:
After the interview discussion, we determine to create custom resource for Lora Adapters. We need to create the api and controllers for it. Detail design will come soon

- [x] Create the API and controller skeleton https://github.com/aibrix/aibrix/pull/23
- [x] Design the Lora API
- [x] Implement the Lora controllers Part I (Service & Endpoint slice controller)
- [x] Implement the Lora controllers Part II (Interact with inference engine for loading)
- [ ] Finish the routing algorithm design and implementation
- [ ] Design and implement cold start manager
- [ ] Design and implement the lora scheduler
- [x] Engine related changes https://github.com/aibrix/aibrix/issues/14

PR #32, opened 2024-07-14
Initial model adapter controller implementation

Jeffwan:
Address https://github.com/aibrix/aibrix/issues/31 and part of #22 

~~DO NOT MERGE it until #27 is merged and I will rebase the code and update this PR~~

### DIFF
--- a/pkg/controller/modeladapter/modeladapter_controller.go
+++ b/pkg/controller/modeladapter/modeladapter_controller.go
@@ -320,10 +320,7 @@ func (r *ModelAdapterReconciler) reconcileService(ctx context.Context, instance 
 	}
 
 	err := r.Get(ctx, objectKey, svc)
-	if err == nil {
-		// service exists, do nothing
-		return nil
-	} else if errors.IsNotFound(err) {
+	if errors.IsNotFound(err) {
 		// Service does not exist, create it
 		svc, err = buildModelAdapterService(ctx, *instance)
 		if err != nil {


pkg/controller/modeladapter/modeladapter_controller.go
```diff
@@ -41,27 +70,64 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) (reconcile.Reconciler, error) {
+	cacher := mgr.GetCache()
+
+	podInformer, err := cacher.GetInformer(context.TODO(), &corev1.Pod{})
```
varungup90, line 75: We can do this but it will be difficult to write unit tests later. I am ok, with adding a TODO for later or you can add them function arguments.

varungup90: There is no business logic in this fuction, so we dont need to write unit test for it. Ignore the previous comment.

pkg/controller/modeladapter/modeladapter_controller.go
```diff
@@ -89,7 +168,340 @@ type ModelAdapterReconciler struct {
...
+	// Retrieve the Service from the Kubernetes cluster with the name and namespace.
+	svc := &corev1.Service{}
+
+	objectKey := types.NamespacedName{
+		Namespace: instance.Namespace,
+		Name:      instance.Name,
+	}
+
+	err := r.Get(ctx, objectKey, svc)
+	if err == nil {
```
varungup90: you can remove the err == nil check.

Just one is `if errors.IsNotFound() {}` is enough.

pkg/controller/modeladapter/modeladapter_controller.go
Jeffwan: I considered to use informer here directly but currently use the high level interface. may do some refactor later here. so this is indeed not being used yet.

pkg/controller/modeladapter/modeladapter_controller.go
Jeffwan: make sense. Let me update the logic here.

PR #57, opened 2024-07-29
[Core] Add model router controller

varungup90:
Adds a quickstart to setup envoy gateway and simple routing on local

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -1,7 +1,9 @@
 package controller
 
 import (
+	"github.com/aibrix/aibrix/pkg/controller/modeladapter"
 	"github.com/aibrix/aibrix/pkg/controller/modelrouter"
+	"github.com/aibrix/aibrix/pkg/controller/podautoscaler"
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/klog/v2"
 	"sigs.k8s.io/controller-runtime/pkg/manager"
@@ -15,8 +17,8 @@ import (
 var controllerAddFuncs []func(manager.Manager) error
 
 func init() {
-	// controllerAddFuncs = append(controllerAddFuncs, podautoscaler.Add)
-	// controllerAddFuncs = append(controllerAddFuncs, modeladapter.Add)
+	controllerAddFuncs = append(controllerAddFuncs, podautoscaler.Add)
+	controllerAddFuncs = append(controllerAddFuncs, modeladapter.Add)
 	controllerAddFuncs = append(controllerAddFuncs, modelrouter.Add)
 }
 


pkg/gateway/README.md
```diff
@@ -0,0 +1,48 @@
+
```
Jeffwan: let's move files to `docs/development/gateway`

pkg/gateway/README.md
```diff
@@ -0,0 +1,48 @@
+
+
+## Installation <a name = "Installation"></a>
+
+```
+helm install eg oci://docker.io/envoyproxy/gateway-helm --version v1.1.0 -n envoy-gateway-system --create-namespace
+
+kubectl wait --timeout=5m -n envoy-gateway-system deployment/envoy-gateway --for=condition=Available
+
+kubectl apply -f quickstart.yaml -n default
```
Jeffwan: quickstart is an application but not the gateway itself. right? If so, let's separate from `installation`

pkg/gateway/README.md
```diff
@@ -0,0 +1,48 @@
...
+docker pull ghcr.io/cncf/mistral:latest
+
+kind load docker-image ghcr.io/cncf/mistral:latest
+```
+
+Port forward to the Envoy service:
+```
+kubectl -n envoy-gateway-system port-forward service/${ENVOY_SERVICE} 8888:80 &
+
+kubectl port-forward service/ollama 11434:6000 &
```
Jeffwan: if we already expose the gateway service, why do you need to port-forward service separately?

pkg/gateway/README.md
```diff
@@ -0,0 +1,48 @@
...
+```
+
+Curl the example app through Envoy proxy:
+```
+curl --verbose --header "Host: www.example.com" http://localhost:8888/get
+
+curl -X POST http://localhost:11434/api/generate -d '{
```
Jeffwan: We can do some refactor after we adopt the mock service for consistency

pkg/gateway/quickstart.yaml
```diff
@@ -0,0 +1,77 @@
...
+metadata:
+  name: eg
+spec:
+  gatewayClassName: eg
+  listeners:
+    - name: http
+      protocol: HTTP
+      port: 80
+---
+apiVersion: v1
```
Jeffwan: do we need separate service account?

pkg/gateway/quickstart.yaml
```diff
@@ -0,0 +1,77 @@
...
+    - backendRefs:
+        - group: ""
+          kind: Service
+          name: ollama
+          port: 6000
+          weight: 1
+      matches:
+        - path:
+            type: PathPrefix
+            value: /
```
Jeffwan: ollama path is different from openai. Let's adopt openai protocol compatible service in the testing.

pkg/gateway/quickstart.yaml
varungup90: Without service account, routing is not working. I will check for reasoning.

pkg/controller/modelrouter/modelrouter_controller.go
```diff
@@ -0,0 +1,127 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+	gatewayv1 "sigs.k8s.io/gateway-api/apis/v1"
+)
+
+const (
+	modelIdentifier       = "models.aibrix.com/model-name"
+	modelHeaderIdentifier = "model"
+	modelPortIdentifier   = "models.aibrix.com/model-port"
+	aibrixEnvoyGateway    = "eg" // parameterize it
```
Jeffwan: that's the envoy gateway name?

pkg/controller/modelrouter/modelrouter_controller.go
```diff
@@ -0,0 +1,127 @@
...
+									Value: modelName,
+								},
+							},
+						},
+					},
+					BackendRefs: []gatewayv1.HTTPBackendRef{
+						{
+							BackendRef: gatewayv1.BackendRef{
+								BackendObjectReference: gatewayv1.BackendObjectReference{
+									Name: gatewayv1.ObjectName(modelName),
```
Jeffwan, line 98: this the service name. here's the tricky part. If we do not manage service, then this won't work. in initial version, at least we need to let user know that they have to use exact same time for the service.

pkg/controller/modelrouter/modelrouter_controller.go
```diff
@@ -0,0 +1,127 @@
+package modelrouter
```
Jeffwan, line 1: this is good at this moment. We definitely have more features need to be added based on deployment configuration. I am still thinking meta or other names make more sense so we do not need to create additinal controllers. This can be discussed later

docs/tutorial/gateway/m2/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Use the official Python base image
```
Jeffwan: can we clean up `Dockerfile` and `requirements.txt`? check `docs/deployment/app`. Can we use that one? Let's make necessary changes there

docs/tutorial/gateway/m1/README.md
```diff
@@ -0,0 +1,26 @@
+# Deploy test model M1
+
+1. Builder mocked base model image
+```dockerfile
+docker build -t aibrix/vllmm1:v0.1.0 -f Dockerfile .
```
Jeffwan: Let's probably bump the version like v0.2.0 etc. we can pass the env `MODEL_NAME` to make sure it return right model you specified

docs/tutorial/gateway/m1/Dockerfile
```diff
@@ -0,0 +1,26 @@
+# Use the official Python base image
```
Jeffwan: duplicate. let's clean them up

pkg/controller/modelrouter/modelrouter_controller.go
varungup90: yes.

pkg/controller/modelrouter/modelrouter_controller.go
varungup90: yes, sounds good.

pkg/controller/modelrouter/modelrouter_controller.go
varungup90: yeah, I know. I am looking into if we can resolve service name from deployment.

varungup90: I will add a TODO.

pkg/gateway/quickstart.yaml
varungup90: removed it.

docs/tutorial/gateway/m1/README.md
varungup90: not needed now.

PR #79, opened 2024-08-12
[feat] Add rpm/tpm extension proc plugin

varungup90:
image


pkg/plugins/ext_proc/plugins.yaml
```diff
@@ -0,0 +1,34 @@
...
+kind: EnvoyExtensionPolicy
+metadata:
+  name: ext-proc-example
+spec:
+  targetRef:
+    group: gateway.networking.k8s.io
+    kind: HTTPRoute
+    name: llama2-70b-router
+  extProc:
+    - backendRefs:
```
Jeffwan: https://gateway.envoyproxy.io/contributions/design/envoy-extension-policy/ 
1. seems the format is a little bit different from docs. which one is the best practice?
2. `targetRef` could be different levels? I see in the doc example, it applies to gateway level, here you applied it to HTTPRoute level.

pkg/plugins/ext_proc/main.go
```diff
@@ -0,0 +1,278 @@
...
+type server struct {
+	ratelimiter ratelimiter.AccountRateLimiter
+}
+type healthServer struct{}
+
+func (s *healthServer) Check(ctx context.Context, in *healthPb.HealthCheckRequest) (*healthPb.HealthCheckResponse, error) {
+	return &healthPb.HealthCheckResponse{Status: healthPb.HealthCheckResponse_SERVING}, nil
+}
+
+func (s *healthServer) Watch(in *healthPb.HealthCheckRequest, srv healthPb.Health_WatchServer) error {
```
Jeffwan: what's the usage of `watch`?  is it the gRPC watcher?  A faster way to detect the failures compared to `Check`?

pkg/plugins/ext_proc/plugins.yaml
```diff
@@ -0,0 +1,34 @@
...
+        - name: grpc-ext-proc
+          port: 50052
+      processingMode:
+        request: {}
+        response: 
+          body: Streamed
```
Jeffwan: doc uses `STREAMED`, is capitalization matter here?

Jeffwan: BTW, what's the `Streamed`?

pkg/plugins/ext_proc/plugins.yaml
```diff
@@ -0,0 +1,34 @@
...
+# ---
+# apiVersion: gateway.networking.k8s.io/v1alpha3
```
Jeffwan: is this still helpful? If not, let's remove this part

pkg/plugins/ext_proc/rate_limiter/rate_limiter.go
```diff
@@ -0,0 +1,31 @@
...
+
+package ratelimiter
+
+import (
+	"context"
+)
+
+type AccountRateLimiter interface {
+	// Get the rate limit for the given key and return the current value.
+	Get(ctx context.Context, accountId string) (int64, error)
```
Jeffwan: accountId is business information. Can we use some other fields like `OPENAI_API_KEY`?

Jeffwan: This should be aligned with openai protocol

pkg/plugins/ext_proc/rate_limiter/redis.go
```diff
@@ -0,0 +1,89 @@
...
+const binSize = 64
+
+type redisAccountRateLimiter struct {
+	client     *redis.Client
+	name       string
+	windowSize time.Duration
+}
+
+// simple fixed window rate limiter
+func NewRedisAccountRateLimiter(name string, client *redis.Client, windowSize time.Duration) AccountRateLimiter {
```
Jeffwan: what's the name? how to use it?

pkg/plugins/ext_proc/rate_limiter/rate_limiter.go
```diff
@@ -0,0 +1,31 @@
...
+
+	GetLimit(ctx context.Context, accountId string) (int64, error)
```
Jeffwan: `GetLimit()` is to check the upper bound of the key?

pkg/plugins/ext_proc/rate_limiter/redis.go
```diff
@@ -0,0 +1,89 @@
...
+	}
+
+	return &redisAccountRateLimiter{
+		name:       name,
+		client:     client,
+		windowSize: windowSize,
+	}
+}
+
+func (rrl redisAccountRateLimiter) Get(ctx context.Context, accountId string) (int64, error) {
```
Jeffwan: I didn't understand this part. seems `genKey` takes the time into the consideration. why can it use `fmt.Sprintf("%s:%s", rrl.name, accountId)` instead?

pkg/plugins/ext_proc/plugins.yaml
varungup90: 1. It is the same, can you highlight the difference. Documentation is old so there are some changes in the yaml config.
2. I tried a week or so back on gateway level. It did not work, as in extension proc was not getting triggered on curl request. I will try again today and confirm that.

pkg/plugins/ext_proc/main.go
varungup90: It is from the sample code I was using to write extension proc. I did not explore this part, let me evaluate this part if it is needed.

pkg/plugins/ext_proc/plugins.yaml
varungup90: Capitalization does not matter here. But I will updated the yaml file to make it capital letters.

There are three options on how the body will be sent to extension from gateway. https://github.com/envoyproxy/gateway/blob/main/api/v1alpha1/ext_proc_types.go#L15

pkg/plugins/ext_proc/rate_limiter/redis.go
varungup90: I took this code piece from llmflow. Name is just any unique identifier to tag the keys as prefix.

pkg/plugins/ext_proc/rate_limiter/rate_limiter.go
varungup90: GetLimit gets the RPM or TPM limit values configured for the user. Get returns the current usage in the past window duration.

pkg/plugins/ext_proc/rate_limiter/redis.go
varungup90: It is creating buckets to add the values to make is easier for aggregation and cleanup (after expiration).

pkg/plugins/ext_proc/plugins.yaml
Jeffwan: here it is 

image

pkg/plugins/ext_proc/main.go
Jeffwan: yeah, it's low priority, we can skip it now.

pkg/plugins/ext_proc/rate_limiter/redis.go
Jeffwan: make sense.

pkg/plugins/ext_proc/rate_limiter/rate_limiter.go
Jeffwan: @varungup90 this is something I feel we can make it more generic. account looks weird in our scenario.

pkg/plugins/ext_proc/plugins.yaml
Jeffwan: > I tried a week or so back on gateway level. It did not work. 

em. Let's double check the official docs. they do support gateway `targetRef`. it might be outdated as you said.

pkg/plugins/ext_proc/rate_limiter/rate_limiter.go
varungup90: it is a simple key for lookup. I will rename to "key".

pkg/plugins/ext_proc/plugins.yaml
varungup90: 1. move to use gateway as targetRef. 
2. Only lowercase "Streamed" is only supported now.
3. Spec for extProc is old, it does not have service -> backendRef hierarchy now.

PR #86, opened 2024-08-16
Add host override to query specific pod

varungup90:
Query specific pod by including pod's ip address and port

```
curl -v http://localhost:8888/v1/chat/completions \
  -H "target-pod: 10.244.1.3:8000" \
  -H "user: varun" \
  -H "Content-Type: application/json" \
  -H "Authorization: Bearer any_key" \
  -d '{
     "model": "llama2-70b",
     "messages": [{"role": "user", "content": "Say this is a test!"}],
     "temperature": 0.7
   }'
```


docs/development/app/gateway.yaml
```diff
@@ -14,4 +14,64 @@ spec:
...
+      path: "/virtual_hosts/0/routes/0"
+      value:
+        name: orginal_route
+        match:
+          prefix: "/"
+          headers:
+            # update ip address as needed and in production this config is not needed as backend will derive the pod ip
+            - name: "target-pod"
+              string_match:
+                exact: "10.244.1.3:8000"
```
Jeffwan: em. this seems like a very static setting? is the `10.244.1.3` backend pod?

pkg/plugins/ratelimiter/README.md
```diff
@@ -24,4 +22,27 @@ curl -v http://localhost:8888/v1/chat/completions \
      "messages": [{"role": "user", "content": "Say this is a test!"}],
      "temperature": 0.7
    }'
+```
+
+# restart envoy gateway after applying envoy patch policy
+```shell
+kubectl apply -f docs/development/app/gateway.yaml 
+k describe envoypatchpolicy epp
+egctl config envoy-proxy route
```
Jeffwan: is this the only way to apply the route? I mean using `egctl`, is there a way to apply it using crd etc so we can use `kubectl` only. Just like to check how many tool chains we need

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/aibrix/aibrix
 
-go 1.21
+go 1.22.5
+
+toolchain go1.22.6
```
Jeffwan, line 5: just curious why do we choose `1.22.5` and toolchain `1.22.6`. seem it's patch version difference.

docs/development/app/gateway.yaml
varungup90: I have added a comment, this setting is not needed in production, it is for test purpose only. I am also trying to add a regex for ip address.

pkg/plugins/ratelimiter/README.md
varungup90: config apply is using the command "kubectl apply -f docs/development/app/gateway.yaml".

these two commands are`get`, just to ensure that config is applied.
kubectl describe envoypatchpolicy epp
egctl config envoy-proxy route

go.mod
varungup90: go mod tidy did that, I did not notice that before. Let me fix it.

docs/development/app/gateway.yaml
varungup90: Added a random router, which selects the pod randomly. It also includes an interface, so new algorithms can be developed.

varungup90: Also commented the target-pod config from gateway.yaml as that is not needed with routing algorithm implemented.

pkg/plugins/ratelimiter/routing_algorithms/router.go
```diff
@@ -0,0 +1,29 @@
+/*
```
Jeffwan: Should the router be separate plugin?
- if we want to have different plugins, we should have routing algorithm separate from rate limiter. 
- if we want to build monolithic LLM gateway, we should rename to something else instead of `ratelimiter`

varungup90: I was thinking, we can start with one plugin as it will easy to debug. So for now, I will rename the ratelimiter dir -> gateway and when I refactor the code, will move routing algorithms in different dir and rename them.

Jeffwan: that make sense

PR #97, opened 2024-08-26
Add custom CRD clientset

varungup90:
Address https://github.com/aibrix/aibrix/issues/98


hack/verify-codegen.sh
```diff
@@ -0,0 +1,53 @@
...
+#
+cd ${SCRIPT_ROOT}
+${CODEGEN_PKG}/generate-groups.sh all \
+ github.com/aibrix/aibrix/pkg/client github.com/aibrix/aibrix/api \
+ "model:v1alpha1 autoscaling:v1alpha1" \
+ --output-base "${TEMP_DIR}" \
+ --go-header-file hack/boilerplate.go.txt
+
+# Copy everything back.
+cp -a "${TEMP_DIR}/${ROOT_PKG}/." "${SCRIPT_ROOT}/"
```
Jeffwan, line 53: Let's ignore the `pkg/.DS_Store`

hack/verify-codegen.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
Jeffwan, line 1: Let's create a follow up PR to verify the code-gen in the CI. it should blocks the Pr if there're outdated version

varungup90: Sure.

hack/verify-codegen.sh
varungup90: Done.

PR #100, opened 2024-08-27
Add custom cache and interface for model adapter scheduling

varungup90:


config/manager/kustomization.yaml
```diff
@@ -4,5 +4,5 @@ apiVersion: kustomize.config.k8s.io/v1beta1
 kind: Kustomization
 images:
 - name: controller
-  newName: aibrix/aibrix-controller-manager
-  newTag: v0.1.0-rc.0
+  newName: controller
```
Jeffwan: can we change back to the original name here?

PR #111, opened 2024-08-29
Apply filter to objects owned by model adapters

varungup90:


pkg/controller/modeladapter/modeladapter_controller.go
```diff
@@ -129,9 +128,9 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 	// use the builder fashion. If we need more fine grain control later, we can switch to `controller.New()`
 	err := ctrl.NewControllerManagedBy(mgr).
 		For(&modelv1alpha1.ModelAdapter{}).
-		Watches(&corev1.Service{}, &handler.EnqueueRequestForObject{}).
-		Watches(&discoveryv1.EndpointSlice{}, &handler.EnqueueRequestForObject{}).
-		Watches(&corev1.Pod{}, &handler.EnqueueRequestForObject{}).
+		Owns(&corev1.Service{}).
```
Jeffwan, line 131: since we create the resource, it should be owns. Looks good to me

PR #122, opened 2024-09-04
Add gateway installation

varungup90:
Address https://github.com/aibrix/aibrix/issues/114

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -161,7 +161,7 @@ endif
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | $(KUBECTL) apply -f -
     ## helm creates objects without aibrix prefix, hence deploying gateway components outside of kustomization	
-	helm install eg oci://docker.io/envoyproxy/gateway-helm --version v1.1.0 -n aibrix-system --create-namespace
+	helm install eg oci://docker.io/envoyproxy/gateway-helm --version v1.1.0 -n aibrix-system
 	$(KUBECTL) wait --timeout=5m -n aibrix-system deployment/envoy-gateway --for=condition=Available
 	$(KUBECTL) apply -f config/gateway/gateway.yaml
 	
--- a/docs/development/app/README.md
+++ b/docs/development/app/README.md
@@ -60,7 +60,7 @@ set aibrix:your-user-name_RPM_LIMIT 10
 
 Test request (ensure header model name matches with deployment's model name for routing)
 ```shell
-curl http://localhost:8888/v1/chat/completions \
+curl -v http://localhost:8888/v1/chat/completions \
   -H "user: your-user-name" \
   -H "model: llama2-70b" \
   -H "Content-Type: application/json" \


Makefile
```diff
@@ -160,10 +160,17 @@ endif
 .PHONY: install
 install: manifests kustomize ## Install CRDs into the K8s cluster specified in ~/.kube/config.
 	$(KUSTOMIZE) build config/crd | $(KUBECTL) apply -f -
+    ## helm creates objects without aibrix prefix, hence deploying gateway components outside of kustomization	
+	helm install eg oci://docker.io/envoyproxy/gateway-helm --version v1.1.0 -n aibrix-system --create-namespace
```
Jeffwan: line 161, `kustomize` should create the namespace. Did you see any problems missing the namespace here? We can leave the namespace here, I just like to confirm the sequence.

pkg/plugins/gateway/gateway.go
```diff
@@ -183,7 +184,7 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, req *extProcPb.Proces
 		}, user, targetPodIP
 	}
 
-	pods, err := s.client.CoreV1().Pods("default").List(ctx, v1.ListOptions{
+	pods, err := s.client.CoreV1().Pods(podutils.NAMESPACE).List(ctx, v1.ListOptions{
```
Jeffwan, line 187: Let's use injected env instead later. the default could be same

config/gateway/kustomization.yaml
```diff
@@ -0,0 +1,3 @@
+resources:
+- redis.yaml
+- gateway-plugin.yaml
```
Jeffwan, line 3: should it include `gateway.yaml` here? or `gateway.yaml` has to be installed earlier?

Makefile
```diff
@@ -160,10 +160,17 @@ endif
...
+	$(KUBECTL) wait --timeout=5m -n aibrix-system deployment/envoy-gateway --for=condition=Available
+	$(KUBECTL) apply -f config/gateway/gateway.yaml
```
Jeffwan, line 166: em. Now we install kustomize package (gateway-plugin) -> dependencies -> gateway.yaml. The sequence is not that clean. Not sure any problem moving gateway.yaml under kustomize? We probably can do 3rd party dependencies (gateway) first, and then aibrix.

Makefile
varungup90: I will remove the `-create namespace`.

config/gateway/kustomization.yaml
varungup90: gateway.yaml is installed in `make install`. So here only gateway-plugin.yaml

Makefile
varungup90: make install does two things-> envoy gateway (using helm command) + gateway dependencies such as envoy patch policy & envoy extension policy (using gateway.yaml)

make deploy install redis and gateway plugins (ext proc).

pkg/plugins/gateway/gateway.go
varungup90: for now will keep podutils.NAMESPACE and later move to injected env

Jeffwan: sounds good.

Makefile
Jeffwan: development flow might be different from one-click prod installation flow. If this is the flow. then we do not have `gateway.yaml` and `helm` package included. 
https://github.com/aibrix/aibrix/blob/0d64aca2c437b74fc64a4f0df9b6e81d7399ed95/README.md?plain=1#L16-L25

PR #132, opened 2024-09-05
Update make generate

varungup90:
Address https://github.com/aibrix/aibrix/issues/131 https://github.com/aibrix/aibrix/issues/130

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -74,7 +74,7 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .PHONY: test
-test: manifests fmt vet envtest ## Run tests.
+test: manifests generate fmt vet envtest ## Run tests.
 	KUBEBUILDER_ASSETS="$(shell $(ENVTEST) use $(ENVTEST_K8S_VERSION) --bin-dir $(LOCALBIN) -p path)" go test $$(go list ./... | grep -v /e2e) -coverprofile cover.out
 
 # Utilize Kind or modify the e2e tests to load the image locally, enabling compatibility with other vendors.
@@ -109,11 +109,11 @@ lint-all: licensecheck lint
 ##@ Build
 
 .PHONY: build
-build: manifests fmt vet ## Build manager binary.
+build: manifests generate fmt vet ## Build manager binary.
 	go build -o bin/manager cmd/controllers/main.go
 
 .PHONY: run
-run: manifests fmt vet ## Run a controller from your host.
+run: manifests generate fmt vet ## Run a controller from your host.
 	go run ./cmd/controllers/main.go
 
 # If you wish to build the manager image targeting other platforms you can use the --platform flag.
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -53,9 +53,9 @@ help: ## Display this help.
 manifests: controller-gen ## Generate WebhookConfiguration, ClusterRole and CustomResourceDefinition objects.
 	$(CONTROLLER_GEN) rbac:roleName=manager-role crd:maxDescLen=0 webhook paths="./..." output:crd:artifacts:config=config/crd/bases
 
-.PHONY: generate
```
Jeffwan, line 56: I think we should leverage controller-gen as much as possible. If anything is missing, we can leverage traditional way to generate them.

Makefile
```diff
@@ -53,9 +53,9 @@ help: ## Display this help.
...
-generate: controller-gen ## Generate code containing DeepCopy, DeepCopyInto, and DeepCopyObject method implementations.
-	$(CONTROLLER_GEN) object:headerFile="hack/boilerplate.go.txt" paths="./..."
+.PHONY: verify
```
Jeffwan: verify-codegen

api/autoscaling/v1alpha1/podautoscaler_types.go
```diff
@@ -121,6 +121,7 @@ type PodAutoscalerStatus struct {
 }
 
 // +kubebuilder:object:root=true
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
```
Jeffwan, line 124: since we have package level definition in `doc.go` I feel this can be removed?

Makefile
varungup90: Ok. Once I fix the issue to run verify using make, will use the name verify-codegen.

Makefile
```diff
@@ -66,7 +74,7 @@ vet: ## Run go vet against code.
 	go vet ./...
 
 .PHONY: test
-test: manifests generate fmt vet envtest ## Run tests.
```
Jeffwan: just like to confirm that `generate` should be added back?

api/autoscaling/v1alpha1/podautoscaler_types.go
Jeffwan: if we do not have `doc.go` we can keep it here.

varungup90: removed the doc.go

Makefile
varungup90: added make verify-codegen in Makefile and github workflow.

Makefile
varungup90: restored make generate

Makefile
varungup90: yes, forgot to add at other places.

PR #143, opened 2024-09-09
Add routing algorithms

varungup90:


docs/development/app/deployment.yaml
```diff
@@ -99,7 +99,7 @@ roleRef:
 apiVersion: rbac.authorization.k8s.io/v1
 kind: Role
 metadata:
-  namespace: default
+  namespace: aibrix-system
```
Jeffwan: this has been fixed in last PR. Let's rebase the main branch

docs/development/app/gateway-plugin.yaml
```diff
@@ -0,0 +1,47 @@
...
+        app: aibrix-gateway-plugins
+    spec:
+      containers:
+        - name: golang-app-container
+          image: aibrix/plugins:v0.1.0
+          ports:
+            - containerPort: 50052
+          env:
+            - name: REDIS_HOST
+              value: aibrix-redis-master
```
Jeffwan, line 36: for the dev env, do we have a redis installed? this yaml doesn't have redis yet

pkg/plugins/gateway/gateway.go
```diff
@@ -425,3 +462,20 @@ func (s *Server) checkTPM(ctx context.Context, user string) (envoyTypePb.StatusC
 
 	return envoyTypePb.StatusCode_OK, nil
 }
+
+func (s *Server) SelectTargetPod(ctx context.Context, routingStrategy string, pods []corev1.Pod) (string, error) {
+	// TODO (varun): evaluate how to enable selection of routing algorithm
+	var route routing.Router
```
Jeffwan, line 474: em. let's avoid initialization every time?

pkg/plugins/gateway/routing_algorithms/least_request.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+}
+
+func (r leastRequestRouter) Get(ctx context.Context, pods []v1.Pod) (string, error) {
+	var targetPodIP string
+	minCount := math.MaxInt
+
+	for _, pod := range pods {
+		podIP := pod.Status.PodIP + ":8000"
+		reqCount, err := r.ratelimiter.Get(ctx, fmt.Sprintf("%v_REQUEST_COUNT", podIP))
```
Jeffwan: the key is something like 10.0.0.1_REQUEST_COUNT? pod ip or name?

pkg/plugins/gateway/routing_algorithms/least_request.go
```diff
@@ -0,0 +1,57 @@
...
+	"fmt"
+	"math"
+
+	ratelimiter "github.com/aibrix/aibrix/pkg/plugins/gateway/rate_limiter"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+)
+
+type leastRequestRouter struct {
+	ratelimiter ratelimiter.AccountRateLimiter
```
Jeffwan, line 31: I am thinking whether ratelimit reflects the current state? it gives the number of least request in the history, but not the current moment, right?

Jeffwan: Technically, number of least request should check `inflight-requests` from the engine and then made the decision

pkg/plugins/gateway/routing_algorithms/least_request.go
varungup90: like this: 10.0.0.1_REQUEST_COUNT

pkg/plugins/gateway/routing_algorithms/least_request.go
varungup90: Yes, initially I implemented a decr function also but issue with decr is that if that fails then counter will never be decremented.

I just thought about it, since we are doing in extension proc and it will always have one installation, we do not need redis. We can simply use in-memory cache also. WDYT?

varungup90: Moved to in-memory cache. Now only inflight requests are used to select the pod.

Jeffwan: in-memory looks good for single instance now. the data plane needs to be high available and we have to address the data synchronization issue later.

config/gateway/gateway-plugin.yaml
```diff
@@ -45,4 +45,4 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: metadata.namespace
-
+      serviceAccountName: controller-manager
```
Jeffwan: can we use a separate sa for gateway plugin? `controller-manager` role setting is pretty open.

config/manager/manager.yaml
```diff
@@ -38,7 +38,7 @@ spec:
           - --leader-elect
           - --health-probe-bind-address=:8081
           - --metrics-bind-address=0
-        image: controller:latest
+        image: aibrix/controller-manager:nightly
```
Jeffwan: https://github.com/aibrix/aibrix/blob/0755f3a14d0398d526da71bbd5cbecde81d6ad30/config/manager/kustomization.yaml#L6-L8 override the image. if we have the change here, we need corresponding change there as well

docs/development/app/deployment.yaml
```diff
@@ -71,52 +71,4 @@ spec:
 #               value: llama2-70b
 #       backendRefs:
 #         - name: llama2-70b
-#           port: 8000
----
```
Jeffwan, line 75: @kr11 I remember you made some changes on the role setting as well? Can you confirm whether deletion works for your testing?

config/manager/manager.yaml
varungup90: right now I have undo this change.

config/gateway/gateway-plugin.yaml
varungup90: done

docs/development/app/deployment.yaml
Jeffwan: Let's merge this change now, if there's issues found, let's fix them.

kr11: Sorry for missing this message. 

Yes, the endpoint `pod_ip/metrics` attempts to fetch the most recent replica counts in order to return a mocked metric usage that is inversely proportional to the number of replicas. Is it necessary to remove the RBAC settings? Could we consider reverting that change? @varungup90 @Jeffwan

PR #181, opened 2024-09-13
Add users CRUD API

varungup90:


Dockerfile.users
```diff
@@ -0,0 +1,36 @@
+## Multistage build
```
Jeffwan: Let's try not to use `users` concept? If that's the only feature, it's too thin I feel. can we have something aggregated and call it metadata etc?

Dockerfile.users
```diff
@@ -0,0 +1,36 @@
...
+RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o users cmd/main.go
+
+# Use distroless as minimal base image to package the manager binary
+# Refer to https://github.com/GoogleContainerTools/distroless for more details
+FROM gcr.io/distroless/static:nonroot
+WORKDIR /
+COPY --from=builder /workspace/users .
+USER 65532:65532
+
+ENTRYPOINT ["/users"]
```
Jeffwan, line 36: seems the only difference is the start command? In that case, can we reuse the same image as controller but use different entrypoints?

Makefile
```diff
@@ -126,14 +127,19 @@ docker-build: ## Build docker image with the manager.
 
 .PHONY: docker-build-plugins
 docker-build-plugins: ## Build docker image with the manager.
-	$(CONTAINER_TOOL) build -t ${PLUGINS_IMG} -f gateway.Dockerfile .
+	$(CONTAINER_TOOL) build -t ${PLUGINS_IMG} -f Dockerfile.gateway .
```
Jeffwan, line 130: Dockerfile.<purpose> is the best practice, others are still same. I suggest to have a separate PR to change them together. we still have `runtime` etc not renamed yet. In this PR, let's keep the original names?

pkg/utils/users.go
```diff
@@ -0,0 +1,71 @@
...
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/redis/go-redis/v9"
+)
+
+type User struct {
```
Jeffwan, line 27: will any other packages use this utils? If not, I suggest to move under the user packages.

pkg/plugins/gateway/gateway.go
```diff
@@ -35,27 +37,34 @@ import (
...
 )
 
+var (
+	defaultRPM           = 100
+	defaultTPMMultiplier = 1000
+)
+
 type Server struct {
 	routers             map[string]routing.Router
+	redisClient         *redis.Client
```
Jeffwan, line 55: is there a feature gate for rate limit feature? In the v0.1.0 release, we need to consider the usability and may disable some features by default. We need to evaluate the steps based on user feedback later

Dockerfile.users
varungup90: good idea, even gateway plugin does not have any difference. I can merge all three in same image with different start commands.

pkg/utils/users.go
varungup90: I started with that, but gateway need to call get user. I can replicate getUser command for gateway and merge utils/users.go with is primary package.

pkg/plugins/gateway/gateway.go
varungup90: Right now there is no feature gate, but if user does not provide it then it default to a high value.

Dockerfile.users
varungup90: I am thinking this feature will eventually beef up as we add more metadata. Litellm also has users endpoint and gradually we may have similar feature parity.

pkg/utils/users.go
varungup90: gateway also needs user struct to unmarshal the response.

Jeffwan: got you. then it makes sense to be here

pkg/plugins/gateway/gateway.go
Jeffwan: last time when I use it, I have to add a user with correct settings. We need to make sure some steps can be skipped

Dockerfile.users
Jeffwan: ok. let's merge this now and revisit it later.

Dockerfile.gateway
```diff
@@ -0,0 +1,4 @@
+## Multistage build
+FROM aibrix/controller-manager:nightly
```
Jeffwan: em. Looks like we can not do this way.. seems this will use this previous build instead of the current.

pkg/plugins/gateway/gateway.go
varungup90: are you referring to manually configuring using redis-cli?

with API, username is required, rpm and tpm are optional.

Dockerfile.gateway
varungup90: For now, should I just revert to old way?

Jeffwan: Yes. Let's revert it back and use `COPY pkg/ pkg/` for simplicity

pkg/plugins/gateway/gateway.go
Jeffwan: Got it. Let's improve the UX in later PR. Let's merge this one first

PR #183, opened 2024-09-16
Add routing for model adapter

varungup90:
https://github.com/aibrix/aibrix/issues/146

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -52,6 +52,10 @@ var (
 	instance Cache
 )
 
+const (
+	modelIdentifier = "model.aibrix.ai"
+)
+
 func GetCache() (*Cache, error) {
 	if !instance.initialized {
 		return nil, errors.New("cache is not initialized")
@@ -128,7 +132,7 @@ func (c *Cache) addPod(obj interface{}) {
 
 	pod := obj.(*v1.Pod)
 	// only track pods with model deployments
-	modelName, ok := pod.Labels[modelv1alpha1.GroupVersion.Group]
+	modelName, ok := pod.Labels[modelIdentifier]
 	if !ok {
 		return
 	}
@@ -144,13 +148,13 @@ func (c *Cache) updatePod(oldObj interface{}, newObj interface{}) {
 	defer c.mu.Unlock()
 
 	oldPod := oldObj.(*v1.Pod)
-	oldModelName, ok := oldPod.Labels[modelv1alpha1.GroupVersion.Group]
+	oldModelName, ok := oldPod.Labels[modelIdentifier]
 	if !ok {
 		return
 	}
 
 	newPod := newObj.(*v1.Pod)
-	newModelName, ok := oldPod.Labels[modelv1alpha1.GroupVersion.Group]
+	newModelName, ok := oldPod.Labels[modelIdentifier]
 	if !ok {
 		return
 	}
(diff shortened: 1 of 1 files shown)


pkg/cache/cache.go
```diff
@@ -41,12 +40,12 @@ var once sync.Once
...
-	initialized              bool
-	pods                     map[string]*v1.Pod
-	modelAdapterToPodMapping map[string][]string
-	podToModelAdapterMapping map[string]map[string]struct{}
-	podRequestTracker        map[string]int
+	mu                sync.RWMutex
+	initialized       bool
+	pods              map[string]*v1.Pod
+	podToModelMapping map[string]map[string]struct{} // pod_name: map[model_name]struct{}
+	modelToPodMapping map[string]map[string]*v1.Pod  // model_name: map[pod_name]*v1.Pod
```
Jeffwan, line 47: this is to make the model more generic? it can be used for both modeladapter and base model?

pkg/cache/cache.go
```diff
@@ -112,9 +111,9 @@ func NewCache(config *rest.Config, stopCh <-chan struct{}) *Cache {
 		}
 
 		if _, err = modeInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
```
Jeffwan: nit: modeInformer ->  modelInformer

pkg/plugins/gateway/gateway.go
```diff
@@ -155,118 +153,66 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, req *extProcPb.Proces
...
-									Key:      "x-rpm-exceeded",
-									RawValue: []byte("true"),
-								},
-							},
-						},
-					},
-				},
-			},
-		}, user, targetPodIP
+		return generateErrorResponse(
```
Jeffwan, line 160: most of the changes in this file are refactor, right? it's to leverage the `generateErrorResponse `

pkg/cache/cache.go
```diff
@@ -128,104 +127,164 @@ func (c *Cache) addPod(obj interface{}) {
 	defer c.mu.Unlock()
 
 	pod := obj.(*v1.Pod)
+	// only track pods with model deployments
+	modelName, ok := pod.Labels[modelv1alpha1.GroupVersion.Group]
```
Jeffwan: here, it uses model adapters group? I remember we didn't update the pod if user applies `ModelAdapter` CR.  on the other hand, we will ask user to specifiy base model label on the deployment. 

the goal here is to track the model adapter or model?

pkg/cache/cache.go
varungup90: yes.

pkg/plugins/gateway/gateway.go
varungup90: yes. Another change is to use internal cache for getting pods for a model or model adapter rather than calling API server.

pkg/cache/cache.go
varungup90: This is to track base model. Model adapter to pods are tracked based on status.instances.

Jeffwan: when why `modelv1alpha1.GroupVersion.Group`? Even the string is same, but `modelv1alpha1.GroupVersion.Group` is not for this purpose?

varungup90: Ok, I was using for const string. I will define another string const value for this purpose.

Issue #234, opened 2024-09-26
RPM for CRUD user could be set negative number


brosoul:
### 🐛 Describe the bug

RPM for CRUD user could be set negative number.
![img_v3_02f3_7a7c59cf-7338-4a71-b077-b1b6a82a47fg](https://github.com/user-attachments/assets/74d2696e-48e7-49ac-b6d4-7d57cff6fe06)

### Steps to Reproduce

```bash
curl http://localhost:8090/CreateUser \
  -H "Content-Type: application/json" \
  -d '{"name": "your-user-name","rpm": -1,"tpm": 1000}'
```

### Expected behavior

Creation failed and provided some prompt information

PR #241, opened 2024-09-26
Add validation for username and rpm/tpm negative value

varungup90:
#234

### DIFF
--- a/pkg/users/users.go
+++ b/pkg/users/users.go
@@ -68,7 +68,7 @@ func (s *httpServer) createUser(w http.ResponseWriter, r *http.Request) {
 	}
 
 	if err := utils.SetUser(u, s.redisClient); err != nil {
-		fmt.Fprintf(w, "error occurred on creating user: %+v", err)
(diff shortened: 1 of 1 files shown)


pkg/utils/users.go
```diff
@@ -54,6 +54,10 @@ func GetUser(u User, redisClient *redis.Client) (User, error) {
 }
 
 func SetUser(u User, redisClient *redis.Client) error {
+	if u.Rpm < 0 || u.Tpm < 0 {
+		return fmt.Errorf("rpm or tpm can not negative")
```
Jeffwan, line 58: what's the http error to the client for this case now?

varungup90: internal server error: 500

Jeffwan: should it be something like invalid argument? 400

varungup90: Good point. I will audit all endpoints and their return error codes.

PR #244, opened 2024-09-26
Fix concurrency issue with gateway RPM plugin

varungup90:
https://github.com/aibrix/aibrix/issues/254

### DIFF
--- a/pkg/cache/cache.go
+++ b/pkg/cache/cache.go
@@ -139,7 +139,7 @@ func (c *Cache) addPod(obj interface{}) {
 
 	c.pods[pod.Name] = pod
 	c.addPodAndModelMapping(pod.Name, modelName)
-	klog.Infof("POD CREATED: %s/%s", pod.Namespace, pod.Name)
+	klog.V(4).Infof("POD CREATED: %s/%s", pod.Namespace, pod.Name)
 	c.debugInfo()
 }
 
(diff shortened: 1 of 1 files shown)


pkg/cache/cache.go
```diff
@@ -159,9 +159,11 @@ func (c *Cache) updatePod(oldObj interface{}, newObj interface{}) {
 		return
 	}
 
+	delete(c.pods, oldPod.Name)
+	c.pods[newPod.Name] = newPod
 	c.deletePodAndModelMapping(oldPod.Name, oldModelName)
 	c.addPodAndModelMapping(newPod.Name, newModelName)
-	klog.Infof("POD UPDATED. %s/%s %s", oldPod.Namespace, oldPod.Name, newPod.Status.Phase)
```
Jeffwan, line 164: can we change this to v(4). We can use log level to control the messages. I see lots of logs in the controller manager

pkg/plugins/gateway/gateway.go
```diff
@@ -182,6 +183,17 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, req *extProcPb.Proces
 			fmt.Sprintf("pre query: error on checking rpm: %v", err.Error())), username, targetPodIP
 	}
 
+	rpm, code, err := s.incrRPM(ctx, username)
```
Jeffwan, line 186: What's the purpose here?

pkg/cache/cache.go
varungup90: Done.

pkg/plugins/gateway/gateway.go
varungup90: Previously increment was done at the end of the request, so problem was that with burst of request. Now rpm is tracked as the request arrives.

PR #273, opened 2024-10-03
Add container override for images

varungup90:
Address https://github.com/aibrix/aibrix/issues/233

### DIFF
--- a/config/overlays/vke/default/kustomization.yaml
+++ b/config/overlays/vke/default/kustomization.yaml
@@ -7,4 +7,10 @@ resources:
 - ../../../rbac
 - manager
 - gateway
-- ../dependency/kuberay-operator
+- ../../../dependency/kuberay-operator
+
+
+images:
+- name: quay.io/kuberay/operator
+  newName: aibrix-container-registry-cn-beijing.cr.volces.com/aibrix/quay.io/kuberay/operator
+  newTag: v1.2.1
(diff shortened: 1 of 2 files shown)


config/overlays/vke/default/gateway/kustomization.yaml
```diff
@@ -0,0 +1,26 @@
+resources:
+  - ../../../../gateway
+  - envoyproxy.yaml
```
Jeffwan, line 3: just curious why can not we use envoyproxy + gatreway_patch.yaml from the original folder? If we just like to override the images? I think the one under default/kustomization.yaml is good enough?

config/overlays/vke/dependency/kustomization.yaml
```diff
@@ -0,0 +1,4 @@
+kind: Kustomization
```
Jeffwan, line 1: if we move image override to this folder. does it mean we just need this single file? 

image

config/overlays/vke/default/gateway/kustomization.yaml
varungup90: There is no envoyproxy + gateway_patch in config/gateway. Its a override added to replace envoyproxy image, which was not needed in US env.

config/overlays/vke/dependency/kustomization.yaml
varungup90: In dependency, envoy-gateway has different namespace. So that will always be separate. In other cases, we can not merge because each one references to their parent also. When I tried to club everything together, resources order is not guaranteed and it was not working.

config/overlays/vke/default/gateway/kustomization.yaml
Jeffwan: Got it. Makes sense. I misunderstand this part

PR #274, opened 2024-10-04
Use vllm metrics for routing

varungup90:


pkg/cache/cache.go
```diff
@@ -339,25 +365,95 @@ func (c *Cache) GetModelsForPod(podName string) (map[string]struct{}, error) {
...
+	metricValue, err := parseMetricFromBody(body, metricName)
+	if err != nil {
+		return 0, fmt.Errorf("failed to parse metrics from pod %s %s %d: %v", pod.Name, pod.Status.PodIP, metricsPort, err)
+	}
+
+	klog.V(5).InfoS("Successfully parsed metrics", "metric", metricName, "PodIP", pod.Status.PodIP, "Port", metricsPort, "metricValue", metricValue)
+	return metricValue, nil
+}
+
+func parseMetricFromBody(body []byte, metricName string) (float64, error) {
```
Jeffwan, line 429: autoscaler has similar features. Let's refactor this part later and make sure cache and autoscaler fetcher can use same library

pkg/cache/cache.go
```diff
@@ -339,25 +365,95 @@ func (c *Cache) GetModelsForPod(podName string) (map[string]struct{}, error) {
...
+		}
+	}
 }
 
-func (c *Cache) GetPodRequestCount() map[string]int {
-	c.mu.RLock()
-	defer c.mu.RUnlock()
+func getMetricsForPod(pod *v1.Pod, metricName string, metricsPort int) (float64, error) {
+	// We should use the primary container port. In future, we can decide whether to use sidecar container's port
+	url := fmt.Sprintf("http://%s:%d/metrics", pod.Status.PodIP, metricsPort)
```
Jeffwan: there's no need to fetch the `/metrics` endpoint for every metrics. Let's just fetch once and follow up metric reading can just parse it.

Jeffwan: This is improvement instead of feature so we can wait for future PR to enhance this part.

pkg/plugins/gateway/algorithms/least_request.go
```diff
@@ -44,22 +43,39 @@ func NewLeastRequestRouter(ratelimiter ratelimiter.RateLimiter) Router {
 	}
 }
 
-func (r leastRequestRouter) Get(ctx context.Context, pods map[string]*v1.Pod) (string, error) {
+func (r leastRequestRouter) Route(ctx context.Context, pods map[string]*v1.Pod) (string, error) {
 	var targetPodIP string
-	minCount := math.MaxInt
-	podRequestCounts := r.cache.GetPodRequestCount()
```
Jeffwan, line 50: this is expected, right? Let's always use fetched metrics instead of doing counting at the gateway side?

pkg/cache/cache.go
varungup90: Agree. Let me discuss with him

pkg/cache/cache.go
varungup90: oh yeah. This is easy change let me do it.

pkg/plugins/gateway/algorithms/least_request.go
varungup90: Yes, I have removed counting for requests and throughput (token per min) from gateway.

pkg/cache/cache.go
varungup90: Refactor the cache.

PR #290, opened 2024-10-10
Read model  name from request body

varungup90:


config/gateway/gateway-plugin.yaml
```diff
@@ -50,3 +50,24 @@ spec:
...
       serviceAccountName: aibrix-gateway-plugin
+---
+# this is a dummy route for incoming request and,
+# then request is routed to httproute using model name OR
+# request is routed based on the target for that model service
+# TODO (varun): check if this dummy route can be removed in future
+apiVersion: gateway.networking.k8s.io/v1
+kind: HTTPRoute
+metadata:
+  name: dummy-router
```
Jeffwan: can we rename to something like `aibrix-reserved-router`?

Jeffwan: What's the necessity of this HttpRoute?

varungup90: Previously with model-name in the header, there was always a base http route for incoming request but now there is no base route. 

renamed it.

PR #299, opened 2024-10-16
Add model API

varungup90:
Address https://github.com/aibrix/aibrix/issues/302

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,10 +34,10 @@ kubectl create -k config/default
 Install stable distribution
 ```shell
 # Install component dependencies
-kubectl create -k "github.com/aibrix/aibrix/config/dependency?ref=v0.1.0-rc.3"
+kubectl create -k "github.com/aibrix/aibrix/config/dependency?ref=v0.1.0-rc.4"
 
 # Install aibrix components
-kubectl create -k "github.com/aibrix/aibrix/config/default?ref=v0.1.0-rc.3"
+kubectl create -k "github.com/aibrix/aibrix/config/default?ref=v0.1.0-rc.4"
 ```
 
 ## Documentation
--- a/config/overlays/vke/default/kustomization.yaml
+++ b/config/overlays/vke/default/kustomization.yaml
@@ -12,5 +12,5 @@ resources:
 
 images:
 - name: quay.io/kuberay/operator
-  newName: aibrix-container-registry-cn-beijing.cr.volces.com/aibrix/quay.io/kuberay/operator
-  newTag: v1.2.1
+  newName: aibrix-container-registry-cn-beijing.cr.volces.com/aibrix/kuberay-operator
+  newTag: v1.2.1-patch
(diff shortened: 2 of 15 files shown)


api/model/v1alpha1/model_types.go
```diff
@@ -0,0 +1,110 @@
...
+import (
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+// EDIT THIS FILE!  THIS IS SCAFFOLDING FOR YOU TO OWN!
+// NOTE: json tags are required.  Any new fields you add must have json tags for the fields to be serialized.
+
+// ModelSpec defines the desired state of Model
+type ModelSpec struct {
```
Jeffwan, line 28: I feel this is very thin layer. If we use PodTemplate, do we still have control on the sidecar and engine?

1. should we inject initContainer or sidecar container inside the user defined template?
2. What if user uses `SGLang` in pod template but specify the vLLM as the engine type?

config/crd/bases/model.aibrix.ai_modeladapters.yaml
```diff
@@ -37,6 +37,7 @@ spec:
               credentialsSecretRef:
                 properties:
                   name:
+                    default: ""
```
Jeffwan: Seems https://github.com/aibrix/aibrix/commit/5aad38ccd06db1e0fb9caa7f96aad38dc288bf53 add the new field. not sure why it didn't show a conflict. could you rebase the main and have another check?

config/gateway/gateway-plugin.yaml
```diff
@@ -29,7 +29,7 @@ spec:
       initContainers:
         - name: init-c
           image: busybox
-          command: ['sh', '-c', "until echo \"ping\" | nc aibrix-redis-master 6379  | grep -c PONG; do echo waiting for service aibrix-redis-master; sleep 2; done"]
```
Jeffwan: If that's an issue, I suggest to cut a separate fix PR.

api/model/v1alpha1/model_types.go
varungup90: 1) If we have requirements then we should definitely inject our own init or sidecar containers.
2) Good point, need to add checks to prevent such disprepancy.

config/crd/bases/model.aibrix.ai_modeladapters.yaml
varungup90: Ok, I will rebase and check this.

config/gateway/gateway-plugin.yaml
varungup90: https://github.com/aibrix/aibrix/pull/309

Jeffwan:
StormService will be widely used in the latest inference architecture, and we don't plan to introduce new APIs at this point. StormService should be sufficient for current needs. Whether an additional model layer is necessary can be revisited in the future. This PR can be closed for now.

PR #342, opened 2024-10-29
Enable ipv6 for envoy proxy

varungup90:


config/overlays/vke/default/gateway/envoyproxy.yaml
```diff
@@ -2,11 +2,26 @@ apiVersion: gateway.envoyproxy.io/v1alpha1
 kind: EnvoyProxy
 metadata:
   name: custom-proxy-config
-  namespace: aibrix-system
```
Jeffwan: em. does it mean the previous configuration was wrong here?

varungup90: it should be aibrix-system only. Updating it. In my local also it is aibrix-system, not sure how it got pushed with different namespace.

PR #345, opened 2024-10-30
Add check if no pods are present

varungup90:


pkg/plugins/gateway/algorithms/least_request.go
```diff
@@ -77,5 +78,9 @@ func (r leastRequestRouter) Route(ctx context.Context, pods map[string]*v1.Pod)
 		}
 	}
 
+	if len(pods) == 0 || targetPodIP == "" {
```
Jeffwan: let's add some tests for the Route interface.

Jeffwan: for the pods list empty case, should we return earlier?

varungup90: Done.

benchmarks/gateway/7b.yaml
```diff
@@ -1,172 +0,0 @@
-apiVersion: apps/v1
```
Jeffwan: @varungup90 I see some benchmark/plot changes have been added in the PR. are they expected or rebase issue?

PR #347, opened 2024-10-31
Add reference grant to support httprouting for different namespace

varungup90:
Add support for cross-namespace routing: https://github.com/aibrix/aibrix/issues/336


pkg/controller/modelrouter/modelrouter_controller.go
```diff
@@ -165,6 +171,38 @@ func (m *ModelRouter) createHTTPRoute(namespace string, labels map[string]string
 		return
 	}
 	klog.Infof("httproute: %v created for model: %v", httpRoute.Name, modelName)
+
+	m.createReferenceGrant(modelName, namespace)
+}
+
+func (m *ModelRouter) createReferenceGrant(name, namespace string) {
+	referenceGrant := gatewayv1beta1.ReferenceGrant{
```
Jeffwan, line 185: a quick question, if we set namespace here, do we still need reference grant object?
image

varungup90: httproute will be created in aibrix namespace where gateway class is that where httproute object needs to. Service backend ref is in different namespace, hence the need to create reference grant.

Jeffwan: em. httproute has to be created in aibrix-system? we did create it in separate namespace right?

varungup90: previously I was using deployment namespace  for httproute but all the local testing was based on the deployment in the aibrix-system. httproute need to be in namespace as gateway class to be accepted.

Jeffwan: I point was If we add a namespace reference in gateway, it is supposed to work, right? I did workaround in downstream. can you check it? We should separate the control plane and data plane.

varungup90: For httproute to be accepted, there are two conditions. 1) httproute and gateway need to be in same namespace and 2) is httproute and backend service are in different namespace then a reference grant is needed.

Error if httproute and gateway are in different namespace.
img_v3_02ge_c66c95f9-964a-4042-9491-368323d5415h

Error if httproute and backend service are in different namespace and no reference grant is present.
img_v3_02ge_7632dbb8-8b1a-4601-801c-1e368eabd53h

Issue #334, opened 2024-10-27
http header -H "routing-strategy: "  empty case is not handled


Jeffwan:
### 🐛 Describe the bug

![image](https://github.com/user-attachments/assets/39b1da62-0591-43e9-bfe7-0a303d693fb5)

Issue #333, opened 2024-10-27
incorrect strategy name can be accepted


Jeffwan:
### 🐛 Describe the bug

user still have chance to type the wrong name, this should be checked for sure

![image](https://github.com/user-attachments/assets/e2425f82-2e70-4c98-8db6-ceb979e6c632)

![image](https://github.com/user-attachments/assets/8d0e210a-5f61-4845-93de-21acab827016)

PR #352, opened 2024-11-07
Add check for acceptable routing strategies

varungup90:
Validate routing strategy in inference request is correct: #333 #334

### DIFF
--- a/pkg/plugins/gateway/gateway.go
+++ b/pkg/plugins/gateway/gateway.go
@@ -21,6 +21,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


pkg/plugins/gateway/gateway.go
```diff
@@ -46,6 +47,7 @@ import (
 var (
 	defaultRPM           = 100
 	defaultTPMMultiplier = 1000
+	routingStrategies    = []string{"random", "least-request", "throughput"}
```
Jeffwan, line 50: is there source of truth or single constant we can use? I remember somewhere else have the definition

PR #364, opened 2024-11-11
Validate routing strategy bug fix

varungup90:
Address #333 #334

### DIFF
--- a/config/rbac/role.yaml
+++ b/config/rbac/role.yaml
@@ -135,6 +135,18 @@ rules:
   - update
   - watch
 - apiGroups:
+  - gateway.networking.k8s.io
+  resources:
+  - referencegrants
+  verbs:
+  - create
+  - delete
+  - get
+  - list
+  - patch
+  - update
+  - watch
+- apiGroups:
   - model.aibrix.ai
   resources:
   - modeladapters
--- a/pkg/controller/modelrouter/modelrouter_controller.go
+++ b/pkg/controller/modelrouter/modelrouter_controller.go
@@ -22,6 +22,7 @@ import (
 	"strconv"
 
 	appsv1 "k8s.io/api/apps/v1"
+	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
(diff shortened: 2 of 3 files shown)


pkg/plugins/gateway/gateway.go
```diff
@@ -154,14 +155,11 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, requestID string, req
 		}
 		if strings.ToLower(n.Key) == "routing-strategy" {
 			routingStrategy = string(n.RawValue)
+			routingStrategyEnabled = true
```
Jeffwan: great. beside enabling check here. if we set envs, we should enable routing as well?

pkg/plugins/gateway/gateway.go
```diff
@@ -480,6 +478,23 @@ func (s *Server) selectTargetPod(ctx context.Context, routingStrategy string, po
...
+	routingStrategy = strings.TrimSpace(routingStrategy)
+	if routingStrategyEnabled && routingStrategy == "" {
+		return false
+	}
+
+	if routingStrategy == "" {
+		routingStrategy = utils.GetEnv("ROUTING_ALGORITHM", "")
+	}
+
+	if routingStrategy != "" && !slices.Contains(routingStrategies, routingStrategy) {
```
Jeffwan, line 491: is there a chance ROUTING_ALGORITHM=""? or " ".

pkg/plugins/gateway/gateway.go
```diff
@@ -480,6 +478,23 @@ func (s *Server) selectTargetPod(ctx context.Context, routingStrategy string, po
...
 }
 
+func validateRoutingStrategy(routingStrategy string, routingStrategyEnabled bool) bool {
+	routingStrategy = strings.TrimSpace(routingStrategy)
+	if routingStrategyEnabled && routingStrategy == "" {
+		return false
+	}
+
+	if routingStrategy == "" {
+		routingStrategy = utils.GetEnv("ROUTING_ALGORITHM", "")
```
Jeffwan, line 488: if this case,  env setting override the request level setting?

varungup90: Request level has higher precedence. If request level is set then ln 487 if block will not hit.

pkg/plugins/gateway/gateway.go
varungup90: This is only for header based. I can rename the variable to routingStrategyHeaderEnabled, something like this.

pkg/plugins/gateway/gateway.go
varungup90: if routing_strategy is set at "" or " ", it will throw error at line 483.

pkg/plugins/gateway/gateway.go
Jeffwan: I mean `enabling routing check`. feel free to rename it or cover the env value check

pkg/plugins/gateway/gateway.go
Jeffwan: make sense

PR #372, opened 2024-11-12
replaced old names AIBricks with AIBrix

nwangfw:
Replaced all old project names from AIBricks to AIBrix

### DIFF
--- a/config/samples/orchestration_v1alpha1_rayclusterfleet.yaml
+++ b/config/samples/orchestration_v1alpha1_rayclusterfleet.yaml
@@ -9,7 +9,7 @@ spec:
   replicas: 1
   selector:
     matchLabels:
-      models.aibricks.ai: llama-2-7b-hf
+      model.aibrix.ai/name: llama-2-7b-hf
   strategy:
     rollingUpdate:
       maxSurge: 25%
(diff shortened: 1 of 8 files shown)


config/samples/orchestration_v1alpha1_rayclusterfleet.yaml
```diff
@@ -9,7 +9,7 @@ spec:
   replicas: 1
   selector:
     matchLabels:
-      models.aibricks.ai: llama-2-7b-hf
```
Jeffwan, line 12: I forgot to mention that, please update "models.aibricks.ai" -> "model.aibrix.ai/name"

config/samples/orchestration_v1alpha1_rayclusterfleet.yaml
```diff
@@ -18,7 +18,7 @@ spec:
   template:
     metadata:
       labels:
-        models.aibricks.ai: llama-2-7b-hf
+        models.aibrix.ai/name: llama-2-7b-hf
```
Jeffwan: it should be `model` instead of `models`, please help change all the references.

Jeffwan: @nwangfw

nwangfw: Thanks for your carefully check. I previously didn't notice this part.

PR #377, opened 2024-11-12
Add envoy gateway streaming support

varungup90:
Address https://github.com/aibrix/aibrix/issues/380

### DIFF
--- a/pkg/plugins/gateway/gateway.go
+++ b/pkg/plugins/gateway/gateway.go
@@ -223,7 +223,7 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 			"error processing request body"), targetPodIP, stream
 	}
 
-	if model, ok = jsonMap["model"].(string); !ok || model == "" || !s.cache.CheckModelExists(model) {
+	if model, ok = jsonMap["model"].(string); !ok || model == "" { // || !s.cache.CheckModelExists(model) # enable when dynamic lora is enabled
 		klog.ErrorS(nil, "model error in request", "requestID", requestID, "jsonMap", jsonMap)
 		return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError,
 			[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{


config/gateway/gateway.yaml
```diff
@@ -36,7 +36,8 @@ spec:
         request:
           body: Buffered
         response: 
-          body: Buffered
```
Jeffwan, line 39: If we configure stream here, it requires extension server to support `streamed` mode, right?

pkg/plugins/gateway/gateway.go
```diff
@@ -17,17 +17,20 @@ limitations under the License.
...
 	"encoding/json"
 	"fmt"
 	"io"
+	"net/http"
 	"slices"
 	"strings"
 	"time"
 
 	"github.com/google/uuid"
+	"github.com/openai/openai-go"
```
Jeffwan, line 31: good to switch to official sdk

pkg/plugins/gateway/gateway.go
```diff
@@ -120,13 +124,13 @@ func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
 			resp, user, rpm, routingStrategy = s.HandleRequestHeaders(ctx, requestID, req)
 
 		case *extProcPb.ProcessingRequest_RequestBody:
-			resp, targetPodIP = s.HandleRequestBody(ctx, requestID, req, user, routingStrategy)
+			resp, targetPodIP, stream = s.HandleRequestBody(ctx, requestID, req, user, routingStrategy)
```
Jeffwan, line 127: in the configuration in `config/gateway/gateway.yaml` we still use buffer for request, is this still consistent?

pkg/plugins/gateway/gateway.go
```diff
@@ -311,64 +337,97 @@ func (s *Server) HandleResponseHeaders(ctx context.Context, requestID string, re
...
+	var res openai.ChatCompletion
+	var model string
+	var usage openai.CompletionUsage
 	headers := []*configPb.HeaderValueOption{}
-	if user.Name != "" {
-		tpm, err := s.ratelimiter.Incr(ctx, fmt.Sprintf("%v_TPM_CURRENT", user), int64(res.Usage.TotalTokens))
-		if err != nil {
+
+	switch stream {
+	case true:
```
Jeffwan, line 350: seems it's binary logic, why not just use if..else..

pkg/plugins/gateway/gateway.go
```diff
@@ -311,64 +337,97 @@ func (s *Server) HandleResponseHeaders(ctx context.Context, requestID string, re
...
+				s.cache.AddRequestTrace(model, usage.PromptTokens, usage.CompletionTokens)
+			}()
+		}()
 
-		headers = append(headers,
-			&configPb.HeaderValueOption{
-				Header: &configPb.HeaderValue{
-					Key:      "x-update-rpm",
-					RawValue: []byte(fmt.Sprintf("%d", rpm)),
+		if user.Name != "" {
```
Jeffwan, line 392: nit: in future, we can add some post-processor interface here. I think update rate limit would be one of them. This makes the code cleaner.

Jeffwan: we do not need to refactor it in this PR

pkg/plugins/gateway/gateway.go
```diff
@@ -203,25 +207,46 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, requestID string, req
...
+	if model, ok = jsonMap["model"].(string); !ok || model == "" { // || !s.cache.CheckModelExists(model) # enable when dynamic lora is enabled
+		klog.ErrorS(nil, "model error in request", "requestID", requestID, "jsonMap", jsonMap)
 		return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError,
 			[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{
 				Key: "x-no-model", RawValue: []byte(model)}}},
-			fmt.Sprintf("no model in request body or model %s does not exist", model)), targetPodIP
+			fmt.Sprintf("no model in request body or model %s does not exist", model)), targetPodIP, stream
+	}
+
+	stream, ok = jsonMap["stream"].(bool)
```
Jeffwan, line 234: Here, it requires user to specify a header for stream usage `x-stream-options`? can it be compatible with openai stream sdk?  does openai-sdk automatically register this header?

config/gateway/gateway.yaml
varungup90: yes

pkg/plugins/gateway/gateway.go
varungup90: Yes, because for request data is not streamed.

pkg/plugins/gateway/gateway.go
varungup90: Yes, I am working on refactor PR which will do this and other clean up for gateway.

pkg/plugins/gateway/gateway.go
varungup90: stream header from open-ai SDK.

image

PR #395, opened 2024-11-15
Add client traffic policy to increase per connection buffer size from 32kb to 256kb

varungup90:


config/gateway/gateway.yaml
```diff
@@ -19,6 +19,19 @@ spec:
       port: 80   
 ---
 apiVersion: gateway.envoyproxy.io/v1alpha1
+kind: ClientTrafficPolicy
+metadata:
+  name: client-timeout
```
Jeffwan: should it be some other names? the major configuration is for bufferLimit

varungup90: done

PR #398, opened 2024-11-15
[CI] improve the ci efficiency by parallelizing the build tasks

nwangfw:
## Pull Request Description

In the initial version, we do not care the efficiency that much and we use single runner to build the images one by one. 

In this PR, we create four runners to build images in parallel and then use another runner to conduct tests.

## Related Issues
Resolves: #[128]

### DIFF
--- a/.github/workflows/installation-tests.yml
+++ b/.github/workflows/installation-tests.yml
@@ -7,14 +7,42 @@ on:
     branches: [ "main" ]
 
 jobs:
-  installation-test:
+  build-images:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        image: [controller-manager, plugins, runtime, users]
     steps:
       - name: Check out code
         uses: actions/checkout@v4
 
       - name: Set up Docker Buildx
         uses: docker/setup-buildx-action@v2
+
+      - name: Build ${{ matrix.image }}
+        run: |
+          GIT_COMMIT_HASH=${{ github.sha }} make docker-build-${{ matrix.image }}
+
+      - name: Save image
+        run: |
+          docker save aibrix/${{ matrix.image }}:${{ github.sha }} > ${{ matrix.image }}.tar
+
+      - name: Upload image artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ matrix.image }}-image
+          path: ${{ matrix.image }}.tar
+          retention-days: 1
+
+  installation-test:
+    needs: build-images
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -133,7 +133,8 @@ define push_image
 endef
 
 .PHONY: docker-build-all
-docker-build-all: docker-build-controller-manager docker-build-plugins docker-build-runtime docker-build-users ## Build all docker images
+docker-build-all: ## Build all docker images in parallel
```
Jeffwan: did you try to use `make -j4` here?

https://github.com/aibrix/aibrix/blob/b0413e9ca4a1e912b783b346d1be0a04f187f54b/.github/workflows/installation-tests.yml#L37-L39

Makefile
```diff
@@ -133,7 +133,8 @@ define push_image
...
+	@$(MAKE) -j4 docker-build-controller-manager docker-build-plugins docker-build-runtime docker-build-users
```
Jeffwan: do we still need this change?

.github/workflows/installation-tests.yml
```diff
@@ -7,15 +7,43 @@ on:
...
+          docker save aibrix/${{ matrix.image }}:${{ github.sha }} > ${{ matrix.image }}.tar
+
+      - name: Upload image artifact
+        uses: actions/upload-artifact@v3
+        with:
+          name: ${{ matrix.image }}-image
+          path: ${{ matrix.image }}.tar
+          retention-days: 1
+
+  installation-test:
```
Jeffwan, line 37: em. are those images not visible to further loading process? is artifact upload required?

Makefile
nwangfw: It doesn't speed up Github Action tests. I am thinking that we can speed up local testing with this change. Let me know if it is better to keep it or discard this change.

Jeffwan: if it doesn't help on the github actions. let's discard this change at this moment.  user should use `make -j4 docker-build-all` instead

nwangfw: Make sense. Will change it back then.

.github/workflows/installation-tests.yml
nwangfw: In the previous version, we only used 1 runner and we don't need upload and download steps. However, we use 5 runners in total (the first 4 runners to build images and the last runner to run the installation test) and that's why we need upload process for the first four runners and the download process for the last runner.
image

PR #411, opened 2024-11-20
[CI] Github Action trigger condition optimized for cost saving

nwangfw:
## Pull Request Description
Github Action trigger condition is optimized. Now each test will only be triggered when the corresponding code is modified.

## Related Issues
Resolves: https://github.com/aibrix/aibrix/issues/253

### DIFF
--- a/.github/workflows/python-tests.yml
+++ b/.github/workflows/python-tests.yml
@@ -3,9 +3,12 @@ name: Python Tests
 on:
   push:
     branches: [ "main" ]
+    paths:
+      - 'python/**'
   pull_request:
     branches: [ "main" ]
-
+    paths:
+      - 'python/**'
 jobs:
   lint:
     runs-on: ubuntu-latest
--- a/.github/workflows/release-build.yaml
+++ b/.github/workflows/release-build.yaml
@@ -1,6 +1,7 @@
 name: Release Build and Push
 
 on:
+  workflow_dispatch:  # Allows manual trigger
   push:
     tags:
       - 'v*'  # This triggers the workflow on any new tag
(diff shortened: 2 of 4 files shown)


.github/workflows/release-build.yaml
```diff
@@ -1,6 +1,7 @@
 name: Release Build and Push
 
 on:
+  workflow_dispatch:  # Allows manual trigger
```
Jeffwan, line 4: this makes sense

.github/workflows/python-tests.yml
```diff
@@ -1,11 +1,15 @@
 name: Python Tests
 
 on:
+  workflow_dispatch:  # Allows manual trigger
```
Jeffwan: do we want manual trigger here?

.github/workflows/docker-push-images.yml
```diff
@@ -1,8 +1,20 @@
...
+      - 'Dockerfile'
+      - 'build/**'
+      - 'pkg/**'
+      - 'cmd/**'
+      - 'python/**'
+      - 'Makefile'
+      - 'go.mod'
+      - 'go.sum'
+    tags:
+      - 'v*'
```
Jeffwan: we don't push to tags (they are protected). This is something unnecessary I think.

.github/workflows/python-tests.yml
nwangfw: deleted

.github/workflows/docker-push-images.yml
nwangfw: deleted

Issue #456, opened 2024-12-02
 Consolidate mocked app and simulator into one unified mock app for development


Jeffwan:
### 🚀 Feature Description and Motivation

We previously built a CPU-mocked app to facilitate fast prototyping and local development. This approach was convenient as it eliminated the need for GPU resources and simulated OpenAI-like interfaces and metrics effectively.

Recently, [PR #430](https://github.com/aibrix/aibrix/pull/430) introduced a separate app for simulation purposes. While it adds value, having separate apps for similar functionalities increases complexity and maintenance overhead.

### Use Case

We need to build an unified app, merge all features together for fast prototyping. 

- Streamline the development process.
- Reduce complexity for local development environments.
- Ensure consistent feature sets and interfaces.

PR #477, opened 2024-12-04
[Misc] Consolidate app and simulator

zhangjyr:
## Pull Request Description
We previously built a CPU-mocked app to facilitate fast prototyping and local development. This approach was convenient as it eliminated the need for GPU resources and simulated OpenAI-like interfaces and metrics effectively.

Recently, https://github.com/aibrix/aibrix/pull/430 introduced a separate app for simulation purposes. While it adds value, having separate apps for similar functionalities increases complexity and maintenance overhead.

This PR consolidates the simulator to the previous mock app. Since the simulator is GPU-specific, a separate base image is needed for the simulator of various GPUs. Currently, the default GPU is a100. Examples of config other GPUs are included.

## Related Issues
Resolves: #456 

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>

### DIFF
--- a/development/app/Dockerfile
+++ b/development/app/Dockerfile
@@ -21,16 +21,17 @@ RUN pip install --no-cache-dir -r requirements.txt
 COPY ./*.py /app/
 
 ENV MODEL_NAME=llama2-7b
-ARG GPU_TYPE=disabled
+ENV HUGGINGFACE_TOKEN="your huggingface token"
+ARG SIMULATION=disabled
 
 # Trigger profiling
-RUN if [ "$GPU_TYPE" != "disabled" ]; then \
-        python app.py --time_limit 1000 --replica_config_device ${GPU_TYPE}; \
+RUN if [ "$SIMULATION" != "disabled" ]; then \
+        python app.py --time_limit 1000 --replica_config_device ${SIMULATION}; \
     fi
 
 # Expose the port the app runs on
 EXPOSE 8000
 
 # Run the application, environment variable is necessary to apply ARG
-ENV GPU_TYPE=$GPU_TYPE
-CMD python app.py --replica_config_device ${GPU_TYPE}
+ENV SIMULATION=$SIMULATION
+CMD python app.py --replica_config_device ${SIMULATION}
--- a/development/app/app.py
+++ b/development/app/app.py
@@ -26,6 +26,7 @@ MODEL_NAME = os.getenv('MODEL_NAME', 'llama2-70b')
 DEPLOYMENT_NAME = os.getenv('DEPLOYMENT_NAME', 'llama2-70b')
 NAMESPACE = os.getenv('POD_NAMESPACE', 'default')
 DEFAULT_REPLICAS = int(os.getenv('DEFAULT_REPLICAS', '1'))
+HUGGINGFACE_TOKEN = os.getenv('HUGGINGFACE_TOKEN', "your huggingface token")
 
 modelMaps = {
     "llama2-7b": "meta-llama/Llama-2-7b-hf",
@@ -589,10 +590,20 @@ if __name__ == '__main__':
     if gpu_device != "disabled":
         # Load the tokenizer for your model
         from transformers import AutoTokenizer
-        tokenizer = AutoTokenizer.from_pretrained(
-            'bert-base-uncased',
-            model_max_length=16384, # Suppress warning
-            clean_up_tokenization_spaces=True)
+        default_model = 'bert-base-uncased'
+        try:
+            token_model = modelMaps.get(MODEL_NAME, default_model)
+            tokenizer = AutoTokenizer.from_pretrained(
+                token_model,
+                token=HUGGINGFACE_TOKEN,
+                model_max_length=16384, # Suppress warning
+                clean_up_tokenization_spaces=True)
+        except Exception as e:
+            logger.error(f"Failed to initialize tokenizer, will use default tokenizer model: {e}")
+            tokenizer = AutoTokenizer.from_pretrained(
+                default_model,
+                model_max_length=16384, # Suppress warning
+                clean_up_tokenization_spaces=True)
 
         simulator = Simulator(SimulationConfig.create_from_cli_args())
         overrides = {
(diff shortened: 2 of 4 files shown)


development/app/Dockerfile
```diff
@@ -12,15 +13,24 @@ WORKDIR /app
...
+RUN apt update && apt install -y curl jq git
 
 RUN pip install --no-cache-dir -r requirements.txt
 
 # Copy the rest of the application code into the container
-COPY . /app/
+COPY ./*.py /app/
+
+ENV MODEL_NAME=llama2-7b
+ARG GPU_TYPE=disabled
```
Jeffwan: minor: I'd rather to have two args, profiler enabled/disabled and gpu type. 
if profile == enabled, gpu_types should exist.

This makes the code more readable for people who is not familiar with this feature

development/app/app.py
```diff
@@ -481,11 +567,57 @@ def metrics():
...
+    # Restore -h functionality
+    if '-h' in sys.argv:
+        SimulationConfig.create_from_cli_args()
+
+    # Launch simulator
+    if gpu_device != "disabled":
+        # Load the tokenizer for your model
+        from transformers import AutoTokenizer
+        tokenizer = AutoTokenizer.from_pretrained(
+            'bert-base-uncased',
```
Jeffwan: Let's use the right model tokenizer for more accurate result?

development/app/config/templates/podautoscaler/podautoscaler.yaml
```diff
@@ -0,0 +1,23 @@
...
+    app.kubernetes.io/managed-by: kustomize
+  namespace: default
+spec:
+  scaleTargetRef:
+    apiVersion: apps/v1
+    kind: Deployment
+    name: mock-llama2-7b
+  minReplicas: 0
+  maxReplicas: 10
+  targetMetric: "avg_prompt_throughput_toks_per_s" # Ignore if metricsSources is configured
```
Jeffwan, line 17: minor and not important at this moment: should we consider to make `targetMetrics` and `metricsSource` compatible? the external metrics service should follow some standards to exposes the metrics and metric fetcher can construct the query using the source and target metrics

development/app/config/heterogeneous/kustomization.yaml
```diff
@@ -0,0 +1,7 @@
+kind: Kustomization
+
+resources:
+- ../simulator
+- simulator_a40
```
Jeffwan, line 5: does it create one PA object or two PA object here?

development/app/config/templates/podautoscaler/podautoscaler.yaml
zhangjyr: The metricsSource has a metric field to specify its own metric. Without changing the spec, I think the logic should be:
1. Keep "targetMetric" required since it is required when no "metricSource" is specified.
2. If "metricSource" is available and "metric" of "metricSource" is specified, the metric overrides "targetMetric" (current implementation)
3. If the "metric" of "metricSource" is not specified. Then "targetMetric" should be used. (not implemented)
We can start an issue with this change if it's okay.

development/app/config/heterogeneous/kustomization.yaml
zhangjyr: Two PA objects are created for a100 and a40, respectively.

development/app/Dockerfile
zhangjyr: I can add another bool switch here. Although the profiling necessity is the reason for separate images, this switch will also enable/disable the simulation feature. So, I will name the switch as "simulation."

zhangjyr: Things get a little complicated as:
1. The vidur argparse overrides mock app argument parsing. So, no new argument can be added with the current implementation. This means even if I added a new ARG for Dockerfile, the mock app will still use one argument.
2. The Dockerfile does not have complex condition logic. I currently use ENV to pass ARG to CMD; the difference between Dockerfile BUILD-ARGs and mock app arguments will be error-prone.

development/app/app.py
zhangjyr: For this part, I find meta-llama/Llama-2-7b-hf is gated, and an authentication token is required. In this case, we will need to guide users to gain authentication for the code to work, or the model will fall back to 'bert-base-uncased'. Will that work?

Jeffwan: got you. I think we should use other models like qwen or phi instead in future. since here is just for testing purpose.I am ok with it.

development/app/config/templates/podautoscaler/podautoscaler.yaml
Jeffwan: make sense. I kind of forgot why we added `metric`  inside `metricSource` there, kind of duplication.  I create an issue https://github.com/aibrix/aibrix/issues/482 to track this and we do not need to discuss this in this thread. it looks good to me

development/app/config/heterogeneous/kustomization.yaml
Jeffwan: We need to discuss the feature gate to enable the autoscaling feature hetegenous. From ux perspective, user may like to create one at model level.

development/app/Dockerfile
Jeffwan: then let's stick to one argument at this moment and rename to `simulation`, that looks good to me

development/app/config/heterogeneous/kustomization.yaml
zhangjyr: Agree. It should be doable

development/app/app.py
zhangjyr: Done. HUGGINGFACE_TOKEN is added to the environment for users to configure.

development/app/Dockerfile
zhangjyr: Done

development/app/app.py
zhangjyr: Dockerfile complains about saving tokens in ENV. I'll fix this.

PR #478, opened 2024-12-04
Refactor generator to generate time-based traces

happyandslow:
This commit adds 
- Leading timestamp for each concurrent dispatch
- Force include input/output token lengths in the dataset
- Option to store as .jsonl


benchmarks/generator/workload_generator.py
```diff
@@ -180,64 +180,64 @@ def generate_from_azure_csv(file_path: str,
...
-    parser.add_argument('--num-requests', type=int, default=10000, help='Number of requests in total')
-    parser.add_argument('--group-interval-seconds', type=int, default=1, help='Grouping interval seconds')
-    parser.add_argument('--trace-type', type=str, required=True, default="synthetic", help='Type of trace consumed')
-    parser.add_argument('--trace-file', type=str, required=False, default=None, help='File containing trace CSV')
-    parser.add_argument('--model', type=str, required=False, default="meta-llama/Llama-2-7b-hf", help='Target model tokenizer')
-    parser.add_argument('--output', type=str, required=False, default="output/output", help='Output path')
+    parser.add_argument('--prompt-file', type=str, required=True, help='File containing prompts.')
+    parser.add_argument('--num-prompts', type=int, default=100, help='Number of prompts to sample.')
+    parser.add_argument('--group-interval-seconds', type=int, default=1, help='Grouping interval seconds.')
+    parser.add_argument('--trace-type', type=str, required=True, default="synthetic", help='Type of trace consumed.')
```
Jeffwan: I suggest to have all `--trace-type` listed here.

benchmarks/generator/workload_generator.py
```diff
@@ -180,64 +180,64 @@ def generate_from_azure_csv(file_path: str,
...
+    parser.add_argument('--trace-file', type=str, required=False, default=None, help='File containing trace CSV.')
```
Jeffwan: Let's add more details in the `help`.  what kind of trace-type needs such file etc.

benchmarks/generator/workload_generator.py
```diff
@@ -180,64 +180,64 @@ def generate_from_azure_csv(file_path: str,
...
+    parser.add_argument('--model', type=str, required=False, default="meta-llama/Llama-2-7b-hf", help='Target model tokenizer.')
```
Jeffwan: let's change the default one to something else. like https://huggingface.co/Qwen/Qwen2.5-Coder-7B-Instruct. llama7b can not be publicly accessible

benchmarks/generator/workload_generator.py
```diff
@@ -180,64 +180,64 @@ def generate_from_azure_csv(file_path: str,
...
+    parser.add_argument('--output', type=str, required=False, default="output/output", help='Output path to the workload.')
+    parser.add_argument('--interval-ms', type=int, required=False, default=1000, help='Granularity of request injection interval in milliseconds.')
+    parser.add_argument('--duration-ms', type=int, default=60000, help='Duration of the trace generated.')
```
Jeffwan, line 213: suggest to use `second` as the time unit. for duration, normally, we at least generate 120-600 etc. won't be any case less than 1s

benchmarks/generator/workload_generator.py
```diff
@@ -180,64 +180,64 @@ def generate_from_azure_csv(file_path: str,
...
-    parser.add_argument('--trace-file', type=str, required=False, default=None, help='File containing trace CSV')
-    parser.add_argument('--model', type=str, required=False, default="meta-llama/Llama-2-7b-hf", help='Target model tokenizer')
-    parser.add_argument('--output', type=str, required=False, default="output/output", help='Output path')
+    parser.add_argument('--prompt-file', type=str, required=True, help='File containing prompts.')
+    parser.add_argument('--num-prompts', type=int, default=100, help='Number of prompts to sample.')
+    parser.add_argument('--group-interval-seconds', type=int, default=1, help='Grouping interval seconds.')
+    parser.add_argument('--trace-type', type=str, required=True, default="synthetic", help='Type of trace consumed.')
+    parser.add_argument('--trace-file', type=str, required=False, default=None, help='File containing trace CSV.')
+    parser.add_argument('--model', type=str, required=False, default="meta-llama/Llama-2-7b-hf", help='Target model tokenizer.')
+    parser.add_argument('--output', type=str, required=False, default="output/output", help='Output path to the workload.')
```
Jeffwan: `output/output` -> `output`?

benchmarks/generator/workload_generator.py
happyandslow: Fixed in new commit.

benchmarks/generator/workload_generator.py
happyandslow: Fixed in new commit.

benchmarks/generator/workload_generator.py
happyandslow: Fixed in new commit.

benchmarks/generator/workload_generator.py
happyandslow: Fixed in new commit.

PR #549, opened 2025-01-02
Handle terminating pod for request routing

varungup90:
https://github.com/aibrix/aibrix/issues/541

### DIFF
--- a/pkg/plugins/gateway/gateway.go
+++ b/pkg/plugins/gateway/gateway.go
@@ -243,6 +243,15 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 			fmt.Sprintf("model %s does not exist", model)), targetPodIP, stream
 	}
 
+	// early reject if no pods are ready to accept request for a model
+	pods, err := s.cache.GetPodsForModel(model)
+	if len(pods) == 0 || len(utils.FilterReadyPods(pods)) == 0 || err != nil {
+		return generateErrorResponse(envoyTypePb.StatusCode_ServiceUnavailable,
+			[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{
+				Key: "x-no-model-deployment", RawValue: []byte("true")}}},
+			fmt.Sprintf("error on getting pods for model %s", model)), targetPodIP, stream
+	}
+
 	stream, ok = jsonMap["stream"].(bool)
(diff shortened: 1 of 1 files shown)


pkg/utils/pod.go
```diff
@@ -144,9 +144,12 @@ func CountReadyPods(podList *v1.PodList) (int64, error) {
 func FilterReadyPods(pods map[string]*v1.Pod) []*v1.Pod {
 	var readyPods []*v1.Pod
 	for _, pod := range pods {
-		if pod.Status.PodIP != "" && IsPodReady(pod) {
```
Jeffwan, line 147: is this method being used by other files? like some other controllers other than cache.go? I am a little bit concerned on the impact if we have other scenarios. 

the key difference of this change is `terminated` pod?

pkg/plugins/gateway/gateway.go
```diff
@@ -243,6 +243,15 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 			fmt.Sprintf("model %s does not exist", model)), targetPodIP, stream
 	}
 
+	// early reject if no pods are ready to accept request for a model
+	pods, err := s.cache.GetPodsForModel(model)
```
Jeffwan, line 247: it filters the terminated pods here?

pkg/utils/pod.go
varungup90: This is used by all routing methods, updatePodMetrics and early rejection for gateway. No impact with the change. 

I also thought about it, and not including terminating pods from ready pods list makes sense, hence I updated this method rather than introducing another helper method.

pkg/plugins/gateway/gateway.go
varungup90: Here, it returns all the pods irrespective of the pod state.

pkg/utils/pod.go
Jeffwan: I mean other controllers other than routing. I remember some ray controllers use it so I create the util file earlier. If so, other logics would be affected. we may consider to create another method if there's references

varungup90: No controllers use this helper method.

image

Jeffwan: Great. then we are safe to modify it.

PR #554, opened 2025-01-08
Gateway plugin report total incoming requests and pending requests

zhangjyr:
## Pull Request Description
1. Upgrade request trace version to 3 that includes two more meta info: "meta_total_reqs" and "meta_pending_reqs."
While meta_pending_reqs ideally equals pending requests out of the current report window, which is:
```
meta_pending_reqs = total pending request - current pending requests
```
The counting of counting requests in the current report window (current pending requests) relies on some kind of request membership in the new created RequestTrace struct, either using term-based or requestID-based.  Currently, no request membership is implemented for "current pending requests," so "meta_pending_reqs" is practically smaller than defined.

2.  Improve lock efficiency by (1) Creating a new RequestTrace struct for trace level locking. (2) Minimizing the cache level locking block. (3) Adding test cases to ensure reporting accuracy (no request missing) during trace window transition.

3. Align the trace report time to the report interval with a max offset of 500ms. Previously, trace reports of tick 30s can be written to the Redis during [30s, 40s) depending on when the gateway plugin starts. This PR fixed this indeterministicity by waiting a short period at the gateway plugin starting. Now, reports will be written to Redis at tick (30s, 30.5s].

4. Resolved a deadlock problem by calling locked debuginfo() in other locked functions().

5. Fixed a bug that the gateway plugin can't handle error responses from the vllm and responded with a misunderstanding message.

## Related Issues

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>

### DIFF
--- a/pkg/plugins/gateway/gateway.go
+++ b/pkg/plugins/gateway/gateway.go
@@ -53,7 +53,7 @@ var (
 	defaultTPMMultiplier = 1000
 	routingStrategies    = []string{"random", "least-request", "throughput", "least-kv-cache", "least-busy-time", "least-latency"}
 
-	ErrorUnknownResponse = errors.New("unknonw response")
+	ErrorUnknownResponse = errors.New("unknown response")
 )
 
 type Server struct {
@@ -236,7 +236,7 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 		return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError,
 			[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{
 				Key: "x-no-model", RawValue: []byte(model)}}},
-			fmt.Sprintf("no model in request body or model %s does not exist", model)), model, targetPodIP, stream, term
+			"no model in request body"), model, targetPodIP, stream, term
 	}
 
 	// early reject the request if model doesn't exist.
@@ -266,7 +266,7 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 			return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError,
 				[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{
 					Key: "x-stream-options", RawValue: []byte("stream options not set")}}},
-				"error processing request body"), model, targetPodIP, stream, term
+				"no stream option available"), model, targetPodIP, stream, term
 		}
 		includeUsage, ok := streamOptions["include_usage"].(bool)
 		if !includeUsage || !ok {
@@ -274,7 +274,7 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 			return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError,
 				[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{
 					Key: "x-stream-options-include-usage", RawValue: []byte("include usage for stream options not set")}}},
-				"error processing request body"), model, targetPodIP, stream, term
+				"no stream with usage option available"), model, targetPodIP, stream, term
 		}
 	}
 
@@ -369,7 +369,7 @@ func (s *Server) HandleResponseBody(ctx context.Context, requestID string, req *
 	var usage openai.CompletionUsage
 	var promptTokens, completionTokens int64
 	headers := []*configPb.HeaderValueOption{}
-	complete := false
+	complete := false || hasCompleted
 
 	defer func() {
 		// Wrapped in a function to delay the evaluation of parameters. Using complete to make sure DoneRequestTrace only call once for a request.


pkg/plugins/gateway/gateway.go
```diff
@@ -246,6 +246,7 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 	// early reject if no pods are ready to accept request for a model
 	pods, err := s.cache.GetPodsForModel(model)
 	if len(pods) == 0 || len(utils.FilterReadyPods(pods)) == 0 || err != nil {
+		klog.ErrorS(err, "no ready pod available", "requestID", requestID, "jsonMap", jsonMap)
```
varungup90: jsonMap is not needed here in logging.

pkg/plugins/gateway/gateway.go
```diff
@@ -305,6 +309,8 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 		klog.InfoS("request start", "requestID", requestID, "model", model, "routingStrategy", routingStrategy, "targetPodIP", targetPodIP)
 	}
 
+	s.cache.AddRequestCount(requestID, model)
```
varungup90: add this in defer block.

pkg/cache/requestTrace.go
```diff
@@ -0,0 +1,165 @@
...
+)
+
+var requestTraceMetaKeys = [...]string{"meta_v", "meta_interval_sec", "meta_precision", "meta_total_reqs", "meta_pending_reqs", "meta_len"}
+
+func (key RequestTraceMetaKey) ToString() string {
+	return requestTraceMetaKeys[key]
+}
+
+const (
+	RequestTraceVersion           = 3
```
varungup90: can you add description for each const.

pkg/cache/requestTrace.go
```diff
@@ -0,0 +1,165 @@
...
+	numRequests     int32     // Total requests seen in the trace window
+	pendingRequests int32     // Total pending requests remain in the trace window
+
+	mu       sync.RWMutex
+	recycler func(any)
+}
+
+// Increase request counting and return the trace term, key is ignored for now.
+func (t *RequestTrace) AddRequest(requestID string, key string) bool {
+	t.mu.RLock()
```
varungup90: it will be write lock.

pkg/cache/requestTrace.go
```diff
@@ -0,0 +1,165 @@
...
+// keyPendingRequests                    = "meta_pending_reqs"
+
+type RequestTrace struct {
+	trace           *sync.Map // map[Log2(input_token):Log2(output_token)]request_count
+	numKeys         int32     // The number of keys in the trace.
+	numRequests     int32     // Total requests seen in the trace window
+	pendingRequests int32     // Total pending requests remain in the trace window
+
+	mu       sync.RWMutex
+	recycler func(any)
```
varungup90: add description for recycler.

pkg/plugins/gateway/gateway.go
zhangjyr: I reviewed the logs and included the necessary info. Previously, it was quickly added and just used for debugging a deadlock bug. The bug is fixed in the new commit.

pkg/cache/requestTrace.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (t *RequestTrace) Lock() {
+	t.mu.Lock()
+}
+
+func (t *RequestTrace) Unlock() {
+	t.mu.Unlock()
+}
+
+func (t *RequestTrace) ToMapLocked(total_pending int32) map[string]int {
```
varungup90: to prevent deadlocks in future, follow this naming convention.

First letter capital of function. will have locks. First letter of function name small wont have lock.

pkg/plugins/gateway/gateway.go
zhangjyr: As we discussed earlier, only requests sent to pods will be counted. Any error that results in the request being discarded will not be counted here. Does that make sense?

pkg/cache/requestTrace.go
zhangjyr: It should not cause a write lock. I have a unit test specifically for ReqeustTrace.

pkg/cache/requestTrace.go
```diff
@@ -0,0 +1,165 @@
...
+		return false
+	}
+	atomic.AddInt32(&t.numRequests, 1)
+	atomic.AddInt32(&t.pendingRequests, 1)
+	return true
+}
+
+// Decrease request counting with term verification
+func (t *RequestTrace) DoneRequest(requestID string, key string) bool {
+	t.mu.RLock()
```
varungup90: You are putting a read lock and then using atomic function to change variables.

There are few problems with this approach.
1) Two functions with read lock can proceed together. If two functions are trying to update same variables, there wont be data race but it can lead to inconsistency.
2) Each function implementor need to be careful by adding atomic. If the function implementation becomes complex, maintainability will be hard.
3) I think you are doing it to improve performance, do you have data which can support having read lock + atomic will be faster than write lock.

pkg/cache/cache.go
```diff
@@ -353,11 +370,11 @@ func (c *Cache) addModelAdapter(obj interface{}) {
 
 	model := obj.(*modelv1alpha1.ModelAdapter)
 	for _, pod := range model.Status.Instances {
-		c.addPodAndModelMapping(pod, model.Name)
+		c.addPodAndModelMappingLocked(pod, model.Name)
 	}
 
 	klog.V(4).Infof("MODELADAPTER CREATED: %s/%s", model.Namespace, model.Name)
-	c.debugInfo()
+	c.debugInfoLocked()
```
varungup90, line 377: no need to add word locked. Lets using function naming convention to make it simple.

First letter capital will have lock and first letter small will be a helper function without lock.

There are couple of updatePodMetrics which are not following this convention, you can clean it up in this PR or I can do it later.

pkg/cache/requestTrace.go
zhangjyr: The first letter, capital or not, decides if the function is private or public. Using this convention for locking is confusing. Following the name convention in the Golang source code, a function that should be locked to be thread-safe but would not acquire a lock will have a postfix "-Locked"
In the case of RequestTrace, I provided both acquiring lock and non-acquiring lock versions of ToMap, for I need the later version to implement batching operations in a lock.

pkg/cache/requestTrace.go
zhangjyr: 1. data inconsistency is not a problem here, because the write operation is guarded by a write lock and will wait for read-locked blocks run to the end.
2.  The RequestTrace is carefully planned for efficiency. Let's not over-design for unforeseen situations for now.
3. I don't have direct data to compare the composition of read lock + atomic vs write lock. The logic behind the performance improvement lies in: (1) atomic operation is faster than mutex lock, (2) concurrent read lock is faster than series mutex lock. We can add a benchmark in the folder to compare these two versions later.

pkg/cache/requestTrace.go
zhangjyr: Sure, comments are added.

pkg/cache/requestTrace.go
zhangjyr: Added

pkg/cache/cache.go
zhangjyr: Yes, let's discuss the convention separately and reach an agreement.

pkg/cache/requestTrace.go
varungup90: Can you benchmark AddRequest and DoneRequest, read lock + atomic and simple write lock.

pkg/cache/requestTrace.go
```diff
@@ -0,0 +1,165 @@
...
+
+func (key RequestTraceMetaKey) ToString() string {
+	return requestTraceMetaKeys[key]
+}
+
+const (
+	// The version of request trace, verison history:
+	// v1: No meta, default
+	// v2: Added meta data include version(meta_v), bucket precision(meta_precision), and interval(meta_interval_sec) to notify client the trace interval.
+	// v3: Added the number of total requests(meta_total_reqs) and pending requests(meta_pending_reqs) for out of window uncomplted requests.
```
varungup90: nit: uncompleted spelling.

pkg/cache/cache.go
```diff
@@ -251,19 +247,40 @@ func NewCache(config *rest.Config, stopCh <-chan struct{}, redisClient *redis.Cl
...
 			}
+			if traceAlignmentTimer != nil {
+				// Wait for alignment
+				<-traceAlignmentTimer.C
+				traceAlignmentTimer = nil
+
+				// TODO: Clean up data if necessary.
+
+				// Start ticker
+				traceTicker = time.NewTicker(RequestTraceWriteInterval)
```
varungup90, line 273: add traceticker.Stop on <-stopCh

pkg/plugins/gateway/gateway.go
varungup90: Add a defer block at line 312 itself. Reason being, let the function complete so that request can be forwarded to inference engine and adding request trace can be done in the background. 

Similar to this
image

pkg/cache/requestTrace.go
```diff
@@ -0,0 +1,181 @@
+/*
```
Jeffwan: requestTrace -> request_trace.go?

pkg/cache/requestTrace.go
zhangjyr: Added, see cache_test.go. The results are as follows:
BenchmarkLagacyAddRequestTrace-10        3118348               392.7 ns/op
BenchmarkAddRequest-10                            6628948               200.3 ns/op
BenchmarkDoneRequest-10                         12630597                95.97 ns/op
BenchmarkAddRequestTrace-10                    5317645               215.6 ns/op
BenchmarkDoneRequestTrace-10                 4552640               268.8 ns/op

pkg/plugins/gateway/gateway.go
zhangjyr: As explained, AddRequestCount is not supposed to be called every time except for requests will be sent to the vLLM. Using defer will count all requests. As for starting a go routine for it, first, I think a 200ns operation is quick enough to be called directly, while the overhead of starting a new goroutine in this case can not be ignored. If 200ns did matter, I would suggest starting a single server go routine dedicated for the operation, and use a channel for communication.

pkg/cache/requestTrace.go
zhangjyr: Fixed

pkg/cache/cache.go
zhangjyr: Fixed

pkg/cache/requestTrace.go
zhangjyr: Fixed

pkg/cache/cache.go
```diff
@@ -803,47 +823,123 @@ func (c *Cache) updateModelMetrics() {
 	}
 }
 
-func (c *Cache) AddRequestTrace(modelName string, inputTokens, outputTokens int64) {
-	c.mu.Lock()
-	defer c.mu.Unlock()
+func (c *Cache) AddRequestCount(requestID string, modelName string) (traceTerm int64) {
+	success := false
+	for {
```
varungup90, line 828: why is for loop required here?

zhangjyr: During writing data to the Redis, the current RequestTrace object may be recycled on getting Rlock and failed, so a for loop is required for the retry purpose.

pkg/plugins/gateway/gateway.go
```diff
@@ -51,6 +52,8 @@ var (
 	defaultRPM           = 100
 	defaultTPMMultiplier = 1000
 	routingStrategies    = []string{"random", "least-request", "throughput", "least-kv-cache", "least-busy-time", "least-latency"}
+
+	ErrorUnknownResponse = errors.New("unknonw response")
```
varungup90: unknown spelling.

pkg/plugins/gateway/gateway.go
```diff
@@ -350,14 +361,22 @@ func (s *Server) HandleResponseHeaders(ctx context.Context, requestID string, re
...
 	var res openai.ChatCompletion
-	var model string
 	var usage openai.CompletionUsage
+	var promptTokens, completionTokens int64
 	headers := []*configPb.HeaderValueOption{}
+	complete := false
+
+	defer func() {
+		// Wrapped in a function to delay the evaluation of parameters. Using complete to make sure DoneRequestTrace only call once for a request.
+		if !hasCompleted && complete {
```
varungup90, line 376: This scenario can happen in streaming case because total tokens stream chunk comes second or third last.

1. stream chunk with total tokens is processed -> set complete = true and DoneRequestTrace is processed.
2. stream chunk comes and processed successfully -> returns complete = false
3. stream chunk comes with EOF -> now hasComplete = false, error on processing stream chunk so set complete = true. DoneRequestTrace will be executed again.

pkg/plugins/gateway/gateway.go
```diff
@@ -223,50 +228,53 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
...
+			"error processing request body"), model, targetPodIP, stream, term
 	}
 
 	if model, ok = jsonMap["model"].(string); !ok || model == "" {
 		klog.ErrorS(nil, "model error in request", "requestID", requestID, "jsonMap", jsonMap)
 		return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError,
 			[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{
 				Key: "x-no-model", RawValue: []byte(model)}}},
-			fmt.Sprintf("no model in request body or model %s does not exist", model)), targetPodIP, stream
+			fmt.Sprintf("no model in request body or model %s does not exist", model)), model, targetPodIP, stream, term
```
varungup90: in the error line, you remove the part of "or model %s does not exist", that is covered in the next if condition check

pkg/plugins/gateway/gateway.go
```diff
@@ -223,50 +228,53 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
...
 	stream, ok = jsonMap["stream"].(bool)
 	if stream && ok {
 		streamOptions, ok := jsonMap["stream_options"].(map[string]interface{})
 		if !ok {
+			klog.ErrorS(nil, "no stream option available", "requestID", requestID, "jsonMap", jsonMap)
 			return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError,
 				[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{
 					Key: "x-stream-options", RawValue: []byte("stream options not set")}}},
-				"error processing request body"), targetPodIP, stream
+				"error processing request body"), model, targetPodIP, stream, term
```
varungup90: nit: replace error message with "no stream option available"

pkg/plugins/gateway/gateway.go
```diff
@@ -223,50 +228,53 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
...
 		}
 		includeUsage, ok := streamOptions["include_usage"].(bool)
 		if !includeUsage || !ok {
+			klog.ErrorS(nil, "no stream with usage option available", "requestID", requestID, "jsonMap", jsonMap)
 			return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError,
 				[]*configPb.HeaderValueOption{{Header: &configPb.HeaderValue{
 					Key: "x-stream-options-include-usage", RawValue: []byte("include usage for stream options not set")}}},
-				"error processing request body"), targetPodIP, stream
+				"error processing request body"), model, targetPodIP, stream, term
```
varungup90: nit: replace error message with "no stream with usage option available"

pkg/plugins/gateway/gateway.go
zhangjyr: Ok, I didn't know how stream really works. It's easy to fix the "once" problem, but I do need to postpone the "complete" to the EOF instead of the usage chunk.

zhangjyr: For now, I only fixed "once" problem by initializing the complete as "false || hasCompleted"

PR #569, opened 2025-01-14
Add request length for traces

happyandslow:
Allow generator to accept request/output lengths as input

### DIFF
--- a/benchmarks/generator/client.py
+++ b/benchmarks/generator/client.py
@@ -16,7 +16,7 @@ async def send_request(client, model, endpoint, prompt, output_file):
             model=model,
             messages=prompt,
             temperature=0,
-            max_tokens=128
+            max_tokens=2048
         )
 
         latency = asyncio.get_event_loop().time() - start_time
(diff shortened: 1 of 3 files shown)


benchmarks/generator/utils.py
```diff
@@ -55,8 +81,28 @@ def plot_workload(workload_dict, interval_ms, output_file: str = None):
...
         os.makedirs(os.path.dirname(output_file), exist_ok=True)
-        plt.savefig(output_file)
-        logging.info(f'Saved workload plot to {output_file}')
+        plt.savefig(f"{output_file}-traffic.pdf")
+        logging.info(f'Saved traffic plot to {output_file}-traffic.pdf')
+        
+        
+    fig, ax = plt.subplots()
+    for workload_name, workload in workload_dict.items():
+        input_lengths = [item["requests"][0]['prompt_length'] for item in workload]
```
gangmuk, line 90: Does it get right input_lengths? Can you quickly check?

happyandslow: Double checked. We use prompt_length and output_length in the workload file to represent input/output lengths

PR #590, opened 2025-01-22
Update routers to use GetPodModelMetric api and misc cleanup in metri…

varungup90:


development/app/app.py
```diff
@@ -25,8 +25,8 @@
 # Global storage for overridden values
 overrides = {}
 
-MODEL_NAME = os.getenv('MODEL_NAME', 'llama2-70b')
-DEPLOYMENT_NAME = os.getenv('DEPLOYMENT_NAME', 'llama2-70b')
+MODEL_NAME = os.getenv('MODEL_NAME', 'llama2-7b')
+DEPLOYMENT_NAME = os.getenv('DEPLOYMENT_NAME', 'llama2-7b')
```
Jeffwan, line 29: nice catch. name has been changed. We probably can inject from env using reference later.

development/app/app.py
```diff
@@ -566,16 +566,19 @@ def metrics():
...
-# TYPE vllm:lora_requests_info gauge
-vllm:lora_requests_info{max_lora="1",running_lora_adapters="text2sql-lora-1",waiting_lora_adapters=""} 1
-"""
+    metrics_output = ""
     for metric in simple_metrics:
         metrics_output += generate_counter_gauge_metric(metric["name"], metric["type"], metric["description"],
                                                         model_name, metric["value"])
         metrics_output += generate_counter_gauge_metric(metric["name"], metric["type"], metric["description"],
-                                                        "lora-1", metric["value"], help_header=False)
+                                                        "text2sql-lora-1", metric["value"], help_header=False)
```
Jeffwan: em. does user register a different name here?

Jeffwan: this is not a blocker for the merge.

development/app/app.py
varungup90: +1

development/app/app.py
varungup90: not sure from where lora-1 name came from. I added text2sql lora name based off model adapter.

PR #599, opened 2025-01-24
Patch release for older vllm engine lora support in gateway plugins

varungup90:


pkg/plugins/gateway/gateway.go
```diff
@@ -410,17 +410,18 @@ func (s *Server) HandleResponseBody(ctx context.Context, requestID string, req *
 					Key: "x-error-response-unmarshal", RawValue: []byte("true"),
 				}}},
 				err.Error()), complete
-		} else if res.Model != model {
```
zhangjyr, line 413: This check is used to handle error messages from the vLLM model. I assumed "model" from request header should equal to "res.Model" returned from the response. If  that is not the case, replacing with len(res.Model) == 0 is ok.

PR #616, opened 2025-01-29
Add framework to setup integration test

varungup90:


development/app/app.py
```diff
@@ -278,6 +278,7 @@ def completion():
         return jsonify(err), 500
 
 
+@app.route('/chat/completions', methods=['POST'])
```
Jeffwan: what's the endpoint for?

varungup90: on using openai-go official openai go library uses /chat/completion and v1/chat/completion is used by batch api.

Jeffwan: I feel this is kind of confusing. golang sdk hit different endpoints?  
1. we are not using batch in testing scenarios
2. `/chat/completion` this is openai not compatible


is there a detail problem on the golang sdk usage?

varungup90: Agree, this is api doc I found: https://github.com/openai/openai-go/blob/5b469242915f7e718c8bf94438080fec06fd6aa9/api.md?plain=1#L71

https://github.com/openai/openai-go/blob/5b469242915f7e718c8bf94438080fec06fd6aa9/chatcompletion.go#L119

Jeffwan: seems they assume `v1` as part of the `UrlBase`.  final url is still v1/chat/completions? https://github.com/openai/openai-go/issues/105

If it does need `/chat/completions`, let document this part in the file.

varungup90: added a middleware router to append /v1.

PR #628, opened 2025-02-04
Fix out of space error in running integration test github workflow

varungup90:


.github/workflows/installation-tests.yml
```diff
@@ -28,6 +28,28 @@ on:
...
+          docker build -t aibrix/vllm-mock:nightly -f Dockerfile .
+      
+      - name: Save image
+        run: |
+          docker save aibrix/vllm-mock:nightly > vllm-mock.tar
+      
+      - name: Upload image artifact
+        uses: actions/upload-artifact@v4
+        with:
+          name: vllm-mock-image
```
Jeffwan: this way would be tricky. it consumes lots of artifact space and we need to pay for the artifacts. do you know the size of the artifacts? If we know the size, we may consider to use large instance as well (also need more cost)

varungup90: Got it. Reason to add a separate job was to build mock image in parallel with other images to reduce total time to run integ test. 

Since the control plane images are built with git sha, should build all the images in parallel within same job and remove the upload/download artifact altogether?

varungup90: Reverted mock image build and deploy as before.

.github/workflows/installation-tests.yml
```diff
@@ -86,17 +105,17 @@ jobs:
...
-          
-            kind load docker-image aibrix/${image}:${{ github.sha }} --name installation-test
-            kind load docker-image aibrix/${image}:nightly --name installation-test
+              # Retag the image
+              # This is for application integration, since it is not that easy to override all commits in manifest
+              # It is better to use nightly to represent the latest image
+              docker tag aibrix/${image}:${{ github.sha }} aibrix/${image}:nightly
           done
 
+          kind load docker-image aibrix/controller-manager:${{ github.sha }} aibrix/gateway-plugins:${{ github.sha }} aibrix/metadata-service:${{ github.sha }} aibrix/runtime:${{ github.sha }} --name installation-test
```
Jeffwan, line 116: I didn't get this part, seems the original way works better? now we need to manually specific the container images again?

.github/workflows/installation-tests.yml
```diff
@@ -54,16 +54,35 @@ jobs:
           name: ${{ matrix.image }}-image
           path: ${{ matrix.image }}.tar
           retention-days: 1
-
+
```
Jeffwan: remove blanks

.github/workflows/installation-tests.yml
varungup90: To load images in parallel.

Makefile
```diff
@@ -137,7 +137,8 @@ define push_image
 endef
 
 .PHONY: docker-build-all
-docker-build-all: docker-build-controller-manager docker-build-gateway-plugins docker-build-runtime docker-build-metadata-service ## Build all docker images
+docker-build-all: 
+	make -j 4 docker-build-controller-manager docker-build-gateway-plugins docker-build-runtime docker-build-metadata-service ## Build all docker images
```
Jeffwan: @nwangfw I remember you did this before. Could you find the related issue? What the performance looks like?

nwangfw: @Jeffwan It is https://github.com/aibrix/aibrix/pull/398.

In the local env, we can reduce 40% docker building time. In Github CI. the overall test time reduced 1/3 (from ~9m to ~6m).

Jeffwan: that's the `the matrix approach`? what about `make -j 4` approach? If there's gains, why did not you use it earlier? I just forgot the context, could be something else I can not remember clearly

nwangfw: > that's the `the matrix approach`? what about `make -j 4` approach? If there's gains, why did not you use it earlier? I just forgot the context, could be something else I can not remember clearly

You' are correct. That is `the matrix approach`. For `make -j 4`, we only reduced the building time in our local docker-desktop env. There is no gain in our Github CI. https://github.com/aibrix/aibrix/pull/398#issuecomment-2480128074

Jeffwan: seems by default it use 2 core machine @varungup90 could you change to `make -j $(nproc)` instead?

.github/workflows/installation-tests.yml
Jeffwan: @nwangfw did some try earlier but we did not notice big difference because our image is very small. I am ok with this change

nwangfw: No. I didn't test parallel loading method previously. The parallel loading idea is great and I am OK with this modification too. Loading should be quick, right? Wondering how much gain that we can achieve via this modification.

varungup90: Time reduced from 6ish to 2ish mins.

Makefile
varungup90: Will update it

.github/workflows/installation-tests.yml
nwangfw: > Time reduced from 6ish to 2ish mins.

Wow, that seems a big improvement!

Makefile
varungup90: Done.

PR #641, opened 2025-02-07
Add prefix cache aware routing

varungup90:


pkg/cache/cache.go
```diff
@@ -68,13 +69,22 @@ type Cache struct {
 	requestTrace      *sync.Map                                            // model_name: RequestTrace
 	numRequestsTraces int32                                                // counter for requestTrace
 	pendingRequests   *sync.Map                                            // model_name: *int32
+
+	prefixBlocks map[uint64]Block //prefix_hash:Block
```
Jeffwan, line 74: should we have a separate components for indexing related codes? this is not related to pods events, indexing is mainly triggered by requests

pkg/cache/cache.go
```diff
@@ -68,13 +69,22 @@ type Cache struct {
...
+}
+
+type Block struct {
+	modelToPods    map[string]map[string]time.Time // model_name: map[pod_name]pod_last_access_time
```
Jeffwan, line 78: the reference is not there yet? It will be in a separate follow up PR?

pkg/cache/cache.go
```diff
@@ -68,13 +69,22 @@ type Cache struct {
...
-	modelIdentifier                       = "model.aibrix.ai/name"
-	podPort                               = 8000
-	defaultPodMetricRefreshIntervalInMS   = 50
-	expireWriteRequestTraceIntervalInMins = 10
+	modelIdentifier                        = "model.aibrix.ai/name"
+	podPort                                = 8000
+	defaultPodMetricRefreshIntervalInMS    = 50
+	expireWriteRequestTraceIntervalInMins  = 10
+	defaultPrefixCacheBlockSize            = 16
+	defaultPrefixCacheEvictionInternalInMS = 50
```
Jeffwan: Could we do in-place update rather than a separate loop? We can have some discussion later. not a blocker at this moment

pkg/cache/cache.go
```diff
@@ -963,3 +1009,90 @@ func (c *Cache) aggregateMetrics() {
 		}
 	}
 }
+
+// returns matchedTokens, unMatchedTokens, matchedPods
+func (c *Cache) MatchPrefix(tokens []int, model string, pods []*v1.Pod) ([]int, []int, []*v1.Pod) {
```
Jeffwan, line 1034: Let's build a cache interface for it. We can implement hashmap or radix version without too many changes here.

pkg/plugins/gateway/algorithms/least_kv_cache.go
```diff
@@ -43,7 +43,7 @@ func NewLeastKvCacheRouter() Router {
 	}
 }
 
-func (r leastKvCacheRouter) Route(ctx context.Context, pods map[string]*v1.Pod, model string) (string, error) {
+func (r leastKvCacheRouter) Route(ctx context.Context, pods map[string]*v1.Pod, model, message string) (string, error) {
```
Jeffwan, line 46: seems the contexts are expanding, we can build a routingContext later to put all parameters inside

pkg/utils/util.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"os"
+
+	"github.com/pkoukk/tiktoken-go"
+	tiktoken_loader "github.com/pkoukk/tiktoken-go-loader"
+	"k8s.io/klog/v2"
+)
+
+const encoding = "cl100k_base"
```
Jeffwan, line 28: let's leave a comment on encoding to model mapping. 
https://cookbook.openai.com/examples/how_to_count_tokens_with_tiktoken

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -0,0 +1,113 @@
...
+			klog.Infof("using AIBRIX_PREFIX_CACHE_MATCH_THRESHOLD_PERCENT env value for prefix cache match threshold percent: %d", intValue)
+			return intValue
+		}
+	}
+	klog.Infof("using default prefix cache match threshold percent: %d", defaultPrefixCacheMatchThresholdPercent)
+	return defaultPrefixCacheMatchThresholdPercent
+}
+
+type prefixCacheRouter struct {
+	cache *cache.Cache
```
Jeffwan, line 55: we talked about earlier. We leverage more information from hooked requests, I feel we can have a different interface here.

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -0,0 +1,113 @@
...
+	if err != nil {
+		return "", err
+	}
+
+	var targetPod *v1.Pod
+	matchedTokens, unMatchedTokens, matchedPods := p.cache.MatchPrefix(tokens, model, readyPods)
+	if len(matchedTokens)*100/len(tokens) > prefixCacheMatchThresholdPercent {
+		targetPod = matchedPods[rand.Intn(len(matchedPods))]
+	} else {
+		targetPod = readyPods[rand.Intn(len(readyPods))]
```
Jeffwan, line 91: It looks good now. We can add a TODO later. We can work on load based algorithms as a backup algorithm.

pkg/cache/cache.go
```diff
@@ -963,3 +1009,90 @@ func (c *Cache) aggregateMetrics() {
...
+	c.mu.Lock()
+	defer c.mu.Unlock()
+
+	for i := 0; i < len(unMatchedTokens); i += prefixCacheBlockSize {
+		end := i + prefixCacheBlockSize
+		if end > len(unMatchedTokens) {
+			end = len(unMatchedTokens)
+		}
+		chunk := unMatchedTokens[i:end]
+		prefixHash := xxhash.Sum64(IntArrayToByteArray(chunk))
```
Jeffwan, line 1086: here. it just consider the current block size? just like to confirm this is not 100% same as vLLM's solution right? their 1st block hash is part of 2nd hash input

varungup90: Yes, hash only for current block. No link list kind of behavior that vllm has. For our usecase we do not need that link list behavior.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: Done.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: Did not get the comment, can you describe more

pkg/utils/util.go
varungup90: Done.

pkg/plugins/gateway/algorithms/least_kv_cache.go
varungup90: Done, added a todo.

pkg/cache/cache.go
varungup90: added a todo for now.

pkg/cache/cache.go
varungup90: added a todo for now.

pkg/cache/cache.go
varungup90: in match prefix, it checks whether the block referenced is for input model and checks which pod match.

pkg/cache/cache.go
varungup90: Agree we can split core cache (pod/model events), request trace and prefix cache data into three separate modules. I will add a todo for now.

pkg/utils/util.go
```diff
@@ -0,0 +1,51 @@
...
+
+	"github.com/pkoukk/tiktoken-go"
+	tiktoken_loader "github.com/pkoukk/tiktoken-go-loader"
+	"k8s.io/klog/v2"
+)
+
+// https://cookbook.openai.com/examples/how_to_count_tokens_with_tiktoken
+const encoding = "cl100k_base"
+
+func TokenizeInputText(text string) ([]int, error) {
```
gaocegege, line 30: Can we just use strings instead of token IDs? Those tokenization costs can really add up, especially since it happens with every request.

pkg/cache/cache.go
```diff
@@ -68,13 +70,23 @@ type Cache struct {
 	requestTrace      *sync.Map                                            // model_name: RequestTrace
 	numRequestsTraces int32                                                // counter for requestTrace
 	pendingRequests   *sync.Map                                            // model_name: *int32
+
+	prefixBlocks map[uint64]Block //prefix_hash:Block
+}
+
+type Block struct {
+	modelToPods    map[string]map[string]time.Time // model_name: map[pod_name]pod_last_access_time
```
gaocegege, line 78: What happens if the gateway server crashes and restarts, or if multiple server instances are running, with this design?

pkg/plugins/gateway/algorithms/prefix_cache.go
Jeffwan: My point is we do not need `cache`, can we have something like `prefixCacheIndexer` a new structure? 
- `prefixCacheIndexer` is for requests. it gets updated only if there's request coming in/out.
- `cache` is for pod events. it gets updated if there's pod event change. 

I suggest to have two different struct for different purpose.

pkg/cache/cache.go
Jeffwan: if that case, we need to do up to O(n) calculations?  n=number of blocks

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: Yes, for that I have added a todo to split cache (pod events), request trace and prefix cache. There are few more small todos related to refactoring, will address them in follow up PR.

pkg/cache/cache.go
varungup90: for the gateway server restart scenario: prefix cache is stored in-memory and will be lost. 

For multiple server scenario: Right now single gateway server instance is supported, when we decide to move to multi server scenario requests can be partitioned based off models or any other key.

pkg/cache/cache.go
varungup90: Yes. If you are referring to that link list behavior could prevent O(n) calculations then it wont be the case. Total computations stays the same.

pkg/utils/util.go
varungup90: yes, we thought about it and decided to tokenize the input. I forgot the reason. cc @jeffwan

pkg/cache/cache.go
gaocegege: Thanks for the explanation.

pkg/plugins/gateway/algorithms/prefix_cache.go
Jeffwan: Sounds good

pkg/cache/cache.go
Jeffwan: it's a little bit different. LinkedList you can do binary search etc for optimization. In this way, we can only do O(n).

pkg/utils/util.go
Jeffwan: introducing tokenizer improve the accuracy but it does add the overhead. We can have different implementations here I feel.

pkg/cache/cache.go
varungup90: Let me look into this, but for our usecase we need to evaluate all blocks to ensure a 50%+ hit rate.

PR #657, opened 2025-02-12
Add interface for prefix cache indexer

varungup90:


pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"time"
+
+	v1 "k8s.io/api/core/v1"
+)
+
+type PrefixCacheIndexer interface {
+	// MatchPrefix matches the longest prefix sequence for input request (passed as input tokens)
+	// and returns matched prefix (as tokens), remaining unmatched input request (as tokens) and pods matching the prefix
+	MatchPrefix(inputTokens []int, model string, pods []*v1.Pod) (matchedTokes []int, unMatchedTokens []int, matchedPods []*v1.Pod)
```
Jeffwan: @DwyaneShi pleas help review the routing indexer interface.

Jeffwan: @gangmuk You'd like to review this as well since you will join the indexer implementation.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// AddPrefix adds the unmatched tokens returned from MatchPrefix which will be used future requests
+	AddPrefix(unMatchedTokens []int, model, pod string)
```
DwyaneShi: we need the matched tokens to be in the params as well such that if the impl wants to use radix tree-like structures , it can rebuild the whole path starting from the root to the unmatched portion.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
```diff
@@ -0,0 +1,210 @@
...
+	var lastTokenMatchIndex int
+
+	for i := 0; i < len(tokens); i += prefixCacheBlockSize {
+		end := i + prefixCacheBlockSize
+		if end > len(tokens) {
+			end = len(tokens)
+		}
+
+		chunk := tokens[i:end]
+		prefixHash := xxhash.Sum64(IntArrayToByteArray(chunk))
```
DwyaneShi: another approach is constructing a logical radix tree structure by considering the hash value of its previous chunk, i don't know which one would perform better, do we plan to have some perf evaluations on these potential approaches? my gut feeling is like this PR's approach would not lead to a stable performance after a long running if we have pretty number of chunks containing the same set of tokens but in different positions.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
```diff
@@ -0,0 +1,35 @@
...
+
+	// Evict is invoked at fixed internal to clean up expired tokens using LRU policy from prefix cache.
+	Evict(now time.Time)
```
DwyaneShi: is the eviction policy fixed to LRU? by the way, the following impl does not maintain an LRU list, instead the eviction is entirely TTL-based, more like a GC.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
gangmuk: typo I think `matchedTokes` -> `matchedTokens`

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
```diff
@@ -0,0 +1,210 @@
...
+			instance.Evict(time.Now())
+		}
+	}()
+
+	return instance
+}
+
+// returns matchedTokens, unMatchedTokens, matchedPods
+// TODO: add an interface with multiple implementations such as hash or radix tree
+func (c *PrefixHashTable) MatchPrefix(tokens []int, model string, pods []*v1.Pod) ([]int, []int, []*v1.Pod) {
```
gangmuk: do we use hash for now? not the tree?

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
```diff
@@ -0,0 +1,35 @@
+/*
```
Jeffwan: a quick questions on the package name. Could we use something simple as the package name like `indexing` or two layers etc. or `prefixcacheindexer` & shorter file names

```
pkg/plugins/gateway/indexing/prefixcache/
    radix.go        # Radix tree-based prefix index
    hash.go         # Hash-based prefix index
    indexer.go      # Common indexing logic
```

```
pkg/plugins/gateway/prefixcacheindexer/
    radix.go        # Radix tree-based prefix index
    hash.go         # Hash-based prefix index
    indexer.go      # Common indexing logic
```

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
Jeffwan: what's the usage of `unMatchedTokens`? is it better to return directly or just matched and caller could know the `unMatched` automatically

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
Jeffwan: do we need last access time etc here?

pkg/plugins/gateway/prefixcacheindexer/prefix_hash_test.go
```diff
@@ -0,0 +1,132 @@
...
+	"testing"
+	"time"
+
+	"github.com/aibrix/aibrix/pkg/utils"
+	"github.com/stretchr/testify/assert"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func Test_PrefixHashTableE2E(t *testing.T) {
```
Jeffwan: > `E2E` 

this is still unit tests?

Jeffwan: does E2E mean the full or empty case? name is a little bit confused

pkg/plugins/gateway/prefixcacheindexer/prefix_hash_test.go
```diff
@@ -0,0 +1,132 @@
...
+			cache: PrefixHashTable{
+				blocks: map[uint64]Block{},
+			},
+			model: "m1",
+			pods: []*v1.Pod{
+				{ObjectMeta: metav1.ObjectMeta{Name: "p1"}},
+				{ObjectMeta: metav1.ObjectMeta{Name: "p2"}},
+			},
+			matchTokens:   []int{},
+			unMatchTokens: []int{9906, 4435, 0, 3639, 264, 7839, 6187, 0, 220, 57668, 53901, 3574, 244, 98220, 6447, 43240, 82696, 58666, 53901, 9554, 15120, 36827, 28308, 232, 6447},
```
Jeffwan: it would be great to create a var, so we can use slice to manage the positions and that's more visible checking the tokens

pkg/plugins/gateway/prefixcacheindexer/prefix_hash_test.go
```diff
@@ -0,0 +1,132 @@
...
+						lastAccessTime: time.Now(),
+					},
+				},
+			},
+			model: "m1",
+			pods: []*v1.Pod{
+				{ObjectMeta: metav1.ObjectMeta{Name: "p1"}},
+				{ObjectMeta: metav1.ObjectMeta{Name: "p2"}},
+			},
+			matchTokens:   []int{9906, 4435, 0, 3639, 264, 7839, 6187, 0, 220, 57668, 53901, 3574, 244, 98220, 6447, 43240},
```
Jeffwan: this is hard to expand, I think let's just use slice later

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: For hash table usecase, we only need to add unmatched tokens hence returned unmatchedtokens.

Technically caller can know unmatched tokens from returned matched tokens but it will add extra computation to figure out.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: @DwyaneShi Does renaming unMatchedTokens to tokens help? For hash table usecase only unMatchedTokens can be passed and for radix tree, original input of tokens can be passed?

@jeffwan When the tokens are added, it will take current time as last access time. Wont be much difference.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
varungup90: Logical radix tree structure will be same as vllm block manager approach. In vllm, logical link list is created where blocks are connected using hash value of previous and current tokens. 

A link list structure or logical radix tree structure can be created but will still need to match all token blocks same performance will be same O(n).

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: Partially true. TTL value for each block is extended on each access, so first unused blocks will expire.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
varungup90: Yes

pkg/plugins/gateway/prefixcacheindexer/prefix_hash_test.go
varungup90: It is unit test, E2E here is more testing entire sequence within unit test.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
gangmuk: access time would be useful even for actual lru-based eviction (currently ttl-based eviction like @DwyaneShi pointed). access time can be used when there are multiple nodes/blocks having the same reference count for lru?

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: No strong opinion on naming. For now, we only have one usecase so we can use current name as we add more such usecases we can rename packages and change dir loc etc. WDYT?

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
gangmuk: just curious. you planned to do it with tree originally and can I ask why you changed the plan? 

when the length of a token is not fixed then match/insert could be tricky in hashtable?

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+type PrefixHashTable struct {
+	mu     sync.RWMutex
+	blocks map[uint64]Block
+}
+
+type Block struct {
+	modelToPods    map[string]map[string]time.Time // model_name: map[pod_name]pod_last_access_time
+	lastAccessTime time.Time                       //block_last_access_time
```
gangmuk: maybe we need reference count for each block? 

and any metric (last access or ref count) that will be used for eviction, I think it should be per-GPU(pod) not per-block? is access time of token in each pod tracked currently?

If the token length of each block/node could vary in the future (I see it is fixed currently), token_length field will be needed too.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
DwyaneShi: Yes, indexing performance is the same, but i was saying the routing accuracy. Suppose we have 3 pods and each chunk has four tokens, let's consider these two requests: 1. \[1 2 3 4\] \[2 3 4 5\] \[3 4 5 6\], 2. \[11 12 13 14\] \[3 4 5 6\]. Request 1 comes first and it is randomly routed to Pod 1, so the cache is \[1 2 3 4\](P1) \[2 3 4 5\](P1) \[3 4 5 6\](P1). Now Request 2 comes, since \[11 12 13 14\] not matched in the cache, let's assume it is randomly routed to Pod 2, then the cache is \[1 2 3 4\](P1) \[2 3 4 5\](P1) \[3 4 5 6\](P1, P2) \[11 12 13 14\](P2). However, if Request 1 comes again, even though it totally hits the cache, but since we have \[3 4 5 6\](P1, P2) in the cache, we return P1, P2 as the matched pod list, the request has chances to be routed to P2 which does not have any its kv cache at all. But if we use the vllm approach, the accuracy will be enhanced a lot.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
DwyaneShi: @varungup90 but for vllm-like hashing approach, if we only know the entire input tokens, we need to lookup again the find which part is unmatched and then insert the unmatched portions into the hash table, which is not as efficient as knowing the matched parts, in which case we can just assume the matched parts are still in the cache and only do insertions.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
DwyaneShi: Got you. I think we don't need to mention LRU in the comment, it's all up to the impl. Param `now` in the API is used to determine block expiracy and does not restrict the eviction policy to be LRU.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
```diff
@@ -0,0 +1,210 @@
...
+		if !ok {
+			blockPods = map[string]time.Time{}
+			block.modelToPods[model] = blockPods
+		}
+
+		blockPods[pod] = time.Now()
+	}
+}
+
+func (c *PrefixHashTable) Evict(now time.Time) {
```
DwyaneShi: the stop-the-world eviction/GC would hurt the performance if we have tons of items in the cache. let's leave a TODO here for further perf study and potential enhancements.

varungup90: Will add a TODO, but thats a trade-off having high cache hit rate vs GC pressure. Important thing is the rate at which cache items are deleted only those will be GC'ed.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
varungup90: To keep it simple, #tokens per block are fixed.

For last access time is used for TTL expiration, no need to track ref count.

varungup90: Implementation for hash based solution is easier and we want to include this feature in 0.2 release.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: Updated the comment to remove LRU and added a todo to also add LRU policy once number of blocks to track in cache is implemented.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
varungup90: No, at each iteration it filters the pods. For request 1, it will filter pods and going forward only P1 will be matched. 

If you check the interface method MatchPrefix, it returns matchedPods, so in above scenario P1 (matched pod) will be returned for first block iteration and only that pod will be used for second block iteration.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: Interface method will have argument name tokens but it is upto the caller what token is passed. For hash implementation, only unmatched tokens is passed to AddPrefix where as for radixtree all tokens will be passed.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: I just saw the edited comment with recommendation. Second option is better, I will update it.

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: @gangmuk For matched block, last_access_time is updated if same block is read.

for AddPrefix, only new blocks are added and for them current time is used.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
DwyaneShi: i didn't find the filter logic, could you please give a pointer?

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
DwyaneShi: I still prefer adding one more param to differentiate matched & unmatched, but I understand your perspective. I'm okay with moving forward to merge it.

gangmuk: @DwyaneShi @varungup90 I think as @DwyaneShi mentioned earlier, we need matched tokens for radix tree. even if the current simple hash based algorithm is used, for other future logic we need it. because this interface will be shared.

pkg/cache/cache.go
```diff
@@ -53,7 +50,6 @@ import (
 var once sync.Once
 
 // type global
-// TODO: split pod/model events, request trace and prefix cache into separate modules.
 type Cache struct {
```
gangmuk, line 53: the name can be more specific like hash_based_cache or something. file name as well.  I am planning to implement preble (tree based one). then there will be another cache implementation.

WDYT

Jeffwan: this is a different cache and you may not need to touch it in your work. you can check `prefixcacheindexer` package

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
varungup90: at the end of MatchPrefix function

image

pkg/plugins/gateway/prefixcacheindexer/prefix_cache_indexer.go
varungup90: Current interface will work from my understanding of Preble implementation. Lets merge the change and interface can be updated once radix tree implementation is kicked off.

pkg/plugins/gateway/prefixcacheindexer/prefix_hash.go
DwyaneShi: how can this logic filer out P2 if `lastMatchedBlock` is \[3 4 5 6\] (which has Pods P1 and P2) for Request 1 while `pods` is \[P1, P2, P3\]?

Issue #666, opened 2025-02-13
Improve the autoscaling benchmark scripts


Jeffwan:
### 🚀 Feature Description and Motivation

Copy some TODO items from issue https://github.com/aibrix/aibrix/pull/650

WIP items:
- [x] Improving aibrix/benchmarks/generator/client.py with async and streaming mode
  - [x] TTFT
  - [x] TPOT
  - [ ] Goodput
- [x] Adjust the QPS scale
- [x] Improving plot script
  - [x] Tokens/s times series plot
  - [x] TTFT times series plot
  - [x] TPOT times series plot
  - [ ] Goodput times series plot

### Use Case

This is to improve the qualify of the autoscaling experiments

Jeffwan:
@happyandslow what's the status of this story. 

1. All the metrics are collected? 
2. Did you make changes to the plot diagram? 
3. is #667 resolved?

Jeffwan:
@happyandslow Any updates?

happyandslow:
> [@happyandslow](https://github.com/happyandslow) Any updates?


The modification is applied to #676
Improving aibrix/benchmarks/generator/client.py with async and streaming mode

- [x] TTFT
- [x] TPOT
- [x] Goodput
- [x] Adjust the QPS scale

Improving plot script

- [x] Tokens/s times series plot
- [x] TTFT times series plot
- [x] TPOT times series plot
- [ ] Goodput times series plot

I'm leaving out Goodput since it is a one time measurement throughout the entire experiment. 

I'm waiting for some free cycles to test these out before PR can be merged.

happyandslow:
#667 should solved (check comments)
The changes I was going to try were #862

Jeffwan:
Great. We can close this issue, we can revisit #862 once it's ready.

PR #676, opened 2025-02-14
Adding streaming client for AIbrix experiments

happyandslow:
Addressing #666 
- Improving aibrix/benchmarks/generator/client.py with async and streaming mode
- Supporting the following metrics in the analysis
- - Tokens/s times series plot
- - TTFT times series plot
- - TPOT times series plot
- - Goodput times series plot

Goodput spec (https://github.com/vllm-project/vllm/blob/main/benchmarks/benchmark_serving.py)


benchmarks/generator/client.py
```diff
@@ -4,12 +4,144 @@
...
+        request_latency = request_end_time - request_start_time
+        
+        logging.warning(
+            f"Request completed in {request_latency:.2f} seconds with throughput 0.0 tokens/s, request {formatted_prompt} response {response}")
+        return response
+        
+
+    
+    
+    # async with session.post(api_url, headers=headers, json=pload) as response:
```
Jeffwan: do we still need it?

benchmarks/generator/client.py
```diff
@@ -4,12 +4,144 @@
...
+        request_start_time = time.perf_counter()
+        try: 
+            async for chunk, _ in response.content.iter_chunks():
+                chunks = [chunk]
+                now_time = time.perf_counter()
+                if first:
+                    time_to_first = now_time - previous_token_time
+                    first = False
+                else:
+                    token_latencies.append(now_time - previous_token_time)
```
gangmuk: Are these performance numbers logged somewhere? I don't see them used or returned.

benchmarks/generator/client.py
```diff
@@ -4,12 +4,144 @@
...
+                previous_token_time = now_time
+            output = b"".join(chunks).decode("utf-8")
+            output = output.rstrip(
+                "\n\t "
+            )  # Remove trailing whitespace characters including EOF, and "[DONE]"
+        except Exception as e:
+            logging.error(f"Failed to read response for request: {e}")
+        
+        try:
+            response = json.loads(output)
```
gangmuk: Where is the client side metrics logged? We were logging it before and I also added a few more metrics (e.g., failed number of requests)

```python
 try:
            data = response.json()
            end_time = asyncio.get_event_loop().time()
            latency = end_time - start_time
            result = {
                "status_code": response.status_code,
                "start_time": start_time,
                "end_time": end_time,
                "latency": latency,
                "throughput": data['usage']['completion_tokens'] / latency,
                "prompt_tokens": data['usage']['prompt_tokens'],
                "output_tokens": data['usage']['completion_tokens'],
                "total_tokens": data['usage']['total_tokens'],
                "input": prompt,
                "output": data['choices'][0]['text'],
            }
            if completion_map[request_id] != 0:
                logging.error(f"Request {request_id} already completed")
                assert False
            completion_map[request_id] = 1
            logging.warning(f"Batch {batch_id}, Request {request_id}, completed in {latency:.2f} seconds with throughput {result['throughput']:.2f} tokens/s")
            logging.warning(f"Total sent requests so far: {len(completion_map)}, completed requests: {sum(completion_map.values())}, completion_ratio: {sum(completion_map.values()) / len(completion_map)*100:.2f}%")
        except Exception as e:
            logging.error(f"Status: {response.status_code}, Raw response: {response.text}")
            logging.error(f"Error parsing response from {args.endpoint}: {str(e)}")
            result = {
                "status_code": response.status_code,
                "start_time": start_time,
                "end_time": None,
                "latency": None,
                "throughput": None,
                "prompt_tokens": None,
                "output_tokens": None,
                "total_tokens": None,
                "input": prompt,
                "output": None,
            }
        
        ## Write result to JSONL file
        output_file.write(json.dumps(result) + "\n")
```

benchmarks/generator/client.py
```diff
@@ -4,12 +4,144 @@
...
+                          model: str, 
+                          api_key: str, 
+                          load_struct: List, 
+                          output_file: io.TextIOWrapper):
+    
+    headers = {
+        "User-Agent": "Benchmark Client",
+    }
+    if api_key is not None or api_key != "":
+        headers["Authorization"] = f"Bearer {api_key}"
```
gangmuk: I think `routing-strategy` header is missing. and also `Content-Type` (e.g., `"Content-Type": "application/json"`)

Jeffwan: for autoscaling case, we do not need to use `routing-strategy`. For routing case, this is mandatory field

benchmarks/generator/client.py
happyandslow: Added

benchmarks/generator/client.py
happyandslow: Added

benchmarks/client/client.py
```diff
@@ -0,0 +1,226 @@
...
+                send_request_streaming(client = client, 
+                                       model = model, 
+                                       endpoint = endpoint, 
+                                       prompt = formatted_prompt, 
+                                       output_file = output_file)
+            )
+            batch_tasks.append(task)
+        num_requests += len(requests)
+    await asyncio.gather(*batch_tasks)
+    logging.warning(f"All {num_requests} requests completed for deployment.")
```
gangmuk, line 106: The batch task would confuse users and batch is not necessary. each linen of the workload jsonl should be one request not a list of requests having the same timestamp. We can update this part in different PR though.

benchmarks/client/client.py
```diff
@@ -0,0 +1,226 @@
...
+            start_time = time.time()
+            asyncio.run(benchmark_batch(
+                client = client,
+                endpoint=args.endpoint, 
+                model=args.model, 
+                load_struct=load_struct, 
+                output_file=output_file, 
+            ))
+            end_time = time.time()
+            logging.info(f"Benchmark completed in {end_time - start_time:.2f} seconds")
```
gangmuk: naming 'batch' can be confusing. it is not related to my above comment. This batch means non-streaming I think.

benchmarks/client/client.py
happyandslow: We might want to test out batched requests in certain scenarios. For single request we will stick with workload traces with fine-grained timestamp. I think the current implementation shouldn't prevent requests (even with the same timestamp) to be dispatched in two loops and we will leave it to the workload generator to figure out what the workload should be. The client just stays simple for now.

benchmarks/client/client.py
happyandslow: Fixed

benchmarks/generator/client.py
happyandslow: Let's resolve the failing scenario in another PR

PR #701, opened 2025-02-18
Add model adapter e2e tests

varungup90:


test/e2e/util.go
```diff
@@ -101,3 +110,38 @@ func createOpenAIClientWithRoutingStrategy(baseURL, apiKey, routingStrategy stri
...
+	})
+	if err != nil {
+		t.Fatalf("chat completions failed : %v", err)
+	}
+	assert.Equal(t, modelName, chatCompletion.Model)
+	assert.NotEmpty(t, chatCompletion.Choices, "chat completion has no choices returned")
+	assert.NotNil(t, chatCompletion.Choices[0].Message.Content, "chat completion has no message returned")
+}
+
+func isModelDeploymentAvailable(t *testing.T, modelName string) bool {
```
Jeffwan: I can not find reference here, is this being used?

varungup90: It is not used, I have removed that function.

PR #702, opened 2025-02-18
Add cpu/memory resources for control plane components

varungup90:
https://github.com/aibrix/aibrix/issues/693

### DIFF
--- a/config/gateway/gateway-plugin/gateway-plugin.yaml
+++ b/config/gateway/gateway-plugin/gateway-plugin.yaml
@@ -38,11 +38,11 @@ spec:
             - containerPort: 50052
           resources:
             limits:
-              cpu: 500m
-              memory: 256Mi
+              cpu: 1
+              memory: 1Gi
             requests:
-              cpu: 10m
-              memory: 64Mi
+              cpu: 1
+              memory: 1Gi
           env:
             - name: REDIS_HOST
               value: aibrix-redis-master


config/gateway/gateway-plugin/gateway-plugin.yaml
```diff
@@ -36,6 +36,13 @@ spec:
           imagePullPolicy: IfNotPresent
           ports:
             - containerPort: 50052
+          resources:
+            requests:
+              cpu: 4
```
Jeffwan: can we use a little bit lower value here? just want to make sure it can be scheduled on any env without pressure. Does `2` makes sense here? memory 2Gi is enough or not?

varungup90: Done

config/gateway/gateway-plugin/gateway-plugin.yaml
```diff
@@ -36,6 +36,13 @@ spec:
           imagePullPolicy: IfNotPresent
           ports:
             - containerPort: 50052
+          resources:
+            limits:
+              cpu: 500m
```
Jeffwan: @varungup90 Could we change to  at least 1 or 2 CPU, this is the data plane component and should be very critical

Jeffwan: 500m and 256Mi I feel is not enough

PR #703, opened 2025-02-18
Process response headers in gateway

varungup90:
https://github.com/aibrix/aibrix/issues/671


pkg/plugins/gateway/gateway.go
```diff
@@ -421,6 +423,39 @@ func (s *Server) HandleResponseHeaders(ctx context.Context, requestID string, re
...
+			},
+		})
+	}
+
+	if isProcessingError {
+		return &extProcPb.ProcessingResponse{
+			Response: &extProcPb.ProcessingResponse_ImmediateResponse{
+				ImmediateResponse: &extProcPb.ImmediateResponse{
+					Status: &envoyTypePb.HttpStatus{
+						Code: envoyTypePb.StatusCode(processingErrorCode),
```
Jeffwan: in this case, we do not have any meaningful information right?

Jeffwan: BTW, the e2e test is still 500.

varungup90: I included all the headers returned from server. It also includes authorization error message as one of the header.

varungup90: updated the e2e test.

Jeffwan: does it still comes with the error message?

Jeffwan: header interception will lose some key error message which makes the debugging hard. I personally think we could postpone this change to v0.2.1 and make sure we have error messages etc

varungup90: Updated the logic, now error code is collected from HandleResponseHeaders, then get the error message from HandleResponseBody.

test/e2e/model_adapter_test.go
```diff
@@ -1,3 +1,19 @@
+/*
```
Jeffwan, line 1: @gangmuk reports missing license header failed the check. However, license check job was not triggered. not sure what's the problem. https://github.com/vllm-project/aibrix/issues/723

good to see it's added now

pkg/plugins/gateway/gateway.go
```diff
@@ -421,6 +428,24 @@ func (s *Server) HandleResponseHeaders(ctx context.Context, requestID string, re
 		})
 	}
 
+	var isProcessingError bool
+	var processingErrorCode int
+	for _, headerValue := range b.ResponseHeaders.Headers.Headers {
+		if headerValue.Key == ":status" {
+			code, _ := strconv.Atoi(string(headerValue.RawValue))
+			if code != 200 {
```
Jeffwan, line 436: 30x redirection are successful codes. Seems we will never have such case? Is it possible to have 20x?  here? better to have a helper method and let's consider those cases

Jeffwan: if we have the helper, we can further simplify the return results, just return the code is enough.

pkg/plugins/gateway/gateway.go
Jeffwan: sounds good

pkg/plugins/gateway/gateway.go
varungup90: I checked the code, codes returned are 200, 400, 401, 404 and 500. HandleResponseBody should only be called if response has ChatCompletitions object without that it will throw error.

Jeffwan (approved):

Issue #417, opened 2024-11-20
Run small models with vLLM CPU mode for local development testing


Jeffwan:
### 🚀 Feature Description and Motivation

We already have mocked app for most feature integration testing, however, this is still not that convenient in some cases. We should check whether it's possible to use small models like opt-125m with cpu only vLLM for testing

varungup90:
We can close this issue: https://github.com/vllm-project/aibrix/tree/main/development/vllm

PR #721, opened 2025-02-20
Add vllm cpu alternative for local development

varungup90:
https://github.com/vllm-project/aibrix/issues/417


development/vllm/README.md
```diff
@@ -0,0 +1,83 @@
+# vLLM CPU application
```
Jeffwan, line 1: We have a development doc in page https://github.com/vllm-project/aibrix/tree/main/docs/source/development we can consider to move some text to that part in future

development/vllm/README.md
```diff
@@ -0,0 +1,83 @@
...
+```shell
+kind create cluster --config=./development/vllm/kind-config.yaml
+```
+(Optional) Load container image to docker context
+
+> Note: If you are using Docker-Desktop on Mac, Kubernetes shares the local image repository with Docker.
+> Therefore, the following command is not necessary. Only kind user need this step.
+
+```shell
+docker pull aibrix/vllm-cpu-env:macos
```
Jeffwan, line 22: nit: Let's use some other formats like arch-version-date etc as the tag

Issue #749, opened 2025-02-26
xxhash64 Collision Risks and Fixes in aibrix Clusters​


kexinoh:
## Summary​
Maliciously crafted prompts may cause hash collisions, leading to prefix cache reuse, which could interfere with subsequent responses and result in unintended behavior.

## Details​
aibrix's prefix caching mechanism relies on the xxhash64 hash algorithm. Since xxhash64 uses a fixed seed value (e.g., seed=0) in its default configuration, its hash outputs are predictable. Attackers could exploit this determinism to craft specific inputs that generate hash collisions.

## Impact​
If aibrix's hash generation logic lacks randomization (e.g., no dynamic seeds or salting), attackers could leverage xxhash64's predictable nature to inject malicious inputs that collide with target prompts, causing cache pollution and subsequent response errors.

## Fix Recommendation​
In a cluster environment, this issue should be addressed by implementing ​statistically unique and cryptographically secure xxhash seed values.（not Like vllm https://github.com/advisories/GHSA-rm76-4mrf-v9r8 )This approach would:

Prevent attackers from predicting or forcing collisions.
Enable global cache sharing across the cluster while maintaining security.

PR #752, opened 2025-02-26
Use random seed for xxhash

varungup90:
## Pull Request Description
Generates random seed value for every gateway initialization

## Related Issues
Resolves: https://github.com/vllm-project/aibrix/issues/749


Jeffwan:
@varungup90 can you squash the error the pass DCO?

kexinoh:
Why not directly generate it using cryptographic randomness? That would be more compliant with the standards.

varungup90:
> Why not directly generate it using cryptographic randomness? That would be more compliant with the standards.

It requires same seed to generate same hash value for prompt's prefix.

kexinoh:
If you'd like, I can help adjust the code to use a cryptographically secure random number generator. Here's a gentle suggestion:

---
For security-focused scenarios in Go, it might be better to use the `crypto/rand` package instead of `math/rand`. You could try something like this:

```go
import (
    "crypto/rand"
    "encoding/binary"
)

func generateSecureSeed() (uint64, error) {
    var seed uint64
    if err := binary.Read(rand.Reader, binary.BigEndian, &seed); err != nil {
        return 0, err
    }
    return seed, nil
}
```
This complies with the random number generation specifications. In fact, `rand.New(rand.NewSource(time.Now().Unix()))` should never be used under any circumstances.

varungup90:
@kexinoh Sounds good, you can send a PR to add generateSecureSeed method in pkg/util.

PR #771, opened 2025-02-28
Reconfigure workload generator for predefined synthetic patterns

happyandslow:
This PR reconfigures workload generator to adopt predefined load types.

### DIFF
--- a/benchmarks/generator/workload_generator.py
+++ b/benchmarks/generator/workload_generator.py
@@ -17,6 +17,7 @@ from sample_request import (load_requests,
 from distribution import (generate_poisson_dist,
                           generate_token_len_from_percentiles,
                           to_fluctuate_pattern_config,
+                          user_to_synthetic_config,
                           )
                           
 from utils import (convert_to_stat_df,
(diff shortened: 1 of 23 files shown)


benchmarks/generator/workload_generator.py
```diff
@@ -242,32 +246,33 @@ def math_function(t, pattern_config, length, prev_value):
     previous_input_len = -1
     previous_output_len = -1
     ts = 0
-    base_req_id = 0
     
-    sharegpt_df = load_requests(dataset_path=prompt_file_path, tokenizer=tokenizer)
+    rps_dist = []
+    input_token_len_dist = []
+    output_token_len_dist = []
     while t < length:
```
gangmuk: where does `t` come from?

benchmarks/generator/workload_generator.py
```diff
@@ -242,32 +246,33 @@ def math_function(t, pattern_config, length, prev_value):
...
-            output_lens=[current_output_len] * current_concurrency, 
-            initial_err_perc=0.1,
-            err_step=0.05,
-            model=model,
-        )
-        workload.append({"timestamp": ts, "requests": concurrent_reqs})  
-        base_req_id += current_concurrency
+        current_input_len = current_input_len if current_input_len > 0 else 1
+        current_output_len = current_output_len if current_output_len > 0 else 1
+        rps_dist.append(current_concurrency)
```
gangmuk: Concurrency could be the misleading term here. As far as I know, concurrency or concurrent requests usually means (the number of) requests that are concurrently running in the system. For example, concurrency control is controlling that. By that definition, workload generator can't control concurrency. So what it meant was maybe a certain number of request per unit time? (RPS if it is second)

benchmarks/generator/workload_generator.py
happyandslow: updated

benchmarks/generator/workload_generator.py
happyandslow: updated

Issue #800, opened 2025-03-05
an endpoint to list all the models deployed in the AiBrix system


premdass:
### 🚀 Feature Description and Motivation

to discover all the models deployed in the AIBrixsystem, it will be useful to have a single endpoint which displays all the models like the openai /models endpoint

### Use Case

to discover all the models deployed in the AIBrixsystem, it will be useful to have a single endpoint which displays all the models like the openai /models endpoint

Jeffwan:
We get few users report this issue. This would be implemented shortly.  We talked about two different options 

1. AIBrix should provides an abstraction to provide **all** available models.  the challenge is current plugin design does not like to manipulate the response. Technically, it should aggregate the models (including lora adapter) from the cache. data plane traffic works out-of-box in multi-model scenarios.
2. refactor plugin to provide model scope /v1/models return results. That means it only care about single base + all the related adapters. but seems this way need more efforts and can not meet production users's need.

Currently, I prefer the 1st way and we need to do such trick for such api. It brings another problem on the api compatibility and we should revisit it now.  https://github.com/vllm-project/aibrix/issues/732

PR #802, opened 2025-03-06
Add /v1/models endpoint to gateway

varungup90:
## Pull Request Description
This change adds /models endpoint to the gateway. Right now response return list of model and lora adapters.

image

## Related Issues
Resolves: #800

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>


Jeffwan:
em. from the interface perspective, this is not compatible way to expose the api. Could we make it more compatible?

varungup90:
> this is not compatible way to expose the api

you mean change to /v1/models?

Jeffwan:
return response as well

cmd/metadata/main.go
```diff
@@ -17,14 +17,41 @@ limitations under the License.
...
 	"github.com/vllm-project/aibrix/pkg/utils"
+	"k8s.io/client-go/rest"
+	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 )
 
 func main() {
 	redisClient := utils.GetRedisClient()
 
+	fmt.Println("starting cache")
```
Xunzhuo: would be better to use `klog`

varungup90:
> return response as well

I have updated to /v1/models. For response, gateway will need to query each inference server's /v1/models endpoint and will require api-key. Figuring out api-key is the tricky part. I suggest, we merge the current change and will open another task to retrieve models information from inference server.

kerthcet:
It would be great to have a standard API schema for all the inference platforms to follow. Then it would be easy for us as well.

Jeffwan:
@kerthcet Agree. I notice many questions on related issues. Currently, openAI style API schema is preferred. However, even vLLM's implementation is not 100% compatible, completion or chat/completions are compatible. I will create an umbrella issue for the API compatibility. At least for short term, I think AIBrix gateway's api should be compatible with engine's and **do not** apply any modifications

config/gateway/gateway-plugin/gateway-plugin.yaml
```diff
@@ -62,6 +62,36 @@ spec:
...
+spec:
+  parentRefs:
+    - name: aibrix-eg
+  rules:
+    - matches:
+        - path:
+            type: PathPrefix
+            value: /v1/models
+      backendRefs:
+        - name: aibrix-metadata-service
```
Jeffwan, line 81: We will have batch feature supported as well. Seems we need to better design a "extension server".  At this moment, metadata sounds good

config/metadata/metadata.yaml
```diff
@@ -48,6 +48,10 @@ spec:
               value: aibrix-redis-master
             - name: REDIS_PORT
               value: "6379" 
+            # TODO: cache is shared across all control plane so add feature flags to enable metric pull
+            # for now setting to a 1 hr for metadata service
+            - name: AIBRIX_POD_METRIC_REFRESH_INTERVAL_MS
```
Jeffwan, line 53: does metadata service rely on any cache data?

Jeffwan: Just see below implementations. Actually, it still leverage cache information. not send request to every pods. In this case, we need to change model frequency to seconds or minute level. at least make sure new deployment metadata can be exposed shortly. 1hr is too long

pkg/metadata/handlers.go
```diff
@@ -17,36 +17,57 @@ limitations under the License.
...
 	return &http.Server{
 		Addr:    addr,
 		Handler: r,
 	}
 }
 
+// models returns base and lora adapters registered to aibrix control plane
+func (s *httpServer) models(w http.ResponseWriter, r *http.Request) {
+	models := s.cache.GetModels()
+	jsonBytes, err := json.Marshal(models)
```
Jeffwan, line 63: Let's make this compatible with vLLM's current implementation.

here's one example https://aibrix.readthedocs.io/latest/features/runtime.html#model-configuration-api

Jeffwan (changes requested):

config/gateway/gateway-plugin/gateway-plugin.yaml
varungup90: Previously, there was one reserved router (httproute) for endpoint "/" and extension server was registered to envoy gateway. All the requests target same httproute and gateway sends all requests to the extension server.

With this change, previous httproute for endpoint "/" is updated to target only "/v1/chat/completions" and /v1/completions" and other endpoint request such as /v1/batch will get 404 error. Secondly extension server's target is chagend from envoy gateway to updated httproute (with two completions endpoint). Basically now extension server only processes request for updated httproute.

Another httproute is added for "/v1/models" and requests are forwards to metadata service.

In future, for example we want to add "/v1/batch", new httproute will be added and new extension server to support that use case will be added. New extension server will only handle requests for batch. Also will need to refactor common code to share across different extension servers. Hope it helps.

config/metadata/metadata.yaml
varungup90: I mentioned in the previous reply that querying pods need api-key information which gateway does not have it. Till we figure out how to achieve that, using internal-cache to return models registered.

image


Frequecy referenced here is not model refresh. Internal cache code is shared across all components and this frequecy is for metrics refresh (check the env variable name). I have added a todo to make cache's background workers such as refresh metrics or request trace to me have enable/disable flags.

pkg/metadata/handlers.go
varungup90: No reservations in following open ai format in response as well. But retrieving model information from pod is tricky as api_key is not available at gateway.

varungup90:
> At least for short term, I think AIBrix gateway's api should be compatible with engine's and **do not** apply any modifications

Lets document places where compatibility is missing and make it high pri to fix it.

config/metadata/metadata.yaml
Jeffwan: Agree. I tend to zone out and that's just for metrics not related to model. It looks good to me

pkg/metadata/handlers.go
Jeffwan: We do not necessary need to fetch from pods, just keep the structure same might be helpful I think. value could be empty.

Jeffwan:
> Lets document places where compatibility is missing and make it high pri to fix it.

We should work on this in advance in a systematic way instead of relying on user reports. i am not sure if there's a way to verify it easily. We can use separate issue to track it

pkg/metadata/handlers.go
Jeffwan: In this case, user can not use sdk to get the response. We can not assume user just use curl

Jeffwan:
We can merge this now and follow up on /v1/model compatibility and all api compatibility issues.

Jeffwan (approved):

shenshouer:
Is this API included in version 0.2.1?

Jeffwan:
@shenshouer not yet. This is a feature and we didn't cherry-pick to release branch. You can use `nightly` or specific commit tag instead.

PR #862, opened 2025-03-14
Autoscaling Benchmark Initial

happyandslow:
## Pull Request Description
Supporting autoscaling experiment scripts


Jeffwan:
image Can these files be generated? If so, let's get ride of them in the source file

happyandslow:
> image Can these files be generated? If so, let's get ride of them in the source file

Yes -- they are generated by workload generator. However, if I remove them, then there will be missing dependencies in the running script.

Jeffwan:
@happyandslow you are supposed to write down the prerequisite to run scripts. Let's remove such dataset in the PR, only scripts to generate those PR are needed

Jeffwan:
@happyandslow is this PR still needed?

happyandslow:
> @happyandslow is this PR still needed?

Yes this PR corresponds to #666

Jeffwan:
@happyandslow is this PR still necessary?

Issue #892, opened 2025-03-21
[RFC]: New eviction strategy for prefix cache indexer


vie-serendipity:
### Summary

This RFC mainly aims to introduce new eviction policy to enhance the robustness of the prefix cache routing system.

### Motivation

The current prefix cache indexer uses a periodic eviction strategy, and the default time is 60min, the memory usage will gradually increase over time, which may cause OOM in large-scale scenarios (I haven't encountered it so far, just a guess based on the code). It is necessary to introduce some advanced strategies to control the size of the memory used.

### Proposed Change

1. Introduce a new interface in `aibrix/pkg/plugins/gateway/evictor` to keep it extensible
2. Make eviction strategy as configurable
3. Implement the earliest periodic eviction strategy
4. Implement the LRU eviction strategy and take it as default eviction strategy

Jeffwan:
Capacity based is eviction would be helpful for OOM issues. Thanks for bringing this up. We plan to refactor the prefix cache strategy in next two week. If you have interest to make some contribution, that would be perfect as well!

vie-serendipity:
@Jeffwan Sure, I'm interested in implementing new eviction strategy, could you assign it to me

Jeffwan:
@vie-serendipity thanks! I've finished the assignment

PR #905, opened 2025-03-26
[API] [Misc]: Support LRU cache with TTL for prefix cache indexer

vie-serendipity:
## Pull Request Description
[Please provide a clear and concise description of your changes here]
Implement LRU cache with TTL to support limited prefix cache, to prevent OOM.

## Related Issues
#892 

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>


pkg/plugins/gateway/cache/lru_store_test.go
```diff
@@ -0,0 +1,105 @@
...
+	// Test updating an existing key
+	store.Put("key1", "value1")
+	store.Put("key1", "value2")
+
+	if val, ok := store.Get("key1"); !ok || val != "value2" {
+		t.Errorf("expected value2, got %v", val)
+	}
+}
+
+func TestLRUStore_ConcurrentEvictions(t *testing.T) {
```
varungup90, line 75: Can you also add a test to validate least recently used key is evicted?

pkg/plugins/gateway/prefixcacheindexer/hash.go
```diff
@@ -23,24 +23,42 @@ import (
...
 
 	"github.com/cespare/xxhash/v2"
+	"github.com/vllm-project/aibrix/pkg/plugins/gateway/cache"
 	"github.com/vllm-project/aibrix/pkg/utils"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 )
 
 const (
+	defaultPrefixCacheBlockNumber            = 100
```
varungup90: increase the default block size to 1 million.

pkg/plugins/gateway/prefixcacheindexer/hash.go
```diff
@@ -23,24 +23,42 @@ import (
...
 	defaultPrefixCacheBlockSize              = 16
-	defaultPrefixCacheEvictionInternalInMS   = 50
-	defaultPrefixCacheEvictionDurationInMins = 60
+	defaultPrefixCacheEvictionInternalInMS   = 1000 // 1 second
```
varungup90: you can rename the variable to defaultPrefixCacheEvictionInternalInSec

pkg/plugins/gateway/cache/lru_store.go
```diff
@@ -0,0 +1,159 @@
...
+package cache
+
+import (
+	"sync"
+	"time"
+)
+
+type LRUStore[K comparable, V any] struct {
+	sync.Mutex
+	freeTable map[K]*entry[K, V]
```
varungup90, line 32: Right now the structure is as follows

hash1 -> model1 -> pod1  + pod1 last access time 
           -> model1 -> pod2  + pod2 last access time
           -> model2 -> pod3 + pod3 last access time
           -> blocks last access time 

Right now eviction is based of last blocks access time. I kept the eviction like this for now to keep it simple. Eventually we want to move to only delete pod1 + its last access time rather than restricting to only delete entire block.

Another requirement is that, for one request it creates 100 hashes. How to make sure all hashes are deleted for that request together. Thats how vllm implements block eviction.

varungup90: Ask is that can you share some design ideas on how to extend it with new implementation?

pkg/plugins/gateway/cache/lru_store.go
```diff
@@ -0,0 +1,159 @@
...
+	e.Lock()
+	for key, entry := range e.freeTable {
+		if now.Sub(entry.lastAccessTime) > e.ttl {
+			keysToEvict = append(keysToEvict, key)
+		}
+	}
+	e.Unlock()
+
+	for _, key := range keysToEvict {
+		e.Lock()
```
Jeffwan, line 121: here. I am curious whether it need key level lock? could it be for loop level?

pkg/plugins/gateway/cache/lru_store.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	if e.next != nil {
+		e.next.prev = e.prev
+	}
+	e.prev = nil
+	e.next = nil
+}
+
+func (l *list[K, V]) removeTail() *entry[K, V] {
+	entry := l.tail.prev
```
Jeffwan, line 170: if this is empty, tail.prev==head. does it still remove elements?
Should we check 

```
	if l.tail.prev == l.head {
		return nil
	}
```

pkg/plugins/gateway/cache/lru_store.go
```diff
@@ -0,0 +1,159 @@
...
+
+package cache
+
+import (
+	"sync"
+	"time"
+)
+
+type LRUStore[K comparable, V any] struct {
+	sync.Mutex
```
Jeffwan: RWMutex for `Get()` or `Len()` Operations?

pkg/plugins/gateway/prefixcacheindexer/hash.go
```diff
@@ -87,33 +105,27 @@ func getPrefixCacheEvictionDuration() time.Duration {
...
 
 type PrefixHashTable struct {
-	mu     sync.RWMutex
-	blocks map[uint64]Block
-	hash   *xxhash.Digest
-	seed   uint64
+	mu    sync.RWMutex
+	hash  *xxhash.Digest
+	seed  uint64
+	store cache.Store[uint64, Block]
```
Jeffwan, line 111: Using cache interface here to replace blocks looks better. this is great

pkg/plugins/gateway/prefixcacheindexer/hash.go
Jeffwan: this value determines the size of the cache, I feel 100 is too small. We need to change default one to large values.  Later, the prefix will be changed to use character instead of word or tokens, results in more blocks.  20000 token request end up with 1250 blocks. we can roughly calculate the memory size (like 1GiB) and see how much block it can persist. 

More intelligent way is to intelligent check container memory limits, it can use injected envs to decide removeTail() but not just based on cap. htis can be left as separate task. we need to have a reasonable default value first.

```
  - name: MY_MEM_LIMIT
    valueFrom:
      resourceFieldRef:
        resource: limits.memory
```

pkg/plugins/gateway/cache/lru_store_test.go
vie-serendipity: I think TestLRUStore_TTL covers the case that validates least recently used key is evicted.

pkg/plugins/gateway/cache/lru_store.go
vie-serendipity: For now `removeTail` will only be called when size is greater than cap, so I think this situation won't occur. But I think adding such a check would make the code more robust.

pkg/plugins/gateway/cache/lru_store.go
vie-serendipity: It makes sense.

pkg/plugins/gateway/cache/lru_store.go
vie-serendipity: Sorry for the late reply.

> Eventually we want to move to only delete pod1 + its last access time rather than restricting to only delete entire block.

I think this is reasonable.

> Another requirement is that, for one request it creates 100 hashes. How to make sure all hashes are deleted for that request together. Thats how vllm implements block eviction.

Just to confirm, I understand that after vllm processes the request, it frees all the blocks together, so you mean that the router implementation should also free them together.

But I think this involves two layers, and in reality, the router and the kv cache don't have actual interaction. Even if the lastAccessedTime is very close, it doesn't guarantee that there's already a cache for the current request on the node. Hence, I understand that the router here implements a best-effort approach. In the current implementation, selecting the pod with the most recent access would be the most likely to have the cached KV.

I think for precise routing, integrating with a distributed kv cache could be a solution, but that would come with a significant cost. What do you think? cc @Jeffwan

pkg/plugins/gateway/prefixcacheindexer/hash.go
vie-serendipity: Conservatively estimating, I think 1GiB can accommodate approximately 200 thousand, so I'll initially set it to 200,000.

> More intelligent way is to intelligent check container memory limits, it can use injected envs to decide removeTail() but not just based on cap.

This is a good idea, heuristic estimation would be more user-friendly.

pkg/plugins/gateway/cache/lru_store.go
vie-serendipity: I don't quite understand your question. Let me clarify that the main reason for traversing to obtain the key without performing any operation first is to avoid locking too long time and blocking the router.

pkg/plugins/gateway/prefixcacheindexer/hash.go
vie-serendipity: okay

pkg/plugins/gateway/prefixcacheindexer/hash.go
vie-serendipity: I'll set it conservatively to 200,000 for now. Refer to https://github.com/vllm-project/aibrix/pull/905#discussion_r2020870210

Jeffwan: sounds good

pkg/plugins/gateway/cache/lru_store.go
Jeffwan: there was a typo in my original message. I mean `lock inside loop` or `lock before loop`. I think the `lock inside loop` helps in the scenarios that we have many eviction keys and there're high volume concurrency request modifying `freeTable` and `lruList`. in other cases, `lock before loop` looks better. I do not have strong option at this moment. I think long eviction window will accumulate more keys, it looks good to me

pkg/plugins/gateway/cache/lru_store.go
Jeffwan: > Hence, I understand that the router here implements a best-effort approach
yes, exactly. 

> I think for precise routing, integrating with a distributed kv cache could be a solution, but that would come with a significant cost.

This is not in the current scope, because of few reasons. 1. Both vLLM and current distributed kv cache solution (like vineyard we are using) do not provide interfaces to expose kv details. distributed kv cache has metadata server but that's for internal communication.  2. Even they provides a way, fetch the source of truth from engine or kv cache service will be costly. I agree that we can revisit this feature later. 


> Another requirement is that, for one request it creates 100 hashes. How to make sure all hashes are deleted for that request together

the cache hit follow the common prefix pattern. the earlier blocks alway have fresh "last access time". In that case, if one block is deleted, the blocks after that block will be deleted for sure. that means even map will support such case out of box? correct me if I understand incorrectly.

pkg/plugins/gateway/prefixcacheindexer/hash.go
Jeffwan: I think we will always use seconds for this parameter. let's change to `XXXInSec`

Jeffwan:
I think most feedback are addressed. Please double check the remaining issues, then we should be good to go. @vie-serendipity @varungup90

pkg/plugins/gateway/prefixcacheindexer/hash.go
```diff
@@ -57,18 +75,18 @@ func getPrefixCacheBlockSize() int {
 }
 
 func getPrefixCacheEvictionInterval() time.Duration {
-	value := utils.LoadEnv("AIBRIX_PREFIX_CACHE_EVICTION_INTERVAL_MS", "")
+	value := utils.LoadEnv("AIBRIX_PREFIX_CACHE_EVICTION_INTERVAL_Secs", "")
```
varungup90: nit: Secs -> SECONDS or SECS (prefer first one)

pkg/plugins/gateway/prefixcacheindexer/hash.go
```diff
@@ -87,33 +105,27 @@ func getPrefixCacheEvictionDuration() time.Duration {
...
 func NewPrefixHashTable() PrefixCacheIndexer {
 	r := rand.New(rand.NewSource(time.Now().Unix()))
 	seed := r.Uint64()
 	instance := &PrefixHashTable{
-		blocks: map[uint64]Block{},
-		hash:   xxhash.NewWithSeed(seed),
-		seed:   seed,
+		hash: xxhash.NewWithSeed(seed),
+		seed: seed,
+		store: cache.NewLRUStore[uint64, Block](getPrefixCacheBlockNumber(),
```
varungup90: nit: since variables are already assigned, use them directly.

pkg/plugins/gateway/prefixcacheindexer/hash.go
```diff
@@ -134,17 +146,15 @@ func (c *PrefixHashTable) MatchPrefix(tokens []byte, model string, pods []*v1.Po
 		_, _ = c.hash.Write(tokens[i:end])
 		prefixHash := c.hash.Sum64()
 		c.hash.ResetWithSeed(c.seed)
-
-		block, ok = c.blocks[prefixHash]
+		block, ok = c.store.Get(prefixHash)
```
varungup90, line 150: store is thread-safe, so we can remove lock from line 134 now. Right?

pkg/plugins/gateway/prefixcacheindexer/hash_test.go
```diff
@@ -63,7 +64,7 @@ func Test_PrefixHashTableE2E(t *testing.T) {
 	assert.Equal(t, 0, len(unMatchedTokens))
 	assert.Equal(t, "p1", matchPods[0].Name)
 
-	cache.Evict(time.Now().Add(60 * time.Minute))
+	time.Sleep(25 * time.Second)
```
varungup90: In the test, we can not sleep for 25 seconds. Can you add a mock for time or update the env variable in test for lower value.

pkg/plugins/gateway/cache/lru_store.go
```diff
@@ -0,0 +1,165 @@
...
+func (e *LRUStore[K, V]) Len() int {
+	e.RLock()
+	defer e.RUnlock()
+	return len(e.freeTable)
+}
+
+func (e *LRUStore[K, V]) evict(now time.Time) {
+	var keysToEvict []K
+
+	e.Lock()
```
varungup90: use read lock

varungup90:
Overall LGTM. Some nits, once you fix them ping me. I will merge the change. Thanks!

pkg/plugins/gateway/prefixcacheindexer/hash.go
vie-serendipity: done

pkg/plugins/gateway/prefixcacheindexer/hash.go
vie-serendipity: done

pkg/plugins/gateway/prefixcacheindexer/hash.go
vie-serendipity: Store is thread-safe, but hash is currently not thread-safe. The current lock scope is too broad, I narrowed it down.

pkg/plugins/gateway/prefixcacheindexer/hash_test.go
vie-serendipity: done

pkg/plugins/gateway/cache/lru_store.go
vie-serendipity: done

pkg/plugins/gateway/cache/lru_store.go
vie-serendipity: I agree with your point.

Jeffwan:
@varungup90 Seems all comments are addressed. I do not have further feedback as well

pkg/plugins/gateway/prefixcacheindexer/hash.go
```diff
@@ -87,40 +105,33 @@ func getPrefixCacheEvictionDuration() time.Duration {
...
-			instance.Evict(time.Now())
-		}
-	}()
-
 	return instance
 }
 
 // returns matchedTokens, unMatchedTokens, matchedPods
 func (c *PrefixHashTable) MatchPrefix(tokens []byte, model string, pods []*v1.Pod) ([]byte, []byte, []*v1.Pod) {
-	c.mu.Lock()
```
varungup90, line 122: @vie-serendipity I assumed LRU store is thread-safe and removing this lock was fine. While working on PR #933 I came across the bug for concurrent read/write for blocks.

I have added the lock back, if you get a chance, can you make LRU store thread-safe? Thank you.

vie-serendipity: okay, I will locate the bug and fix it.

vie-serendipity: I'm uncertain whether you encountered the bug after completely removing all locks from the `PrefixHashTable` implementation.
If so, I found the map obtained from the LRU store is not thread-safe, which should be locked when modifying it.
https://github.com/vllm-project/aibrix/blob/6c325828b22098ef17240c7fbd50d032ce93449c/pkg/utils/prefixcacheindexer/hash.go#L106-L122

varungup90: In PR: 905, lock was removed only from MatchPrefix (AddPrefix had the lock). I encountered same issue that block map is not cloned when read from LRU store.

I propose that we make LRU store thread safe. There are two options
1) Block map returned from LRUStore is cloned so that way locks can be removed from hash table implementation.
2) Use of sync map in LRU store + clone block map on get function. This is preferable because it will remove the lock from LRUStore as well.

vie-serendipity: Thanks for your explanation, I will submit a PR to fix it.

PR #919, opened 2025-03-28
Enable parallel client using thread pool in benchmark client

happyandslow:
## Pull Request Description
This PR resolves clien-side bottleneck issue by initializing parallel AsyncOpenAI client. 
- Spawning requests using thread pool.
- Adding timeout and retry config to clients

### DIFF
--- a/benchmarks/client/client.py
+++ b/benchmarks/client/client.py
@@ -6,17 +6,42 @@ import openai
 import json
 import io
 import traceback
+import threading
+from queue import Queue
+from concurrent.futures import ThreadPoolExecutor
 
 
 from typing import List
 from utils import (load_workload, wrap_prompt_as_chat_message)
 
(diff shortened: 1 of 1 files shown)


benchmarks/client/client.py
```diff
@@ -6,25 +6,50 @@
...
 
 
 from typing import List
 from utils import (load_workload, wrap_prompt_as_chat_message)
 
 logging.basicConfig(level=logging.INFO)
+task_queue = Queue()
+thread_pool_size = 8
+
+def worker(client, model, send_request_func, output_file):
```
Jeffwan, line 21: is multi-threading  necessary for this client?

Jeffwan:
If this change is for some issues, I suggest to link issues etc

Jeffwan (approved):

PR #918, opened 2025-03-28
[Misc] Add pods stats example: running requests.

zhangjyr:
## Pull Request Description
This PR is an example of adding real-time pod stats: running requests. The real-time stat is synced to the metric store for reading.

**Important: Before submitting, please complete the description above and review the checklist below.**

---

<details>
<summary><strong>Contribution Guidelines (Expand for Details)</strong></summary>

<p>We appreciate your contribution to aibrix! To ensure a smooth review process and maintain high code quality, please adhere to the following guidelines:</p>

<h3>Pull Request Title Format</h3>
<p>Your PR title should start with one of these prefixes to indicate the nature of the change:</p>
<ul>
    <li><code>[Bug]</code>: Corrections to existing functionality</li>
    <li><code>[CI]</code>: Changes to build process or CI pipeline</li>
    <li><code>[Docs]</code>: Updates or additions to documentation</li>
    <li><code>[API]</code>: Modifications to aibrix's API or interface</li>
    <li><code>[CLI]</code>: Changes or additions to the Command Line Interface</li>
    <li><code>[Misc]</code>: For changes not covered above (use sparingly)</li>
</ul>
<p><em>Note: For changes spanning multiple categories, use multiple prefixes in order of importance.</em></p>

<h3>Submission Checklist</h3>
<ul>
    <li>[ ] PR title includes appropriate prefix(es)</li>
    <li>[ ] Changes are clearly explained in the PR description</li>
    <li>[ ] New and existing tests pass successfully</li>
    <li>[ ] Code adheres to project style and best practices</li>
    <li>[ ] Documentation updated to reflect changes (if applicable)</li>
    <li>[ ] Thorough testing completed, no regressions introduced</li>
</ul>

<p>By submitting this PR, you confirm that you've read these guidelines and your changes align with the project's contribution standards.</p>

</details>

Issue #761, opened 2025-02-27
Stateful information sync for ext-proc Instances


Jeffwan:
### 🚀 Feature Description and Motivation

Our external process in the Envoy Gateway tracks request routing using prefix cache awareness, making it stateful. To ensure consistency and availability across multiple instances, we need to design a state synchronization mechanism.

- Current ext-proc builds a hash blocks/radix tree to track requests.
- Multiple instances need to share this prefix cache state and avoid data inconsistencies between instances.
- Ensuring minimal performance impact while syncing state.

external distributed store/election/event-driven sync etc seems a little bit heavy. We need to consider the lightweight approach for implementation.

### Use Case

unblock prefix-cache awareness request scheduling in HA case

firebook:
I think we can use Redis as the state storage to cache prefix-based data and track real-time request counts. This would be a relatively straightforward solution. Moreover, since the Redis component has already been introduced in the current rate limiting implementation, this approach would not introduce any new dependencies to the system.

Additionally, if users prefer not to use Redis for storage, we can still maintain the in-process memory storage approach. This remains developer-friendly for testing or quick start scenarios.

firebook:
Any update?

firebook:
@Jeffwan @varungup90 any update?

varungup90:
@firebook I haven't got the opportunity to work on it yet. It is in my high priority list. 

If you are interested, please take the lead to propose implementation design, I can help you in the process. Thank you.

justadogistaken:
@Jeffwan @varungup90 
Any update?
I guess we fail to deploy multiple instances util scheduling info is made stateful.

justadogistaken:
Would it be ok to deploy multiple instances(3+), but only one instance can do scheduling? Relying etcd or redis to select master.

justadogistaken:
Or could we design a two-level scheduler? first level to route traffic to sub-scheduler with model service. The Second level router schedules traffic by strategy(load/prefix cache)

PR #933, opened 2025-04-02
Performance improvements for prefix cache routing

varungup90:
## Pull Request Description
- Use running request tracked within gateway plugin. Previous approach of fetching active requests from inference engine was prone to burst request.
- Add load aware in pre-match-prefix and post-match-prefix. Pre match prefix uses absolute difference in max and min running requests count. Post match prefix uses standard deviation to restrict over-loading on prefix match pod. Lastly as a fallback least-request routing approach is used.

Env
- Model: llama3-8B
- Cluster: L20*8
- Input dataset: `python3 benchmark/generator/generate_realistic_prefix_share_workload.py`

Output: prefix_cache is ~45% faster compared to random routing.

Output summaries and detailed pdf report

**random routing**
image
[random.pdf](https://github.com/user-attachments/files/19575812/random.pdf)

**least-request**
image
[least-request.pdf](https://github.com/user-attachments/files/19575813/least-request.pdf)

**prefix_cache**
image
[prefix_cache.pdf](https://github.com/user-attachments/files/19575808/prefix_cache.pdf)

Blips occur when prefix_match_pods are overloaded and fallback to non prefix cache pod happens.
image

**prefix_cache with 2*std_dev load factor**
image
[prefix_cache_load_factor_2.pdf](https://github.com/user-attachments/files/19595038/prefix_cache_load_factor_2.pdf)

If we use second standard deviation i.e. pack more requests on pods with prefix cache present then #blips reduce as expected and overall P50 or P99 TTFT improve ~10% compare to first standard deviation. 
image

## Related Issues

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>


Jeffwan:
image
Seems most of the request hit the cache through the `prefix-cache`. there're some requests in the middle results in higher TTFT. Not sure if there's room to keep improving it.

config/gateway/gateway-plugin/gateway-plugin.yaml
```diff
@@ -58,6 +58,8 @@ spec:
               value: "6379" 
             - name: AIBRIX_POD_METRIC_REFRESH_INTERVAL_MS
               value: "50"
+            - name: AIBRIX_PREFIX_CACHE_BLOCK_SIZE
```
Jeffwan, line 63: 1. Why is there a mismatch between the values in the default and release manifests?

2. Why not keep consistent with vLLM block size?  

using higher block size results in lower hits

pkg/cache/cache_metrics.go
```diff
@@ -329,9 +329,6 @@ func (c *Store) queryUpdatePromQLMetrics(metric metrics.Metric, queryLabels map[
 // TODO: replace in-place metric update podMetrics and podModelMetrics to fresh copy for preventing stale metric keys
 func (c *Store) updatePodRecord(pod *Pod, modelName string, metricName string, scope metrics.MetricScope, metricValue metrics.MetricValue) error {
 	if scope == metrics.PodMetricScope {
-		if modelName != "" {
```
Jeffwan, line 332: is this change in the current PR scope?

pkg/plugins/gateway/algorithms/least_request.go
```diff
@@ -112,8 +76,41 @@ func (r leastRequestRouter) Route(ctx *types.RoutingContext, pods types.PodList)
...
+	if len(targetPods) > 0 {
+		targetPod, _ = utils.FilterPodByName(targetPods[rand.Intn(len(targetPods))], readyPods)
+	}
+	return targetPod
+}
+
+func getRequestCounts(cache cache.Cache, readyPods []*v1.Pod) map[string]int {
+	podRequestCount := map[string]int{}
+	for _, pod := range readyPods {
+		runningReq, err := cache.GetMetricValueByPod(pod.Name, metrics.RealtimeNumRequestsRunning)
```
Jeffwan, line 113: do you observe any issue on waiting and swapped?  why are they removed from the list?

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -18,113 +18,205 @@ package routingalgorithms
...
 	"fmt"
+	"math"
 	"math/rand"
+	"sort"
 	"strconv"
 
-	"github.com/vllm-project/aibrix/pkg/plugins/gateway/prefixcacheindexer"
-	"github.com/vllm-project/aibrix/pkg/plugins/gateway/prefixcacheindexer/tokenizer"
+	"github.com/vllm-project/aibrix/pkg/cache"
+	"github.com/vllm-project/aibrix/pkg/plugins/gateway/algorithms/prefixcacheindexer"
```
Jeffwan: The current naming feels a bit unclear. Previously, we used prefixcacheindexer to refer to the component handling indexing. 

I think we need better structure here.

Option 1: We can have a folder for prefixcache and move files under `prefixcacheindexer` directly under it or just move the `indexer` folder under it.  

```
gateway/
├── algorithms/
│   └── prefixcache/
│       ├── indexer/
```

Option 2: Keep it as original folder, the algorithm folder still leverage outside util to extend the algorithm part. 

```
gateway/
├── algorithms/
 │       ├── prefix_cache.go
 │       ├── prefix_cache_and_load.go
 │       ├── prefix_cache_distributed_hash.go
```
\

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -18,113 +18,205 @@ package routingalgorithms
...
+	"github.com/vllm-project/aibrix/pkg/plugins/gateway/algorithms/tokenizer"
```
Jeffwan: Do you think we should keep tokenizer folder under `gateway/tokenizer` instead of `gateway/algorithms/tokenizer`? I feel it's more of a utility that could be reused across different components—not just the algorithms. For example, features like request estimation based on prompt size could live in a shared/common layer. The algorithms package would just consume those interfaces. It might be cleaner to move general-purpose logic out of the algorithm-specific scope.

pkg/plugins/gateway/algorithms/prefixcacheindexer/cache/lru_store.go
Jeffwan, line 1: em. same here. could it leave it in original package?

pkg/plugins/gateway/gateway.go
```diff
@@ -133,7 +138,22 @@ func (s *Server) selectTargetPod(ctx *types.RoutingContext, pods types.PodList)
...
-	return router.Route(ctx, pods)
+
+	if pods.Len() == 0 {
+		return "", fmt.Errorf("no pods to forward request")
+	}
+	readyPods := utils.FilterRoutablePods(pods.All())
+	if len(readyPods) == 0 {
+		return "", fmt.Errorf("no ready pods available for fallback")
+	}
+	if len(readyPods) == 1 {
```
Jeffwan, line 149: do you mean if `len(readyPods) >= 1` or `==`?

varungup90:
> image Seems most of the request hit the cache through the `prefix-cache`. there're some requests in the middle results in higher TTFT. Not sure if there's room to keep improving it.

Yea, I am debugging that part. One of the scenario, I came across is as follows.

T1: Request (R1) - Prefix_Match_Pods: [P1: 97%, P2: 97%], Active_Request_Count: [P1: 2, P2: 2, P3: 0] 
For Request (R1) - In post prefix_match, load is imbalanced as P1 & P2 request count is higher > mean + std_dev (for all requests) and fallback to least_request count will happen. Selected target_pod = P3.

Before R1 completes, R2 and R3 arrive with prefix_match_pods: [P3: 90%], active_request_count: [P1: 2, P2: 2, P3: 1 (or 2 for R3)]
For R2 & R3,  P3 is selected. Behavior is as expected for this scenario. I am checking all scenario, to see for potential improvements.

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -18,113 +18,205 @@ package routingalgorithms
...
-	RouterPrefixCache types.RoutingAlgorithm = "prefix-cache"
-)
-
-func init() {
-	RegisterDelayedConstructor(RouterPrefixCache, NewPrefixCacheRouter)
-}
-
 const (
-	defaultPrefixCacheMatchThresholdPercent = 50
+	defaultPodRunningRequestImbalanceAbsCount = 8
```
Jeffwan, line 35: what's the magic number from?

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -18,113 +18,205 @@ package routingalgorithms
...
 var (
-	prefixCacheMatchThresholdPercent = getPrefixCacheMatchThresholdPercent()
+	RouterPrefixCache                  types.RoutingAlgorithm = "prefix-cache"
+	podRunningRequestImbalanceAbsCount int                    = getPodRequestImbalanceAbsCount()
 )
 
-func getPrefixCacheMatchThresholdPercent() int {
-	value := utils.LoadEnv("AIBRIX_PREFIX_CACHE_MATCH_THRESHOLD_PERCENT", "")
+func getPodRequestImbalanceAbsCount() int {
+	value := utils.LoadEnv("AIBRIX_PREFIX_CACHE_POD_RUNNING_REQUEST_IMBALANCE_ABS_COUNT", "")
```
Jeffwan: Let's update the docs for new added envs

Jeffwan: Different GPU has different compute capacity, using absolute numbers looks like not that scalable. How can we generate this number?

pkg/plugins/gateway/algorithms/tokenizer/characters.go
```diff
@@ -16,16 +16,12 @@ limitations under the License.
...
-	sTokenizer := NewStringTokenizer()
+func NewCharacterTokenizer() Tokenizer {
+	return &characterTokenizer{}
+}
 
-	bArr, err := sTokenizer.TokenizeInputText("this is first message")
-	assert.NoError(t, err)
-	assert.Equal(t, []byte{116, 104, 105, 115, 105, 115, 102, 105, 114, 115, 116, 109, 101, 115, 115, 97, 103, 101}, bArr)
+func (s characterTokenizer) TokenizeInputText(text string) ([]byte, error) {
+	return []byte(text), nil
```
Jeffwan: use `[]rune` here, otherwise, byte is too fine grain and will break the unicode

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -18,113 +18,205 @@ package routingalgorithms
...
+			maxValue = value
+		}
+	}
+
+	if maxValue-minValue > podRunningRequestImbalanceAbsCount {
+		targetPod, _ := utils.FilterPodByName(targetPods[rand.Intn(len(targetPods))], readyPods)
+		return targetPod, true
+	}
+
+	return nil, false
```
Jeffwan: better to define an `imbalanced` variable to represent true/false for readability

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -18,113 +18,205 @@ package routingalgorithms
...
+func getTargetPodOnLoadImbalance(cache cache.Cache, readyPods []*v1.Pod) (*v1.Pod, bool) {
+	targetPods := []string{}
+	minValue := math.MaxInt32
+	maxValue := math.MinInt32
+
+	podRequestCount := getRequestCounts(cache, readyPods)
+	for podname, value := range podRequestCount {
+		if value <= minValue {
+			minValue = value
+			targetPods = append(targetPods, podname)
```
Jeffwan: in this case, If the podrequest count gives 5,4,3,2,1, your targetPods may get all of them. 

It's better to use a fixed value, and make large request count pod name can be evict. 2 or 3 makes sense (from Power of Two Choices https://smallrye.io/smallrye-stork/1.1.1/load-balancer/power-of-two-choices/)

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -18,113 +18,205 @@ package routingalgorithms
...
+		if matchedPods[podnames[i]] == matchedPods[podnames[j]] {
+			return podRequestCount[podnames[i]] < podRequestCount[podnames[j]]
+		}
+		return matchedPods[podnames[i]] > matchedPods[podnames[j]]
+	})
+
+	// select targetpod with highest %prefixmatch and request_count within stddev
+	for _, podname := range podnames {
+		reqCnt := float64(podRequestCount[podname])
+		if reqCnt <= meanRequestCount+stdDevRequestCount {
```
Jeffwan: can we run some unit test to approve this is efficiency? do we need some factor on `std` etc?

config/gateway/gateway-plugin/gateway-plugin.yaml
varungup90: Current assumption for setting block size is based off gateway's default tokenizer - character based.

1. default is used for local development, hence lower block size because local test strings will have shorter length. For release, increased the block size for actual dataset.

2. vLLM block size is based of tokens where one token can represent multiple chars. For using character based tokenizer, block size of 16 is too low.

Better alternative is to add documentation,
character based tokenizer, default: 16 and release: 64 to 128
tiktoken based tokenizer, default: 4 and release: 16.

pkg/cache/cache_metrics.go
varungup90: yes. Running request count is now tracked in gateway from PR: #918 and running request count is added in the metric store.

Since running request are tracked at pod level and updatePodRecord with scope PodMetricsScope does not require check for model name. @zhangjyr Correct me if I am wrong.

pkg/plugins/gateway/algorithms/least_request.go
varungup90: Tracking running request count within gateway was introduced in PR: #918. Changed least_request to use running_requests tracked by gateway and remove dependency of fetching metrics (running, swapped and waiting) from vllm. 

Reason to not use vllm metrics is that it is prone to burst requests.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: I see two options, 
1) Right now only gateway algorithms use tokenizer, thats why I kept it under gateway/algorithms folder. In future as new feature is built outside the scope of gateway/algorithms to use tokenizer then we can move tokenizer folder as necessary. Preemptively moving tokenizer folder under plugins/gateway does not seem right.

2) Second alternative is that we can move tokenizer under pkg/utils.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: For algorithms, we should keep them as option 2.
```
gateway/
├── algorithms/
 │       ├── prefix_cache.go
 │       ├── prefix_cache_and_load.go
 │       ├── prefix_cache_distributed_hash.go
```

We can refactor prefixcacheindexer which continues to have hash & tree based implementation which are specific to prefix_cache usecase and move prefixcacheindex/cache/lru_store files to util package for any future usecase. Final outcome will look like this

```
gateway/
├── algorithms/
 |       ├── prefixcacheindexer
 |                |── hash.go
 |                |── tree.go           
 │       ├── prefix_cache.go
 │       ├── prefix_cache_and_load.go
 │       ├── prefix_cache_distributed_hash.go

utils
├──lrustore
│      ├── lru_store.go            
```

pkg/plugins/gateway/algorithms/prefixcacheindexer/cache/lru_store.go
varungup90: from previous reply. lru_store is only used for prefix cache usecase, so better it be part of prefixcacheindexer package or moved to pkg/utils. It does seem right have it as part of pkg/plugins/gateway package.

pkg/plugins/gateway/gateway.go
varungup90: `==`.

If there is only one ready pod, then no need to running routing algorithm. Just return that ready pod.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: Its a magic number, no theory behind deriving it. While reading other open-source routers, I came across sglang implementation and it uses abs value of 32 for load balancing. 

It is configurable value and leave it to user to optimize it based off their workload and GPU type.

For 4K input token and 2k output token, 8 provided a appropriate balance for A10 and A100. Higher number than 8, I found results to degrade.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: We can leverage GPU optimizer to generate recommendations for abs numbers to define imbalance.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: Good catch, I fixed this bug in the least request. I will fix it here as well.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: If I am understanding correctly, you question is why one std_dev and it can potentially be 2 or 3 standard deviations as well?

Unit test can not prove the efficiency, because higher the standard deviation is more requests can be packed on same pod with prefix_match. 

One alternative is to have add configurable multiply factor to standard deviation and default value can be 1.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: I did not get the comment, this is a helper method, where to define the imbalance variable?

pkg/plugins/gateway/algorithms/tokenizer/characters.go
varungup90: Interface for tokenizer is []byte. Why will conversion to byte array break unicode?

pkg/cache/cache_metrics.go
zhangjyr: Yes, the modelName check is not necessary here. However, I think the check can be used as a validation to make sure users do mean pass metrics.PodMetricScope instead of metrics.PodModelMetricScope.

varungup90:
> > image Seems most of the request hit the cache through the `prefix-cache`. there're some requests in the middle results in higher TTFT. Not sure if there's room to keep improving it.
> 
> Yea, I am debugging that part. One of the scenario, I came across is as follows.
> 
> T1: Request (R1) - Prefix_Match_Pods: [P1: 97%, P2: 97%], Active_Request_Count: [P1: 2, P2: 2, P3: 0] For Request (R1) - In post prefix_match, load is imbalanced as P1 & P2 request count is higher > mean + std_dev (for all requests) and fallback to least_request count will happen. Selected target_pod = P3.
> 
> Before R1 completes, R2 and R3 arrive with prefix_match_pods: [P3: 90%], active_request_count: [P1: 2, P2: 2, P3: 1 (or 2 for R3)] For R2 & R3, P3 is selected. Behavior is as expected for this scenario. I am checking all scenario, to see for potential improvements.

@Jeffwan I tried with second standard deviation to load balance for prefix matched pods and TTFT improves 10% more, number of blips reduce. Initially my concern was that packing more requests on same pod can have adverse effect but since these extra packed requests already have kv cache so overall performance is not negatively affected.

pkg/cache/cache_metrics.go
varungup90: For the second part, to validate model name is indeed empty when scope is metrics.PodMetricScope. There are multiple calls to this function directly or indirectly and ensuring each caller sets model name as empty is quite an overhead. Even if model name is not empty, it wont affect the functionality.

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: Added a factor for std deviation. For benchmark test, a factor of 2 improves the TTFT by 10% compare to factor of 1, update the PR description with details.

pkg/plugins/gateway/algorithms/tokenizer/characters.go
Jeffwan: yeah Characters like emojis and chinese characters can span multiple bytes, so splitting by byte may break a character in the middle—it's too fine-grained for those cases. Using []rune preserves character boundaries and can still be converted to []byte if needed. The interface side should be fine with this for now

config/gateway/gateway-plugin/gateway-plugin.yaml
Jeffwan: 1. I feel we do not necessarily need to have two set of configuration for dev and release for this setting. keep them same simplify lots of things including support or docs.

2. 128 character size makes sense.

pkg/cache/cache_metrics.go
Jeffwan: Is it related to prefix-cache performance improvement? I feel this is a separate change? I am ok to leave it here, just like to confirm that

pkg/plugins/gateway/algorithms/tokenizer/characters.go
varungup90: For first point of character break, from prefix match point it will not affect final outcome in grand scheme. For second point of converting []rune to []bye, it will add quite a overhead. Same was the problem with using string split and then converting to []byte.

pkg/cache/cache_metrics.go
varungup90: Reason to include this change is

1) With this implementation, running requests are now tracked by gateway and updated this updatePodRecord method.
2) This check is incorrect, there is no need to enforce a check for modelname != "" in PodMetricScope, and if it is enforced then all callers need to be updated or add check to make sure model name field passed is empty.

config/gateway/gateway-plugin/gateway-plugin.yaml
varungup90: Increased the block size to 128 for release.

In vllm if token length is less than block size then it is ignored i.e. hash block is not created. I also moved prefix-cache implementation with same approach. Now if we use default configuration as 128 and some user tests with small string then it will fallback to least-request creating confusion around prefix-cache is not working as expected. 

I am ok to use 128 block size for default and release, with updated documentation to test with message greater than 128 characters. Let me know your final thoughts which option is preferred.

pkg/plugins/gateway/algorithms/least_request.go
Jeffwan: did we compare the diff between gateway cache vs source of truth? What if we moved to multiple replicas?

pkg/plugins/gateway/algorithms/prefix_cache.go
Jeffwan: We do not need to block this PR but this is something we have to consider in future. If we do such way, that will introduce the dependency on gpu-optimizer which makes the solution a little bit heavy.  /cc @zhangjyr

pkg/plugins/gateway/algorithms/prefix_cache.go
Jeffwan: I mean define variable to like  `var imbalanced bool` and assign values which make it easy to understand. i was confused by true or false earlier and finally notice the bool indicate the imbalance or not. This is a minor ask

pkg/plugins/gateway/algorithms/least_request.go
varungup90: 1) I validated the correctness for tracking running requests within gateway cache vs source of truth (inference engine).

2) For stateful use cases such as tracking running requests or prefix hash blocks, only one gateway leader instances should be used for accuracy. For running multiple gateway instances in all active pattern, requires backend storage implementation to sync state across gateway instances: https://github.com/vllm-project/aibrix/issues/761

pkg/plugins/gateway/algorithms/prefix_cache.go
Jeffwan: moving to utils sounds good. These are common codes and we do not need to preemptively move the folder.

pkg/plugins/gateway/algorithms/prefix_cache.go
Jeffwan: sounds good to me. Let's move them to utils

pkg/plugins/gateway/algorithms/least_request.go
Jeffwan: I recommend to leave a note here.  This needs to be refactored at the time we moved to HA case

pkg/plugins/gateway/algorithms/tokenizer/characters.go
Jeffwan: 1. Current byte way makes the splitting kind of meaningless, I kind of feel this brings some problem later in debugging. 

2. directly convert from []rune to []byte introduces overhead, I will check whether there's better ways to do it through string by leverage index in rune[] later. We can leave this to future PRs.

varungup90: Added a note for 1) and todo for 2).

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: Got it, added a variable.

config/gateway/gateway-plugin/gateway-plugin.yaml
Jeffwan: For this type of configuration, let's ensure consistency. The only differences between the release and default settings should be related to resource configurations, reflecting their respective resource constraints.

pkg/cache/cache_metrics.go
Jeffwan: Sounds good.

pkg/plugins/gateway/algorithms/least_request.go
varungup90: Done.

Jeffwan (approved):
/lgtm excellent!

pkg/utils/lrustore/lru_store.go
```diff
@@ -88,12 +89,10 @@ func (e *LRUStore[K, V]) Put(key K, value V) bool {
 }
 
 func (e *LRUStore[K, V]) Get(key K) (V, bool) {
```
vie-serendipity, line 91: I don't quite understand why the relevant entry isn't updated here. Shouldn't corresponding entry be moved to the front of the lruList when there is a matching request?

varungup90: Two reasons to remove the update here.

1) Right now LRU store is only used by hash table implementation. In MatchPrefix, it call the get method to retrieve the blocks but all these blocks will be re-iterated in AddPrefix which call put method for LRUStore where they will be updated with current time and moved to the head. 
2) After removing update from get method, its lock can be changed to RLock and improve the performance a little. Once the LRUStore implements sync map and does not require locking, we can add back update + move to head back in Get method.

justadogistaken:
I would like to inquire about the two algorithms: prefix-cache and prefix-cache-preble. Which one performs better, and how significant is the difference?

Issue #903, opened 2025-03-25
Move the benchmark codes to aibrix python package


Jeffwan:
### 🚀 Feature Description and Motivation

We can consider to make code reusable and move to https://github.com/vllm-project/aibrix/tree/main/python/aibrix

![Image](https://github.com/user-attachments/assets/06999554-6f59-48da-a43d-5ff1af9fbc71)

This make the generator/benchmark code reusable.

### Use Case

As a user, I want a single benchmark scripts to help me cover the inference engine testing

happyandslow:
Create end-to-end process to automate workload generation to workload running #947

Jeffwan:
@happyandslow the work has been done is a little bit different from what we proposed here? make it as a package rather than scripts? We can talk about the necessity

happyandslow:
Tracking this in #1066

PR #947, opened 2025-04-07
End-to-end script for workload runnning process

happyandslow:
## Pull Request Description
This process build an end-to-end script to automate data generation/workload generation/running generated workload.

## Related Issues
Resolves: #903


benchmarks/README.md
```diff
@@ -0,0 +1,109 @@
...
+
+## Run analysis
+
+Run analysis on benchmark result using: 
+```
+./benchmark analysis
+```
+Configure path and performance target via [config/base.sh](config/base.sh).
+
+
```
duli2012, line 113: Can the e2e benchmark output be displayed with Prometheus dashboard?

happyandslow: Not at the moment but we can certainly add this in the future.

Jeffwan:
@happyandslow this is not approved yet..

PR #954, opened 2025-04-08
fix: complete RayClusterFleet example for multi-node vLLM inference

ModiIntel:
Summary
This PR updates the existing RayClusterFleet example for qwen-coder-7b-instruct to make it fully functional for multi-node inference with vLLM and Ray.

What's fixed or added:
Corrected the command and args for the Ray head pod to properly start vllm serve

Explicitly set tensor-parallel-size and distributed-executor-backend

Added the aibrix-runtime sidecar container

Added a missing Service to expose port 8000 (vLLM)

Added an HTTPRoute to route OpenAI-compatible requests via Gateway API

Added the environment variable PYTORCH_CUDA_ALLOC_CONF=expandable_segments:True to handle memory fragmentation

Ensured Ray workers are correctly bootstrapped and terminated

Why this matters
The original example didn't include necessary components like the service or routing layer and lacked Ray-aware vLLM initialization. This update makes it a solid reference for users aiming to use distributed inference with multiple GPU nodes.

Tested and validated on an EKS cluster with 3x g4dn.2xlarge nodes.

### DIFF
--- a/samples/distributed/fleet-two-node.yaml
+++ b/samples/distributed/fleet-two-node.yaml
@@ -1,20 +1,20 @@
 apiVersion: orchestration.aibrix.ai/v1alpha1
 kind: RayClusterFleet
 metadata:
+  name: qwen-coder-7b-instruct
   labels:
     app.kubernetes.io/name: aibrix
     app.kubernetes.io/managed-by: kustomize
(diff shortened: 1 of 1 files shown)


Jeffwan (approved):
thanks! this is super helpful. 

/lgtm

PR #987, opened 2025-04-21
[BUG] ratelimit: fix the wrong TPM key name

runzhen:
## Pull Request Description

Should use `user.Name` as the key instead of the "user" struct itself,  keep consistent with the function `checkTPM()` when lookup the Redis key.


varungup90 (approved):

runzhen:
Hi @varungup90 @Jeffwan ,  do you know why the installation tests failed ?  I saw it failed with 2 different type of error logs. Can I run this test in my local host ?

kerthcet:
I personally think this deserves a test.

runzhen:
Hi @Jeffwan @kerthcet @varungup90  

After merge the latest code from main to my branch, the tests are passed.
  
I also increase another VTC test case's TPM value, looks like the test inference speed is fast than 1000. 
From an other perspective, it proves the TPM rate limit works now.  :-)

Jeffwan:
yeah, the token usage part coverage needs to be improved for sure. Let's add more test cases in separate issues.

PR #999, opened 2025-04-23
[Docs] Adding benchmark doc

happyandslow:
## Pull Request Description
This PR add documentation to benchmark tools

PR #1033, opened 2025-04-30
Support multiple configs in synthetic shared dataset

happyandslow:
## Pull Request Description
This PR supports dataset generation with different synthetic cache sharing patterns


benchmarks/config/base.sh
```diff
@@ -3,22 +3,22 @@
...
+export MODEL_NAME="deepseek-ai/DeepSeek-R1-Distill-Llama-8B"
+export TOKENIZER="deepseek-ai/DeepSeek-R1-Distill-Llama-8B"
 
 # ---------------
 # STEP 1: DATASET GENERATION
 # -------
 # Dataset config
 export DATASET_DIR="./output/dataset/"
-export PROMPT_TYPE="synthetic_multiturn" #"synthetic_multiturn"  "synthetic_shared", "sharegpt", "client_trace"
+export PROMPT_TYPE="synthetic_shared" #"synthetic_multiturn"  "synthetic_shared", "sharegpt", "client_trace"
```
duli2012, line 14: Do we document the possible PROMPT_TYPE, WORKLOAD_TYPE values somewhere?

duli2012 (approved):

benchmarks/config/base.sh
happyandslow: Yes -- these should all be documented in #999 . I'm modifying this for the ease of the first use case of our benchmark. But ideally I think this should be either "synthetic_multiturn" or "synthetic_shared" since the other to types requires additional dependencies.

PR #1038, opened 2025-05-01
Add Pareto Sampler for Multiturn Dataset Generation

happyandslow:
## Pull Request Description
This PR 
- Add a Pareto Sampler to multiturn dataset generation
- Fix timing bug in client


benchmarks/config/dataset/synthetic_multiturn.sh
```diff
@@ -1,6 +1,6 @@
-export PROMPT_LENGTH=100
-export PROMPT_STD=10
-export NUM_TURNS=10
-export NUM_TURNS_STD=1
-export NUM_SESSIONS=10
+export PROMPT_LENGTH=80
+export PROMPT_STD=472
```
xieus, line 2: How did we come out with this STD value?

happyandslow: This is the default value sampled from shareGPT dataset. We are just using this as a default setting but this could be dynamically configured during test time.

duli2012 (approved):

benchmarks/config/dataset/synthetic_multiturn.sh
xieus: Discussed with Le offline. This is addressed.

Issue #1040, opened 2025-05-02
dataset generation using generator


gangmuk:
### 🐛 Describe the bug

```bash
python utility.py convert \
    --path ${TRACE} \
    --type trace \
    --tokenizer deepseek-ai/deepseek-llm-7b-chat
usage: utility convert [-h] --path PATH --output OUTPUT --type {trace,sharegpt}
utility convert: error: the following arguments are required: --output
```

I followed the readme in dataset-generator and ran into issue when executing`utility.py`. tokenizer argument does not exist and output argument is not specified.

### Steps to Reproduce

```bash
python multiturn_prefix_sharing_dataset.py \
    --prompt-length-mean 100 \
    --prompt-length-std 10 \
    --num-turns-mean 10 \
    --num-turns-std 1 \
    --num-sessions-mean 10 \
    --num-sessions-std 1
```

```bash
export TRACE="output.jsonl"
python utility.py convert \
    --path ${TRACE} \
    --type trace \
    --tokenizer deepseek-ai/deepseek-llm-7b-chat
```

output
```bash
python utility.py convert \
    --path ${TRACE} \
    --type trace \
    --tokenizer deepseek-ai/deepseek-llm-7b-chat
usage: utility convert [-h] --path PATH --output OUTPUT --type {trace,sharegpt}
utility convert: error: the following arguments are required: --output
```

### Expected behavior

not entirely sure. maybe changing output.jsonl format to another format that can be used in workload-generator?

### Environment

main branch

happyandslow:
The utility.py convert (with --type trace) is intended for converting client log (generated by client.py). Once you generate output.json. from multiturn_prefix_sharing_dataset.py. Your dataset is ready to be used by workload generator.

gangmuk:
but isn't this still unintended?
```bash
python utility.py convert \
    --path ${TRACE} \
    --type trace \
    --tokenizer deepseek-ai/deepseek-llm-7b-chat
usage: utility convert [-h] --path PATH --output OUTPUT --type {trace,sharegpt}
utility convert: error: the following arguments are required: --output
```

utility.py does not have tokenizer as argument?


another question about output of generated `output.jsonl`. this is the example output.jsonl. it does not have timestamp for example. if this is input to the `client.py`, I believe it needs timestamp field, doesn't it?

```
[...]

Issue #1029, opened 2025-04-30
Tracking multiple benchmark code issues


Jeffwan:
### 🐛 Describe the bug

Seems there're many code issues in current benchmark. I will not cut issues one by one but create an umbrella issue to track them.

### Steps to Reproduce

N/A

### Expected behavior

It should work out of box.

### Environment

- AIBrix version: main branch

Jeffwan:
## 1. Doc error. should it `./benchmark.sh`

![Image](https://github.com/user-attachments/assets/d91c64c8-c72a-4587-8e86-6cf5d038b70b)
![Image](https://github.com/user-attachments/assets/17f4b19c-7b42-4dcb-8c4c-3c7ae3a12091)

Jeffwan:
### 2.  what's the meaning of benchmark.sh? 

![Image](https://github.com/user-attachments/assets/cddfd277-fd84-4b36-aae2-661e49db624b)

it hides all the parameters, why not just use python directly.

Jeffwan:
### 3. how to correctly use the argument? 

Should the config override the default ones?

![Image](https://github.com/user-attachments/assets/7de7d0a2-4586-49d4-a2b5-aa00057f338e)

![Image](https://github.com/user-attachments/assets/896682ad-036f-438c-b848-cf54dc3c00c8)

Jeffwan:
### 4.  data format issue

Seems there's no user concept for multi-turn conversation? prompts do not have roles? should we add roles and use conversation format?

![Image](https://github.com/user-attachments/assets/fabb0eae-86bc-4b0c-bba5-a511eb870483)

Jeffwan:
### 5. workload file format not explained in the doc. Could you give some sample for users?

something like below
```
{"timestamp": 19, "requests": [{"prompt": "I need to understand data science for my startup idea. Can you help? Could you also explain how this relates to natural language processing? For context, I have experience with cybersecurity but I'm new to this specific area. I've been trying to understand this concept for months and would appreciate a clear explanation. I'm asking because I need to deploy a machine learning model for a project. For context, I have experience with cryptocurrency but I'm new to this specific area. Could you", "prompt_length": 101, "output_length": null, "session_id": 0}]}
```

Jeffwan:
### 6. benchmark

too many configuration files in the benchmark. this can be passed in a json string directly in the configuration. No need to create many files. 

```
{
    "fluctuate": 696, 
    "mean": 1024, 
[...]

PR #1042, opened 2025-05-02
Dataset generator output argument fix

happyandslow:
## Pull Request Description
This PR updates dataset generator's missing argument.

## Related Issues
Resolves: #1040

### DIFF
--- a/benchmarks/generator/dataset-generator/README.md
+++ b/benchmarks/generator/dataset-generator/README.md
@@ -107,7 +107,6 @@ export TRACE="trace.jsonl"
 python utility.py convert \
     --path ${TRACE} \
     --type trace \
-    --tokenizer deepseek-ai/deepseek-llm-7b-chat \
     --output output.jsonl
 ```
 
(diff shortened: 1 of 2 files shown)


benchmarks/generator/dataset-generator/README.md
```diff
@@ -75,6 +127,7 @@ wget https://huggingface.co/datasets/anon8231489123/ShareGPT_Vicuna_unfiltered/r
 python utility.py convert  \
     --path ${TARGET_DATASET} \
     --type sharegpt \
-    --tokenizer deepseek-ai/deepseek-llm-7b-chat
+    --tokenizer deepseek-ai/deepseek-llm-7b-chat \
```
gangmuk: is this arg supposed to be in `utility.py`? I don't see it used for now? and it does not exist in the argparse argument as well?

happyandslow: Good catch. Pushed a fix.

Jeffwan:
@gangmuk @happyandslow the PR looks good to you? no further changes, right?

gangmuk:
> @gangmuk @happyandslow the PR looks good to you? no further changes, right?
yeah at least for this bug. it can be merged. if another issues are found, they can be done in separate PR in the future. 

After taking a look at dataset-generator and workload-generator, I generally think they can be simplified. there are multiple different formats of output in each step and not sure they are all needed, and if they are, where they are needed (there are multiple conversions user can/should do). I didn't take further look after encountering the above issue, so I don't have specific opinion about how to improve. leaving comments here for future reference

Jeffwan:
@gangmuk I left most of my feedback here. https://github.com/vllm-project/aibrix/issues/1029 you can add it there as well. I am waiting for Le to triage and address some of them

Jeffwan (approved):

PR #1043, opened 2025-05-02
Adding interval scaling factor for client

happyandslow:
## Pull Request Description
This PR adds a --time-scale factor to benchmark client for the ease of rate adjustment from the client side during the benchmark process.


DwyaneShi (approved):
LGTM.

PR #1049, opened 2025-05-05
[Fix] Fix multiple issues for benchmark implementation

happyandslow:
## Pull Request Description
This PR fixes multiple existing issues with benchmark implementation

## Related Issues
Resolves: #1029


benchmarks/benchmark.sh
```diff
@@ -178,6 +181,7 @@ run_client() {
         --model \"$TARGET_MODEL\" \
         --api-key \"$API_KEY\" \
         --time-scale \"$TIME_SCALE\" \
+        --routing-strategy \"$ROUTING_STRATEGY\" \
```
Jeffwan, line 184: if we plan to use benchmark against other framework, this is not needed.

Jeffwan:
Overall looks good to me. @happyandslow can you or du help on this issue so we can test these code in each CI run.  https://github.com/vllm-project/aibrix/issues/1050

Jeffwan (approved):

Issue #1095, opened 2025-05-16
ModelAdapters do not dynamically route to new pods


TheCodeWrangler:
### 🐛 Describe the bug

When I scale up my deployment using a lora adapter I see that all the traffic to the lora adapter always goes to the pod that initially was there when the ModelAdapter was created.

Requests to the base model are still load balanced across the new backends

Additionally after scaling down if the initial pod is removed i see errors from the system when requesting the lora module BUT `kubectl describe modeladapter lora-name` shows that it is still Running (though likely pointing to a dead resource)

### Steps to Reproduce

1. Deploy a lora adapter with HPA on deployment
2. Load test to trigger scale up
3. Observe the number of running requests per pod are all on one pod

### Expected behavior

Load should be balanced across all running pods

### Environment

2.1.0

TheCodeWrangler:
Thinking about this more I am using a keda autoscaler on my backends.... Do I need to use AIbrix Auto scalers?

dittops:
I'm facing the same issue with v3.0.0. The controller is not triggering the adapter load for the new pods.

PR #1132, opened 2025-05-23
[Misc] Support adapter scaling to all replicas

dittops:
## Pull Request Description

Support adapter scaling to all replicas. 
- allow controller to sync adapter instances with all active pods
- load adapter on each pod
- update EndpointSlice with all pod IPs
- adjust resources and tests for multi-pod support

## Related Issues
Resolves: #1095

### DIFF
--- a/pkg/controller/modeladapter/modeladapter_controller.go
+++ b/pkg/controller/modeladapter/modeladapter_controller.go
@@ -37,7 +37,6 @@ import (
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/meta"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/types"
 	corelisters "k8s.io/client-go/listers/core/v1"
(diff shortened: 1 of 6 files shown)


Jeffwan:
@dittops Great! I will spend some time this week to review this change

pkg/controller/modeladapter/modeladapter_controller.go
```diff
@@ -558,36 +526,33 @@ func (r *ModelAdapterReconciler) reconcileLoading(ctx context.Context, instance
...
+			}
+			return err
+		}
 
-	// selectPod could be in termination, in this case, we just do nothing.
-	if targetPod.DeletionTimestamp != nil {
-		return nil
-	}
+		if targetPod.DeletionTimestamp != nil {
+			continue
```
Jeffwan, line 626: TODO: add a log message here for any pods skipped

pkg/controller/modeladapter/modeladapter_controller.go
```diff
@@ -791,166 +753,48 @@ func (r *ModelAdapterReconciler) reconcileService(ctx context.Context, instance
...
 	found := &discoveryv1.EndpointSlice{}
-	// instance could be clean up in earlier reconciliation, we need to clean up the instance from endpoint list.
-	if len(instance.Status.Instances) == 0 {
-		klog.Warningf("model adapter %s has not been deployed to any pods yet or being deleted, skip creating endpointslice", klog.KObj(instance))
 
-		// reset endpoint slice
-		if err := r.Get(ctx, types.NamespacedName{Namespace: instance.Namespace, Name: instance.Name}, found); err != nil {
-			if apierrors.IsNotFound(err) {
-				klog.Warningf("Failed to fetch the endpoint slice %s", klog.KObj(instance))
+	podList := []corev1.Pod{}
```
Jeffwan, line 894: seems the logic has been simplified a lot. I will double check the logics

pkg/controller/modeladapter/modeladapter_controller.go
```diff
@@ -343,72 +342,41 @@ func (r *ModelAdapterReconciler) DoReconcile(ctx context.Context, req ctrl.Reque
...
-		if err != nil {
-			return ctrl.Result{}, err
+	added := false
+	for name := range activeMap {
+		if !StringInSlice(instance.Status.Instances, name) {
+			instance.Status.Instances = append(instance.Status.Instances, name)
+			added = true
 		}
-		if len(activePods) != 0 {
-			selectedPod, err = r.schedulePod(ctx, instance, activePods)
```
Jeffwan, line 400: em. @dittops can you confirm the scheduling logic has been disabled in the new workflow? any reasons?

Jeffwan:
@dittops I think the only part I was not that sure is the scheduling part. can you give more details?

dittops:
Have used the  following logic for Adapter loading/unloading

1. The pods are selected with Label-Based Matching - adapter.model.aibrix.ai/enabled: "true"
2. The selected pods are added to the Status.Instances list
3. Use the reconcileLoading function to iteratively load the adapters on each of the pods from the Instances list.

Jeffwan:
@dittops the workflow sounds good. from the change change, I notice the lora scheduling logic has been deleted.  In this case, how to select pods? 
image

dittops:
scheulePod was used to pick one pod and then assigned to instance.Status.Instances. Instead of choosing one pod, the new approach uses ALL pods that match the selector and adds all matching pods to instance.Status.Instances

If we need to keep the schedulePod, we can move the getActivePodsForModelAdapter inside schedulePod to select all pods and then return a list

Jeffwan:
@dittops  Yeah, I think the behavior has changed a bit recently.

Option 1: Schedule the LoRA model to specific pods based on the specified replicas.
Option 2: Load the LoRA into all base model replicas so that all models are identical — this is the approach you're switching to.

While Option 2 is a valid pattern that we can support, I strongly recommend sticking with Option 1 (with multi-replica support) as the primary solution for now. In our case, some high-rank LoRA models are quite large, and it's not practical to scale using Option 2. We could consider adding Option 2 as a separate feature later.

What do you think?

dittops:
@Jeffwan Are you referring to adding a replica count in the ModelAdapter and use that for scheduling?
eg:
```
  spec:
    replicas: 3  # Only load on 3 pods
    podSelector:
      matchLabels:
        model.aibrix.ai/name: base-model
```

Jeffwan:
@dittops exactly. https://github.com/vllm-project/aibrix/blame/main/api/model/v1alpha1/modeladapter_types.go#L53

Jeffwan:
@dittops apologies for late response. I am recently refactoring lora work to provide better production level support. I want to merge this one first before I refactor the codes. However Seems I can not rebase main branch changes, could you help rebase the branch?

image

dittops:
@Jeffwan, I have rebased. Could you take a look?

Jeffwan:
image

@dittops I think the problem is vLLM community update the CI after this PR is out, result in some failures on the CI checking. I try to close this PR and reopen it, but seems it doesn't help that much. Could you help cut a new PR, we can directly work on new one and discard this one.

Jeffwan:
@dittops now it's working. thanks! no need to cut new PRs

Jeffwan:
the change overall looks good to me. I will address gemini's feedback in later refactor

PR #1180, opened 2025-06-05
fix: gateway benchmark info

Xunzhuo:
fix: gateway benchmark info


Jeffwan:
very straightforward change. looks good to me

Jeffwan (approved):

Issue #642, opened 2025-02-10
[RFC]: Load-aware pattern-based routing policy with profile support


zhangjyr:
### Summary

Having access to the GPU profile used by the GPU optimizer, we propose to add a new routing policy that utilizes performance profiles per input/output token pattern to better work with the GPU optimizer for fewer SLO violations.

### Motivation

Our experiment results show that both heterogeneous and homogeneous GPU clusters need a pattern-based routing policy to minimize SLO violations.

### Proposed Change

With the GPU profile, we have:
1. GPU's max arrival rate (RPS) abides by the defined SLO for different models, # of input tokens, and # of output tokens. We define a pattern as P = (GPU, Model, Input tokens, Output tokens)
2. (Optional) Associated performance metrics: E2E Latency (E2E), TTFT,  TPOT, and the SLO used by the profile.
Using Little's law, we can get the average pending requests that abide by SLO: 

Pending requests(P) = RPS(P) * E2E(P)

For example, if an LLM can server (input 256, output 16) request with 4 RPS and E2E latency 0.5s, the pending requests would expect to be 4 * 0.5 = 2. More examples:
- 256/16: 4RPS, 0.5s = 2 Pending requests
- 4096/32: 1RPS, 5s = 5 Pending requests
Note that the profile indicates that adding more requests will break the stable status, cause pending requests to accumulate over time, and break SLO. 

We define the capacity of a pattern (P) using the number of pending requests:

Capacity(P) = Pending requests(P)

So, we can normalize the load per request corresponding to the workload and GPU per (# input tokens, # output tokens) pairs as:

Load(P) = 1 / Capacity(P) = 1 / (RPS(P) * E2E(P)

The proposed routing policy has two components:
1. Load-aware routing policy
2. Pattern-based queue policy

For routing policy, we propose to change from the current push-based routing strategy to a pull-based routing strategy. Push-based routing strategy simply pushes requests to the vLLM serving instance and request queue at the instance side regardless of instance capacity. The pull-based routing strategy maintains a global queue and only pulls from the global queue if the instance side has spare capacity. With GPU profiles-based capacity definition, we can implement a pull-based routing strategy in the gateway without changing the vLLM instance. We find a least-load routing strategy is sufficient.

[...]

Issue #606, opened 2025-01-24
Consider to support delay scheduling in Gateway


Jeffwan:
### 🚀 Feature Description and Motivation

Delay scheduling request to avoid over-assignment to some inference engines. We actually have the discussion on the push or pull based solutions. This would be a compliment of the push solution.

varungup90:
@zhangjyr reassigned to you, since you have already started working on it.

Jeffwan:
@zhangjyr I remember our review feedback earlier is to have decoupled PR from the cache refactoring.  cache refacor PR was merged, delay scheduling PR was not out right? I think it's not urgent for v0.3.0 release. If that one need more time on testing and performance evaluation, I think leave it to v0.4.0 makes more sense. Feel free to let us know your thoughts.

zhangjyr:
In fact. My SLO-based routing policy introduced a new abstraction called QueueRouter, which allows reordering requests based on the queue type. The QueueRouter is a framework-level router that bridges between the ordering of requests (queue) and the routing of requests. So the QueueRouter can be regarded as a solution for this ticket. It provides the capability to implement a delayed router, and the SLO router is an example of a delayed router.

Jeffwan:
@zhangjyr few questions, 

- SLO-based routing policy is associated with this issue https://github.com/vllm-project/aibrix/issues/642?
- If other routing policies need delay scheduling as a common capability to leverage, is it possible? Or it only works with SLO-based routing policy? I am trying to understand whether we need to leave this story. Seems the scope is a little bit different.

zhangjyr:
> If other routing policies need delay scheduling as a common capability to leverage, is it possible? Or it only works with SLO-based routing policy? I am trying to understand whether we need to leave this story. Seems the scope is a little bit different.

Yes, it is possible. The QueueRouter accepts two parameters, a queue and a router. The idea to implement a delayed scheduler cooperates with other routing policies is like:
1. Create a queue to implement delay logic.
2. The queue may accept other routing policies as a backend and behave as a router proxy.
3. In the Peer and Route(router proxy) method, the queue can reorder requests based on the customized optimization target.
Given a specific optimization target, we can work on the detailed implementation case by case.

Jeffwan:
Sounds good. Then this story probably can focus more on how to enable delay scheduling for other routing solution or something bake into default logic that all can leverage . As we discussed that #642 will be moved to v0.4.0 release, I think we need to postpone to v0.4.0 as well.

PR #1192, opened 2025-06-12
[Misc] SLO-aware router with profile support

zhangjyr:
## Pull Request Description
Introducing SLO-aware router with profile support. This PR introduces three new SLO-aware routing policies:
1. slo (or slo-least-load-pulling)
2. slo-least-load
3. slo-pack-load
All three routing policies will prioritize requests with a profiled SLO target.

In addition to the slo-family routing policies, this PR added built-in queues to support request reordering and future delay scheduling. In particular, QueueRouter enables the pull mode within the gateway. Below is a comparison of pulling mode and default push mode:
1. Push mode: The router dispatches requests to the server, possibly overloading the server.
2. Pull mode: The server pulls requests from the router based on the server's capacity.
With profile support, the gateway now has server capacity knowledge and can achieve pull mode within the gateway.

Additional feature added in this PR:
1. Add a fallback routing policy mechanism for developers to designate a default routing policy if the specified routing policy fails.
2. Wrap the routing policy registration mechanism in RouterManager, allowing it to be reused for managing a family of related routing policies. (Including simplify Select() to follow RouterProviderFunc)
3. Add profile cache API to manage model-based performance profiles.
4. Improving the profile generator in the GPU manager to include SLO information and detailed metrics.

Preliminary results show SLO policy can achieve the SLO target for composite workload on heterogeneous GPUs:
image
Workload: mixed sharegpt and bird workload with a ratio of 7:4
GPU: 1A10, 4L20
SLO: Latency per token 0.05s

## Related Issues
Resolves: #642 #606

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>


Jeffwan:
I notice there're some refactor changes (e.g. internal interface change etc) Technically, that affects other aspects, could it be some separate changes? I mean splitting the changes into common parts (stakeholder needs to review it) and slo specific changes (review could be loose and feature can be protected by feature gate). 

If the splitting is too complicated, we can have 1st round review and check how to move forward

zhangjyr:
@Jeffwan, I think the only internal interface change is the Select(). The function is called only in one place, and if you find it is not appropriate, we can restore it.

cmd/plugins/main.go
```diff
@@ -77,7 +78,7 @@ func main() {
 		panic(err)
 	}
 
-	cache.InitForGateway(config, stopCh, redisClient)
+	cache.InitForGateway(config, stopCh, redisClient, routing.NewSLORouter)
```
Jeffwan: is there better way to handle this case?  `routing.NewSLORouter` is a specific solution but `cmd/plugins/main.go` is for common purpose? Can we have factory for such initialization?

config/gateway/gateway-plugin/gateway-plugin.yaml
```diff
@@ -2,7 +2,7 @@ apiVersion: v1
 kind: Service
 metadata:
   name: gateway-plugins
-  namespace: aibrix-system
+  namespace: system
```
Jeffwan, line 5: system will be overitten eventually to `aibrix-system` here right? changing to `system` to be aligned with default setting?

pkg/cache/cache_api.go
```diff
@@ -132,3 +142,18 @@ type RequestTracker interface {
...
+	// Parameters:
+	//   pod: Pod object
+	//   modelName: Name of the model
+	GetModelProfileByPod(pod *v1.Pod, modelName string) (*ModelGPUProfile, error)
+
+	// GetModelProfileByDeploymentName gets model profile for a deployment
+	// Parameters:
+	//   deploymentName: Name of the deployment
+	//   modelName: Name of the model
+	GetModelProfileByDeploymentName(deploymentName string, modelName string) (*ModelGPUProfile, error)
```
Jeffwan, line 158: TODO: we may use other objects to orchestrate pods in future. in that case, deployment might be changed in future. This looks good at this moment. 

one more problem is, deployment without namespace can not be used to identify a deployment. we need to append the namespace field

pkg/cache/cache_impl.go
```diff
@@ -176,27 +177,41 @@ func (c *Store) GetMetricValueByPodModel(podName, podNamespace, modelName string
...
-	if enableGPUOptimizerTracing {
-		success := false
-		for {
-			trace := c.getRequestTrace(modelName)
-			// TODO: use non-empty key if we have output prediction to decide buckets early.
-			if traceTerm, success = trace.AddRequest(requestID, ""); success {
-				break
+	// Current implementation assumes AddRequestCount() will not be called concurrently.
+	// TODO: Implment "wait for trace term" logic if AddRequestCount() is called concurrently.
+	if ctx == nil || ctx.CanAddTrace() {
```
Jeffwan, line 182: is the refactor for common case?

pkg/cache/cache_init.go
```diff
@@ -37,25 +38,37 @@ var (
...
 
 	// Pod related storage
 	metaPods utils.SyncMap[string, *Pod] // pod_namespace/pod_name -> *Pod
 
-	// Mapping relationships
+	// Model related storage
 	metaModels utils.SyncMap[string, *Model] // model_name -> *Model
 
+	// Deploymnent related storage
+	deploymentProfiles utils.SyncMap[string, *ModelGPUProfile] // deployment_name -> *ModelGPUProfile
```
Jeffwan: same here. we can use `namespace/deployment` as the key

pkg/cache/informers.go
```diff
@@ -98,6 +97,7 @@ func (c *Store) addPod(obj interface{}) {
 	// only track pods with model deployments
 	modelName, ok := pod.Labels[modelIdentifier]
 	if !ok {
+		// klog.InfoS("ignored pod without model label", "name", pod.Name)
```
Jeffwan: use log level instead?

pkg/cache/load_provider.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+var (
+	ErrorNotSupport          = fmt.Errorf("not support")
+	ErrorSLOFailureRequest   = fmt.Errorf("slo failure request")
+	ErrorLoadCapacityReached = fmt.Errorf("load capacity reached")
+)
+
+// LoadProvider provides an abstraction to get the utilizatin in terms of specified metrics
+type LoadProvider interface {
```
Jeffwan, line 32: what implementations expected here? I am trying to understand in which scenarios we can extend this part?

pkg/cache/model.go
```diff
@@ -16,16 +16,22 @@ limitations under the License.
...
+// ModelRouterProviderFunc defines the function to provider per-model router
+type ModelRouterProviderFunc func(modelName string) (types.Router, error)
+
 type Model struct {
 	// Pods is a CustomizedRegistry that stores *v1.Pod objects.
 	// The internal map uses `namespace/name` as the key and `*v1.Pod` as the value.
 	// This allows efficient lookups and caching of Pod objects by their unique identifier.
 	Pods *utils.CustomizedRegistry[*v1.Pod, *utils.PodArray]
 	// Metrics utils.SyncMap[string, metrics.MetricValue] // reserved
+	OutputPredictor types.OutputPredictor
```
Jeffwan, line 36: better to add a comment if these two fields are only for slo router

pkg/cache/output_predictor.go
```diff
@@ -0,0 +1,265 @@
...
+	// InputColdPrediction predicts the output to be the same as the input.
+	InputColdPrediction
+	// PessimiticColdPrediction predicts the output to be maximum maxOutputTokens.
+	PessimiticColdPrediction
+)
+
+// ColdPredictionStrategy defines the strategy when there is no history for the predictor.
+type ColdPredictionStrategy int
+
+type SimmpleOutputPredictor struct {
```
Jeffwan: typo. `SimmpleOutputPredictor` -> `SimpleOutputPredictor`

pkg/cache/output_predictor.go
```diff
@@ -0,0 +1,265 @@
...
+		bucket = limit - 1
+	}
+	return bucket
+}
+
+func (p *SimmpleOutputPredictor) tryRotate(ts time.Time) {
+	if ts.Sub(p.history.headTimestamp) < MovingInterval {
+		return
+	}
+	go p.rotate(ts)
```
Jeffwan, line 251: is goroutine necessary here?

pkg/cache/output_predictor.go
```diff
@@ -0,0 +1,265 @@
...
+}
+
+func (p *SimmpleOutputPredictor) bucket2idx(inputBucket, outputBucket int) int {
+	return inputBucket*p.inputBuckets + outputBucket
+}
+
+func (p *SimmpleOutputPredictor) token2bucket(tokens int, limit int) int {
+	bucket := 0
+	if tokens > 0 {
+		bucket = int(math.Round(math.Log2(float64(tokens))))
```
Jeffwan, line 236: round result is not stable, changing to `math.Floor` in the bucket scenarios looks better

pkg/cache/output_predictor.go
```diff
@@ -0,0 +1,265 @@
...
+	case PessimiticColdPrediction:
+		return MaxOutputLen
+	default:
+		// Default to optimistic
+		return 1
+	}
+}
+
+func (p *SimmpleOutputPredictor) bucket2idx(inputBucket, outputBucket int) int {
+	return inputBucket*p.inputBuckets + outputBucket
```
Jeffwan: is this used for mapping the 2D dimension -> single dimension? in that case, should it be `inputBucket*p.outputBuckets`

pkg/plugins/gateway/algorithms/least_busy_time.go
```diff
@@ -26,9 +26,7 @@ import (
 	klog "k8s.io/klog/v2"
 )
 
-var (
-	RouterLeastBusyTime types.RoutingAlgorithm = "least-busy-time"
-)
+const RouterLeastBusyTime types.RoutingAlgorithm = "least-busy-time"
```
Jeffwan, line 30: for such changes not directly related to this change, let's cut separate PRs in future

pkg/plugins/gateway/algorithms/least_load.go
```diff
@@ -0,0 +1,121 @@
...
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/klog/v2"
+)
+
+var (
+	ErrorNoAvailablePod = fmt.Errorf("no pod available")
+)
+
+type leastLoadRouter struct {
+	provider       cache.LoadProvider
```
Jeffwan, line 43: if user specific least load algorithm, but with profiling disabled, could that be an issue here?`provider` and `cappedProvider`

pkg/plugins/gateway/queue/simple_queue.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+	if initialCapacity >= types.DefaultQueueCapacity {
+		queue.zeroout = make([]V, initialCapacity)
+	}
+	return queue
+}
+
+func (q *SimpleQueue[V]) Enqueue(value V, _ time.Time) error {
+	var zero V
+	if value == zero {
```
Jeffwan, line 59: what's the potential type of value? seems this part is a little bit tricky

pkg/plugins/gateway/queue/simple_queue.go
```diff
@@ -0,0 +1,205 @@
...
+
+func (q *SimpleQueue[V]) Cap() int {
+	q.mu.RLock()
+	defer q.mu.RUnlock()
+
+	return cap(q.queue)
+}
+
+func (q *SimpleQueue[V]) physicalPosRLocked(pos int64) int64 {
+	return pos - q.baseCursor
```
Jeffwan, line 162: there's no possibility to get negative number, right?

pkg/plugins/gateway/queue/simple_queue.go
```diff
@@ -0,0 +1,205 @@
...
+			}
+		} else {
+			var nilVal V
+			for start < len(newQueue) {
+				newQueue[start] = nilVal
+				start++
+			}
+		}
+	}
+	q.queue, q.baseCursor = newQueue, q.baseCursor+dequeuePos
```
Jeffwan, line 203: could it be a problem is the other goroutine invoke `physicalPosRLocked`?  can we introduce something like and use it in `physicalPosRLocked` and `setBaseCursor ` in `expand`?

```
func (q *SimpleQueue[V]) getBaseCursor() int64 {
	return atomic.LoadInt64(&q.baseCursor)
}

```

pkg/cache/output_predictor.go
```diff
@@ -0,0 +1,265 @@
...
+	return forwarded
+}
+
+func (hist *rotatingHistory) resetTail(distributions outputDistribution, sums []int32, outputBuckets int) {
+	hist.Tail().reset(distributions, sums, outputBuckets)
+	hist.tail = (hist.tail + 1) % len(hist.window)
+	atomic.AddInt32(&hist.size, -hist.Tail().getSkipped())
+}
+
+func NewSimmpleOutputPredictor(maxInputTokens, maxOutputTokens int, window time.Duration) *SimmpleOutputPredictor {
```
Jeffwan: let's briefly talk about the algorithm here? as a comment

pkg/types/router_context.go
```diff
@@ -27,83 +28,242 @@ import (
...
+	PendingLoad float64   // Normalized pending load of request, available after AddRequestCount call.
+	TraceTerm   int64     // Trace term identifier, available after AddRequestCount call.
+	RoutedTime  time.Time
 
 	targetPodSet chan struct{}
 	targetPod    atomic.Pointer[v1.Pod]
+	lastError    atomic.Pointer[error]
 	debugDelay   time.Duration
+	tokens       []int
+	predictor    OutputPredictor
```
Jeffwan: one of my concerns is which field can be used for profile disabled routing algorithms? As a routing algorithm developer, which field should I expected to be available if I enable/disable some features.

pkg/plugins/gateway/queue/slo_queue.go
```diff
@@ -0,0 +1,513 @@
...
+		return 0.0, 0.0, err
+	}
+	target = profile.SLOs.TPOT*float64(tokens) + profile.SLOs.TTFT // TTFT is optional if the workload is decoding dominant.
+	// Since profile metrics only include mean values, the assumption here is mean(e2e) = mean(ttft) + mean(tpot) * tokens
+	// In the case that the workload is decoding dominant, mean(e2e) = mean(tpot) * tokens, while mean(ttft) is ignorable.
+	expected, err = profile.LatencySeconds(signature...)
+	return
+}
+
+func (q *SLOQueue) higherRank(rank1 float64, rank2 float64) float64 {
```
Jeffwan, line 466: directly return bool looks simplier

pkg/plugins/gateway/queue/slo_queue.go
```diff
@@ -0,0 +1,513 @@
...
+	"github.com/vllm-project/aibrix/pkg/utils"
+	"k8s.io/klog/v2"
+)
+
+const (
+	fifoOnNonSLOViolation    bool = false
+	queueOverallSLO          bool = false
+	monogenousGPURouting     bool = true
+	monogenousGPURoutingOnly bool = monogenousGPURouting && false
+	initialTotalSubQueues    int  = 8  // Expect no more than 8 subqueues
```
Jeffwan, line 40: are these magic numbers const or should be adjusted based on the available resources?

cmd/plugins/main.go
zhangjyr: Added ModelRouterFactory in the next review.

config/gateway/gateway-plugin/gateway-plugin.yaml
zhangjyr: Yes, all other configurations use system instead of aibrix-system. Just keep consistency here.

pkg/cache/cache_api.go
zhangjyr: In the case of deployment using other objects, the GPU optimizer would have been changed as well (it monitors deployment only). For the support of ray clusters, let me keep a note, leave this comment open, and add an issue after merging.

Can you explain the cases where "deployment without namespace can not be used to identify a deployment"?

pkg/cache/cache_impl.go
zhangjyr: This will work for the previous routers. Yes.

pkg/cache/load_provider.go
zhangjyr: LeastKVCache and LeastRequest can both be refactored to offer different LoadProviders and reuse the LeastLoad router.

pkg/cache/model.go
zhangjyr: These two fields can be generally used by others:
1. The Pods registry outputs PodArray, allowing the model-associated pods to be categorized by GPU/deployment.
2. OutputPredictor offers output token prediction and is directly used by the RoutingContext. 
They are both decoupled with SLO routing.

pkg/cache/output_predictor.go
zhangjyr: This is to avoid the window rotating, blocking the routing for the following reason: Window rotating involves an expansive global write lock/unlock. Without a go routine, the lock will be effectively put on every request on the request path.

pkg/cache/output_predictor.go
zhangjyr: Considering the SLO target, math.Ceil() might be better than math. Floor to overestimate the token bucket. However, on a log scale, the Round() method will effectively Ceil() more requests but still Floor() some bordering values. May I ask for examples of why the round will be a problem?

pkg/plugins/gateway/algorithms/least_load.go
zhangjyr: No, it will fall back to the least-request router. slo-test.go has tests for these situations.

pkg/plugins/gateway/queue/simple_queue.go
zhangjyr: Since V is an arbitrary type (pointer or struct), the empty value has to be defined like this. In most cases, V would be a pointer, and this statement is equivalent to value == nil.

pkg/plugins/gateway/queue/simple_queue.go
zhangjyr: Yes, pos and baseCursor will both increase only, where baseCursor will only increase in expand() (for packing and expansion) after all positions smaller are dequeued (dequeuePos).

pkg/plugins/gateway/queue/simple_queue.go
zhangjyr: BaseCursor is updated only in lock(), and physicalPosRLocked() is Rlocked(), which will not be executable during lock() except being specifically called. The -RLocked() naming postfix asks developers to call this function specifically in Rlock() contexts.

pkg/cache/output_predictor.go
zhangjyr: The predictor simply collects the seen output token histogram for each input bucket and uses weighted random for prediction. Comments will be added

pkg/types/router_context.go
zhangjyr: Current SLO family routers are profile-based, if disabled, we will fallback to least-request. The profile is essential in the current implementation because the profile provides the model-based SLO target and server capacity in terms of load. For a non-model-specific general SLO (e.g., 120s latency), if the SLO can be set using environment variables, I think the SLO queue can also be combined with other stateless routers to provide some extent of improvement. OutputPredictor itself is profile-decoupled and can work independently.

pkg/plugins/gateway/queue/slo_queue.go
zhangjyr: Yes, it will be clearer. But since the higherRank will be used in either asc or desc contexts, I would prefer the output being equal-aware to minimize swap operations during sort.

pkg/plugins/gateway/queue/slo_queue.go
zhangjyr: They are constants and are there only for evaluation purposes. These are features switches inherited from my router simulator. I started from the router simulator for the best configuration, and sync switches for convenience later.

pkg/cache/cache_init.go
zhangjyr: The key is in fact in the format aibrix:profile_[model_name]_[deployment_name], the name is unique across namespaces given:
1. model name are unique across namespaces.
2. deployment_names have the same namespace as the model name.
I've updated the comment here for clarification.

pkg/plugins/gateway/queue/slo_queue.go
Jeffwan: sound good

pkg/cache/model.go
Jeffwan: em. in that case, I think we should have more documentation or clarification. because we do not have other cases leveraging such feature yet.

pkg/cache/cache_api.go
Jeffwan: I mean `namespace/deployment_name` as the key,

config/gateway/gateway-plugin/gateway-plugin.yaml
Jeffwan: sounds good here

pkg/cache/output_predictor.go
Jeffwan: it's not a big issue but introduce some vague bordering issue if the number is close to `.5`, not that stable as `Floor` in token bucketing scenarios.

Jeffwan:
@zhangjyr please rebase the branch one more time. I tried to merge it but notice some rebase conflicts

pkg/cache/cache_api.go
zhangjyr: The key is in fact in the format aibrix:profile_[model_name]_[deployment_name], the name is unique across namespaces given:
1. model name are unique across namespaces.
2. deployment_names have the same namespace as the model name.

zhangjyr:
> @zhangjyr please rebase the branch one more time. I tried to merge it but notice some rebase conflicts

@Jeffwan Done

pkg/cache/model.go
zhangjyr: Done

pkg/utils/pod.go
```diff
@@ -36,6 +37,17 @@ const (
 	defaultPodMetricPort = 8000
 )
 
+var (
+	ReplicaSetDeploymentFinder = regexp.MustCompile(`^(.*)-\w+$`)     // Deployment-[random name]
+	RayClusterFleatFinder      = regexp.MustCompile(`^(.*)-\w+-\w+$`) // RayClusterFleat-[random name]-[random name]
```
varungup90: spelling fleat -> fleet.

pkg/types/router_context.go
```diff
@@ -27,83 +28,244 @@ import (
...
+	RequestTime time.Time // Time when the routing context is created.
+	PendingLoad float64   // Normalized pending load of request, available after AddRequestCount call. See cache.PendingLoadProvider
+	TraceTerm   int64     // Trace term identifier, available after AddRequestCount call.
+	RoutedTime  time.Time // Time consumed during routing.
 
 	targetPodSet chan struct{}
 	targetPod    atomic.Pointer[v1.Pod]
-	debugDelay   time.Duration
+	lastError    atomic.Pointer[error]
+	tokens       []int           // Cache of tokenized prompts
```
varungup90, line 58: can you elaborate the comment for tokens

zhangjyr: Do you mean how it is used or how it is created? For use, this is just a private cache because promptLength might be called multiple times during routing.

varungup90: Both, what tokens are stored and when used.

Given that now routing context as lot of metadata for request, in a follow up PR a readme can be added.

pkg/types/router_context.go
```diff
@@ -27,83 +28,244 @@ import (
...
+// Elapsed returns the elapsed time since the request was created.
+func (r *RoutingContext) Elapsed(currentTime time.Time) time.Duration {
+	return currentTime.Sub(r.RequestTime)
+}
+
+// PromptTokens returns the tokenized prompt of the request.
+func (r *RoutingContext) PromptTokens() ([]int, error) {
+	if r.tokens == nil {
+		var err error
+		r.tokens, err = utils.TokenizeInputText(r.Message)
```
varungup90, line 107: It will depend on the tokenizer used.

pkg/plugins/gateway/algorithms/least_load.go
```diff
@@ -0,0 +1,121 @@
...
+	for _, pod := range pods.All() {
+		if pod.Status.PodIP == "" {
+			// Pod not ready.
+			continue
+		}
+
+		util, err := r.provider.GetUtilization(ctx, pod)
+		if err != nil {
+			lastErr = r.updateError(lastErr, err)
+			klog.ErrorS(err, "Skipped pod due to fail to get utilization in leastLoadRouter", "pod", pod.Name)
```
varungup90: log it with request_id and this has the potential to flood the logs.

pkg/plugins/gateway/algorithms/least_load.go
```diff
@@ -0,0 +1,121 @@
...
+			continue
+		}
+
+		klog.V(5).Infof("pod: %v, podIP: %v, util: %.2f",
+			pod.Name, pod.Status.PodIP, util)
+
+		var consumption float64
+		if r.pulling {
```
varungup90, line 96: what does pulling mean here?

pkg/plugins/gateway/algorithms/slo.go
```diff
@@ -0,0 +1,87 @@
...
+package routingalgorithms
+
+import (
+	"github.com/vllm-project/aibrix/pkg/cache"
+	"github.com/vllm-project/aibrix/pkg/plugins/gateway/queue"
+	"github.com/vllm-project/aibrix/pkg/types"
+)
+
+const (
+	RouterSLO                 types.RoutingAlgorithm = "slo"
```
varungup90: can you add a readme for these algos

pkg/types/router_context.go
zhangjyr: utils.TokenizeInputText specifically uses tiktoken, I optimize the function by reuseing the tiktoken object.

pkg/plugins/gateway/algorithms/least_load.go
zhangjyr: "pulling" means the pull mode in the PR description: 
"Below is a comparison of pulling mode and default push mode:
Push mode: The router dispatches requests to the server, possibly overloading the server.
Pull mode: The server pulls requests from the router based on the server's capacity.
With profile support, the gateway now has server capacity knowledge and can achieve pull mode within the gateway."

pkg/plugins/gateway/algorithms/least_load.go
zhangjyr: Yes, I've added a new feature to avoid multiple error logs for the same request. In particular, a profile missing or not supporting can be a common problem for all pods. Only one error log is needed if that happens.

pkg/types/router_context.go
zhangjyr: The first time the PromptTokens() or PromptLength() is called, the tokens are cached in the context.  I primarily use prompt length for output prediction, so the prefix-cache method of tokenization may not be suitable. Although a model-specific tokenizer is preferred, the current output prediction only predicts on coarse-grained log2 buckets, so I think the tiktoken fits.

development/app/Makefile
```diff
@@ -113,5 +113,17 @@ test-gateway2:
             "max_tokens": 512 \
         }'
 
+test-router:
```
Xunzhuo, line 116: can this make target rename to `test-slo-router`?

pkg/cache/cache_api.go
Xunzhuo: Cannot we deploy the same model name in different namespace?

pkg/cache/cache_init.go
```diff
@@ -108,41 +128,107 @@ func New(redisClient *redis.Client, prometheusApi prometheusv1.API) *Store {
 	return store
 }
 
-func NewTestCacheWithPods(pods []*v1.Pod, model string) *Store {
-	c := &Store{}
+// NewForTest initializes the cache store for testing purposes, it can be repeated call for reset.
+func NewForTest() *Store {
```
Xunzhuo, line 132: Can move to cace_test.go?

pkg/cache/cache_profile.go
```diff
@@ -0,0 +1,70 @@
...
+		default:
+		}
+
+		keys, cursor, err = c.redisClient.Scan(ctx, cursor, "aibrix:profile_*", 10).Result() // 10 is the count, adjust as needed.  Use a larger count in production.
+		if err != nil {
+			klog.Errorf("failed to scan deploymet profileds due to: %v", err)
+			return
+		}
+
+		for _, key := range keys {
```
Xunzhuo, line 42: Thoughts on doing it concurrently?

pkg/cache/errors.go
```diff
@@ -0,0 +1,57 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package cache
+
+import "errors"
+
+var (
+	ErrorTypeMetricNotFound = &CacheError{error: errors.New("metric not found")}
```
Xunzhuo, line 21: Besides these two errors, any other errors in Cache can abstracted?

pkg/cache/model_gpu_profile.go
```diff
@@ -0,0 +1,193 @@
...
+	ErrProfileNoTPOT       = fmt.Errorf("profile has no TPOT data")
+)
+
+const ModelGPUNameTemplate = "aibrix:profile_%s_%s"
+
+// Assuming indexes are of equal distances, the signature_tolerance
+// control how a value in the middle of two indexes should aligned to a
+// certain index. While 0.5 indicate neutual preference. 0.25 set
+// threshold closer to lower index and then prefer higher index.
+const SignatureTolerance = 0.5
```
Xunzhuo, line 42: Can this be configurable?

pkg/cache/model_gpu_profile.go
```diff
@@ -0,0 +1,193 @@
...
+
+	return boolVal
+}
+
+type ModelGPUProfile struct {
+	Deployment string      `json:"gpu"` // k8s deployment that specified model and GPU information.
+	Cost       float64     `json:"cost"`
+	Tputs      [][]float64 `json:"tputs"`   // Max RPS per correspondent index.
+	Indexes    [][]float64 `json:"indexes"` // [output tokens, input tokens]
+	Created    float64     `json:"created"`
```
Xunzhuo: What does the created mean here? It looks like a boolean var naming.

pkg/cache/model_gpu_profile_test.go
```diff
@@ -0,0 +1,46 @@
...
...3741334], [7.6749238663857575, 7.646377471511056, 7.469959534415053, 4.021491925138638, 2.0825276518486944], [3.8569367968050043, 3.8474025810402654, 3.788259676585335, 2.0288591107923666, 0.0], [0.0, 0.0, 0.0, 0.0, 0.0]], \"indexes\": [[4, 8, 16, 32, 64, 128, 256, 512], [128, 256, 512, 1024, 2048]], \"created\": 1748888478.258241, \"e2e\": [[0.09952875833841972, 0.12239365121233277, 0.21428067207685672, 0.31731165578588844, 0.5534239103773143], [0.17259948405786418, 0.19343522920506076, 0.3083386933861766, 0.43137065838091077, 0.6822485379409045], [0.2401115512580145, 0.34489422958809884, 0.47929574372596107, 0.6583689095533919, 0.940295041234931], [0.48604107543011194, 0.6922839175269473, 0.827185874566203, 1.0668736191641073, 1.4869349837116896], [0.8164577254245523, 0.9879068507999181, 1.041013903748244, 2.1203144049376714, 2.666910813357681], [1.51887807746185, 1.7140269558737053, 2.342019975812873, 2.50050834288937, 2.8703157891728917], [2.9778688216709996, 3.245595489592524, 3.9633901528839486, 4.021366707490524, 4.36835270334268], [5.479355860430514, 5.721908029631013, 6.244554732975084, 7.508401151361177, 11.322325259115313]], \"slos\": {\"percentile\": 99, \"e2e\": 5.0}}"
```
Xunzhuo, line 24: suggest to separate from code, into cache/testdata/xx-xxx-xx-profile.json, and the testcase can be extensible with different scenarios

pkg/cache/pending_load_provider.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/vllm-project/aibrix/pkg/metrics"
+	"github.com/vllm-project/aibrix/pkg/types"
+	v1 "k8s.io/api/core/v1"
+)
+
+// PendingLoadProvider estimate server utilization in terms of pending load:
+// PendingLoad = 1 / PendingRequests = 1 / (Throughput * Latency),
+// where PendingRequests = Throughput * Latency follows Little's Law,
+// and Throughput(RPS) and Latency are from loaded profile per feature (input tokens, output tokens)
+type PendingLoadProvider struct {
```
Xunzhuo, line 28: Can add a test case for the pending load provider?

pkg/plugins/gateway/algorithms/fallback.go
```diff
@@ -0,0 +1,47 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+package routingalgorithms
+
+import (
+	"github.com/vllm-project/aibrix/pkg/types"
+)
+
+const DefaultFallbackAlgorithm types.RoutingAlgorithm = RouterRandom
```
Xunzhuo, line 22: So if the algorithm does explicitly set the fallback strategy, we use random, right?

Xunzhuo: Can we remove the pkg/plugins/gateway/algorithms/util.go `SelectRandomPodAsFallback` instead of the current elegant way?

test/e2e/routing_strategy_test.go
```diff
@@ -77,6 +77,16 @@ func TestRandomRouting(t *testing.T) {
 		chi2Stat, significanceLevel)
 }
 
+func TestSLORouting(t *testing.T) {
+	iterration := 10
+
+	for i := 0; i < iterration; i++ {
+		req := "hello test"
+		targetPod := getTargetPodFromChatCompletion(t, req, "slo")
+		assert.NotEmpty(t, targetPod, "target pod should not be empty")
```
Xunzhuo: Nit, can we enrich the e2e?

Xunzhuo (commented):
Thanks for the works, left some small comments!

pkg/cache/cache_profile.go
```diff
@@ -0,0 +1,70 @@
...
+			}
+
+			var updated ModelGPUProfile
+			err = updated.Unmarshal(val)
+			if err != nil {
+				klog.Warningf("error unmarshalling deployment profiles for %s: %v, Value: %s", key, err, val)
+				continue // Skip to the next key
+			}
+
+			c.UpdateModelProfile(key, &updated, false)
```
Xunzhuo, line 63: a question for this, I wonder who/where added the gpu profile cache to redis, who is the producer?

development/app/Makefile
zhangjyr: In fact, test-router is just for showcases. I can change the strategy to least-request.

pkg/cache/cache_init.go
zhangjyr: This makes sense. However, this set of functions are designed to be called by tests in other packages (e.g., routing algorithm), moving NewForTest() to cache_test.go will make them inaccessible from there.

pkg/cache/cache_profile.go
zhangjyr: The profile update is off the request path, and there is no urgency to keep the profile real-time. On the other side, concurrent update can cause unnecessary network burst (redis get), memory footprint(store bytes), and CPU footprint(unmarshal).

pkg/cache/errors.go
zhangjyr: There are many. But this PR is not really about Error standardization. I created this abstraction for my part of the code, and hopefully, other collaborators will come to the realization of the necessity and adopt this approach.

pkg/cache/model_gpu_profile.go
zhangjyr: Yes, but not needed. This constant is here just for programming and evaluation flexibility. On a log2 scale bucket, 0.5 rounds more elements to a higher bucket, which leads to adopting a more conservative service profile. This is what we expect in an SLO-aware policy.

pkg/cache/model_gpu_profile.go
zhangjyr: It is a timestamp in Unix sec.sub-sec format, which is used to test if the profile has been updated. I will add a comment for this field.

pkg/cache/model_gpu_profile_test.go
zhangjyr: Good idea, but let's make the change if more test cases are added here.

pkg/cache/pending_load_provider.go
zhangjyr: What test cases would you like to be added?

pkg/plugins/gateway/algorithms/fallback.go
zhangjyr: > So if the algorithm does explicitly set the fallback strategy, we use random, right?

If the algorithm does not explicitly set the fallback, random will be used. Yes.

> Can we remove the pkg/plugins/gateway/algorithms/util.go `SelectRandomPodAsFallback` instead of the current elegant way?

Let's open a new code refactoring issue after this PR is merged. This PR has already been too large, I'm refraining from further refactoring.

test/e2e/routing_strategy_test.go
zhangjyr: Lol, you got me. The test was there because I thought unit testing could not be done for the slo policy. However, the current slo_test.go into the algorithm covers most e2e tests. I will remove this part of tests.

pkg/cache/cache_profile.go
zhangjyr: The submission of profiles is a part of the previous GPU optimizer PR. See step 4 of https://aibrix.readthedocs.io/latest/features/heterogeneous-gpu.html for details.

pkg/cache/model_gpu_profile.go
Xunzhuo: Thoughts on `CreatedAt`?

pkg/cache/cache_profile.go
Xunzhuo: Thanks!

Xunzhuo (approved):
Thanks! This is massive, and in case no more conflicts, will prioritize the PR merge

zhangjyr:
@Jeffwan, are there further comments?

Issue #1205, opened 2025-06-18
Support embeddings endpoints


gcalmettes:
### 🚀 Feature Description and Motivation

With the [validation of the request body recently added](https://github.com/vllm-project/aibrix/pull/1067), embeddings request cannot be validated under aibrix.
It would be nice to also support the `/v1/embeddings` endpoint in addition to the `/v1/chat/completions` and `/v1/completions` endpoints already supported. 

### Use Case

This would be useful in order to use AiBrix for embeddings-related workloads

### Proposed Solution

I'll open a dedicated PR to add embeddings support

Jeffwan:
This is important and feel free to open a PR. /cc @varungup90

Jeffwan:
part of #846

robin2008:
Will the PR also address the rerank endpoints?

Jeffwan:
@robin2008 rerank needs a separate endpoint, we will add it soon. could you confirm which endpoint are you using now?

@varungup90 please check `/rerank` `/v1/rerank` and `/v2/rerank` interface in `vllm/entrypoints/openai/api_server.py`

robin2008:
we are using vllm `/v1/rerank` endpoint provided by https://github.com/vllm-project/vllm/pull/12376, thanks.

varungup90:
This is completed.

Issue #1175, opened 2025-06-04
[httproute] add  /v1/version is not found


ying2025:
### 🚀 Feature Description and Motivation

 I add `/v1/version` method, and it's not found, where is it block? I add the`/v1/models` method, it return all the models, and I find it visit `aibrix-metadata-service` service.What should I do if I want use `/v1/version` and `/v1/models`

### Use Case

```
    matches:
    - headers:
      - name: model
        type: Exact
        value: nodetest128-w73sg65c
      path:
        type: PathPrefix
        value: /v1/completions
    - headers:
      - name: model
        type: Exact
        value: nodetest128-w73sg65c
      path:
        type: PathPrefix
        value: /v1/chat/completions
    - headers:
      - name: model
        type: Exact
        value: nodetest128-w73sg65c
      path:
        type: PathPrefix
        value: /v1/models
    - headers:
      - name: model
        type: Exact
        value: nodetest128-w73sg65c
      path:
        type: PathPrefix
        value: /v1/version
```

varungup90:
You need to create a [reserved route](https://github.com/vllm-project/aibrix/blob/main/config/gateway/gateway-plugin/gateway-plugin.yaml#L120) for /v1/version that will have backend ref as aibrix-metadata-service and implementation [here](https://github.com/vllm-project/aibrix/blob/main/pkg/metadata/handlers.go#L54).

ying2025:
> aibrix-metadata-service

aibrix-metadata-service return all the models, and it lack of model related information. Do you mean forwarding through aibrix metadata service to the corresponding interface of the specific pod?

Xunzhuo:
@varungup90 we should consider creating platform related CR, not surfacing end-user

ying2025:
Otherwise, for some requests in non-standard OpenAPI formats maybe also consider compatibility

varungup90:
> @varungup90 we should consider creating platform related CR, not surfacing end-user

That's my goal to provide pointers to add /v1/version.

@ying2025 To clarify, /v1/version is not supported right now by AIBrix. If you are interested please help in adding it, shared some pointers above to help in writing code.

ying2025:
> > [@varungup90](https://github.com/varungup90) we should consider creating platform related CR, not surfacing end-user
> 
> That's my goal to provide pointers to add /v1/version.
> 
[...]

PR #1208, opened 2025-06-19
[API] add support for embeddings api

gcalmettes:
## Pull Request Description

This PR adds the support for the `/v1/embeddings` openAI api endpoint in addition to the currently supported `/v1/chat/completions` and `/v1/completions` endpoints

Implementation considerations:
- currently, for every models declared, AIBrix a HTTPRoute matching  `/v1/chat/completions` and `/v1/completions` path prefixes.  In order to not create routes that are not supported by the model (or routes that a user wants to not be available to a model if the model supports several modalities), this PR adds a new optional label `model.aibrix.ai/supported-routes` on the deployment that allows to selectively define the `HTTPRouteMatch` to be created. If the label is not present, then the default routes will be created (`/v1/chat/completions` and `/v1/completions`)

## Related Issues
Resolves: https://github.com/vllm-project/aibrix/issues/1205

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>


pkg/controller/modelrouter/modelrouter_controller.go
```diff
@@ -107,6 +113,43 @@ func Add(mgr manager.Manager, runtimeConfig config.RuntimeConfig) error {
 	return err
 }
 
+// getSupportedRoutesMatchFromLabelsOrDefault returns the HTTPRouteMatch based on the model route labels value
+func getSupportedRoutesMatchFromLabelsOrDefault(labels map[string]string, modelHeaderMatch gatewayv1.HTTPHeaderMatch) []gatewayv1.HTTPRouteMatch {
+	labelValueToRoutePathPrefix := map[string][]string{
+		modelRouteEmbeddings:      {"/v1/embeddings"},
+		modelRouteChatCompletions: {"/v1/completions", "/v1/chat/completions"},
```
scwgoire: Those should probably be stored in constants, they are used in `validateRequestBody`

gcalmettes: Yes definitely!
I am actually refactoring the whole logic a bit as we need to also process the response body differently based on the request type (chat completions, embeddings, etc ...).
I'll also store the strings in constants so we can use them in all the methods without duplicating them

gcalmettes:
@Jeffwan @varungup90 what would be the best way to discuss implementation details regarding the changes introduced by this PR ?

The current code of the aibrix plugins is tightly coupled to the chat-completions/completions api of openAI (e.g.: the responses body are directly unmarshalled in `openai.ChatCompletion` (or `openai.ChatCompletionChunk` if stream) in the `HandleResponseBody`). So adding the support for the embeddings api in fact also becomes the basis to deferentially process the body based on the request type: chat completions vs embeddings (and vs other request types, audio, images, later on, as aibrix matures and can handle other modalities).

I made an [implementation proposition](https://github.com/gcalmettes/aibrix/blob/b1cd8401e2440c4bc259357ab83ae1035b2f8b6b/pkg/plugins/gateway/gateway_rsp_body.go#L38) in this PR (I still have to work on it as right now the `handleChatCompletionsResponseBody` and `handleEmbeddingsResponseBody` method have too much duplicated code (and in fact only the types the request and the usage are unmarshalled into are really changing).
But I'll be happy to discuss the implementation details in case you guys had already brainstormed about this use case.

pkg/plugins/gateway/util.go
```diff
@@ -142,6 +161,48 @@ func getChatCompletionsMessage(requestID string, chatCompletionObj openai.ChatCo
...
+	case *[][]int64:
+		size = len(*input)
+	default:
+	}
+
+	if size == 0 {
+		klog.ErrorS(nil, "no input in the request body", "requestID", requestID)
+		return buildErrorResponse(envoyTypePb.StatusCode_BadRequest, "no messages in the request body", HeaderErrorRequestBodyProcessing, "true")
+	}
+	if size > 1024 {
```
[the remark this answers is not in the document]
gcalmettes: done :heavy_check_mark: , and used the default value used by OpenAI (2048) instead

pkg/controller/modelrouter/modelrouter_controller.go
gcalmettes: done :heavy_check_mark:

Jeffwan:
@gcalmettes are you in the slack channel? we can talk about more implementation details there? a google doc also works

Jeffwan:
@varungup90 Since we're aiming to support API compatibility, could you also help review the change proposed by @gcalmettes? The coupling issue raised is indeed a concern for future extensibility. /cc @Xunzhuo

Xunzhuo:
assign myself for track the review, will schedule some bandwidth soon.

gcalmettes:
@Jeffwan yes I am in the slack channel (I believe you're talking about the vllm slack correct ? Or is there another slack I should join ?)

Jeffwan:
@gcalmettes Yes, the exact channel we discussed in.

varungup90:
This is simple representation for current architecture.
image

For adding /v1/embeddings, or there is another [issue](https://github.com/vllm-project/aibrix/issues/1175) to add /version and /metrics. They can be addressed with different options proposed below

Option 1 - Current implementation takes option 1 approach. 
- Current implementation of chat plugin is designed for completion endpoint and is not relevant for embedding endpoint.
- Better long term approach will be to use option 4, and we build a new gateway plugin for embedding.

Option 2 - In option 2, request routes to metadata service component which already handles /v1/models endpoint. Here some basic validation can be done and light weight changes to input/output response.

Option 3 - Here, request directly routes to model but user will need to include some extra headers which is slightly different then what user would have done if directly querying vllm. Option 2, here can help with appending those extra headers and users can have simple request without providing extra headers.

Option 4: Build a new aibrix gateway plugin, specifically designed for handling /v1/embedding. To justify building a new plugin, we would need complex use cases similar to completions (advanced routing, KV cache offloading, P/D etc).

From the above options, I think option 2 provides right balance. Easy to implement, and users request does not need to provide extra metadata for routing. In future, as the /v1/embedding API call evolves, we can move to option 4. cc @Xunzhuo @Jeffwan thoughts?

image

scwgoire:
Very nice analysis!

My 2 cents on this one: options 1 or 4 seem the best to me. I think the plugin(s) should be seen as a smart router where each route has a set of available routing strategies.

- Option 1: requires a large refactoring but doesn't require spawning multiple plugins
- Option 4: can still benefit from code-sharing and just differ by the payload parsing, just a little more complex to deploy and needs to add a "plugin selection by path"

In both cases, Embeddings would definitely benefit from most of the routing strategies already implemented in the chat completions plugin.

This would also open the path to other OpenAI routes such as audio transcriptions (whisper is coming to vLLM) which have way different payloads.

Option 2 seems very counterintuitive to me, adding a data path into metadata service doesn't look right in the "big picture".

I'm pretty new to aibrix, so I may not be the right guy to fully understand your rationale, so take all of this with an XXL grain of salt 😁 

Cheers!

PS: option 3 is always nice to be available, as a fallback for not-yet-implemented or edge cases

Jeffwan:
@gcalmettes Do you get time to polish this change? it would be great to add it into v0.4.0 release /cc @varungup90

gcalmettes:
@Jeffwan yes I can definitely rework the implementation.
However, it seems the discussion about which option to tackle is not necessarily settled :smile: 

I aggree that currently option 1 is tightly bound to chat-models (in particular, [the `message` and `stream` notions](https://github.com/vllm-project/aibrix/blob/7fe0a91d2f7e82bdc054539aa22bf6aa98dd5e13/pkg/plugins/gateway/gateway_req_body.go#L39) are not relevant to embeddings), however it seems to me that if we go with option 3 we loose all the smart routing that aibrix provides when requesting a model. In particular, algorithms such as leas_busy, least_request, are still relevant to the embeddings models.
Tokens count extractions would also need to be reimplemented in the metadata service if option 3 is chosen.

Jeffwan:
@gcalmettes @varungup90 Since the engine itself provides the `/v1/embedding` API, I believe option 1 is the right approach for now. Technically, the gateway plugin should support multiple endpoints, but it's currently bound to the chat or chat completion APIs. If it's just a matter of code refactoring, I believe it's acceptable. However, if there are any technical blockers you're aware of, feel free to bring them up—we can discuss and address them together.  Let me know your thoughts.

BTW, `/v1/models` etc is a little bit different since we need model aggregation.  option 2 is more propitiate for batch api etc, kind of an extension server if vLLM etc doesn't provide the openai compatible endpoints.

varungup90:
> @gcalmettes @varungup90 Since the engine itself provides the `/v1/embedding` API, I believe option 1 is the right approach for now. Technically, the gateway plugin should support multiple endpoints, but it's currently bound to the chat or chat completion APIs. If it's just a matter of code refactoring, I believe it's acceptable. However, if there are any technical blockers you're aware of, feel free to bring them up—we can discuss and address them together. Let me know your thoughts.
> 
> BTW, `/v1/models` etc is a little bit different since we need model aggregation. option 2 is more propitiate for batch api etc, kind of an extension server if vLLM etc doesn't provide the openai compatible endpoints.

What I would have preferred is to build a new plugin for embedding and refactor the code for current chat pluggin such that all the features for routing or token count can be used by any plugin in future as well. For now, given the urgency lets proceed with current implementation and in future lets refactor the code such that any future plugin can easily be implemented.

gcalmettes:
@varungup90 agreed.
I'll isolate the changes in dedicated functions as much as possible, so it'll be easier to extract the logic for embeddings in its own plugin when we switch on a dedicated plugin for this endpoint.

happyandslow:
Just checking the status of this PR @gcalmettes @varungup90  Are we planning on merging this feature soon?

gcalmettes:
superseeded by #1570

Issue #1216, opened 2025-06-22
Add Comprehensive Tests and Test Infrastructure Improvements for Gateway Server


ModiCodeCraftsman:
### 🚀 Feature Description and Motivation

Add comprehensive unit and integration tests for the gateway server component, including request body handling, routing logic, and error scenarios.

### Use Case

Currently, the gateway server’s core logic lacks thorough automated test coverage. This makes it harder to confidently refactor, extend, or debug the gateway’s routing and request handling features. By introducing robust tests, we can ensure reliability, catch regressions early, and facilitate future development.

### Proposed Solution

Implement new unit tests for the gateway server’s request body handling and routing logic.
Add test helpers and mocks to simplify and standardize test setup.
Refactor and improve the e2e test script for better automation and reliability.
Ensure all new and existing tests pass and provide meaningful coverage for critical gateway paths.

Jeffwan:
@ModiCodeCraftsman I see you cut an PR earlier. Are you interested in contributing the more tests and test infra? If so, please go ahead and let us know if anything we can help. If you expect other maintains to fill the gap, let us know as well.

ModiCodeCraftsman:
@Jeffwan Those tests are added as part of [this issue](https://github.com/vllm-project/aibrix/issues/1101). Since this Gateway server is in that area, I have added tests to improve coverage before making the changes. 

@varungup90 FYI

PR #1217, opened 2025-06-22
Add new test cases for gateway server

ModiCodeCraftsman:
This PR adds comprehensive unit tests and supporting test utilities for the gateway server component. The main changes include:

New Tests:
- Added gateway_req_body_test.go to cover request body handling logic, including routing, error handling, and edge cases.
- Added gateway_test_helpers.go with mock implementations for easier and more robust testing.
- Extended gateway_test.go with new tests for pod selection and routing strategies.

Test Infrastructure Improvements:

- Refactored and improved the e2e test script (run-e2e-tests.sh) for better port-forward management and cleanup.

[Issue](https://github.com/vllm-project/aibrix/issues/1216)


pkg/plugins/gateway/algorithms/router.go
```diff
@@ -31,6 +34,8 @@ var routerConstructor = map[types.RoutingAlgorithm]types.RouterProviderRegistrat
 
 // Validate validates if user provided routing routers is supported by gateway
 func Validate(algorithms string) (types.RoutingAlgorithm, bool) {
+	routerMu.RLock()
```
varungup90, line 37: due to data race condition for router factory variable?

pkg/plugins/gateway/gateway_test.go
```diff
@@ -143,3 +149,146 @@ func Test_buildEnvoyProxyHeaders(t *testing.T) {
...
+	// Define test cases for different pod selection and error scenarios
+	tests := []struct {
+		name          string
+		pods          types.PodList
+		mockSetup     func(*mockRouter, types.RoutingAlgorithm)
+		expectedError bool
+		expectedPodIP string
+	}{
+		{
+			name: "routing.Select returns error",
```
varungup90: In this test, router.Route() is returning error, not the routing.Select(). Please update the test description.

pkg/plugins/gateway/gateway_test.go
```diff
@@ -143,3 +149,146 @@ func Test_buildEnvoyProxyHeaders(t *testing.T) {
...
+			pods: &mockPodList{pods: []*v1.Pod{{
```
varungup90: Can you help me understand why mockPodList is needed here, directly using podList would be fine too.

pkg/plugins/gateway/gateway_test.go
```diff
@@ -143,3 +149,146 @@ func Test_buildEnvoyProxyHeaders(t *testing.T) {
...
+				// No expectations needed as no ready pods
+			},
+			expectedError: true,
+		},
+		{
+			name: "single ready pod",
+			pods: &mockPodList{pods: []*v1.Pod{{
+				Status: v1.PodStatus{
+					PodIP:      "1.2.3.4",
+					Conditions: []v1.PodCondition{{Type: v1.PodReady, Status: v1.ConditionTrue}},
```
varungup90, line 224: Also add one more pod with ConditionFalse.

pkg/plugins/gateway/gateway_test.go
varungup90: For these tests, even random router can be used.

pkg/plugins/gateway/algorithms/router.go
ModiCodeCraftsman: Yes, this ensures thread-safe concurrent R/W and prevents data race conditions.

pkg/plugins/gateway/gateway_test.go
ModiCodeCraftsman: > Can you help me understand why mockPodList is needed here, directly using podList would be fine too.

The original motivation for using a mockedPodList type was to avoid potential issues arising from the internal logic of certain PodArray methods (such as ListByIndex and related indexing behavior) during testing. This allowed us to isolate test scenarios from the implementation details of PodArray. However, following the review and for consistency, the mock type has now been removed and all relevant tests have been updated to use the actual PodArray type directly.

> For these tests, even random router can be used.

My intention with the current approach was to use a mock router to precisely control the behavior of the Route function, including simulating errors and verifying whether Route is called in specific scenarios. This allows for more granular testing of error handling and call expectations within the server logic.

If your suggestion was to simplify the tests by using the actual random router implementation, could you please clarify if there are specific scenarios where you think the mock is unnecessary, or if there’s a particular benefit you see in using the real router here? I want to make sure I fully understand your point and am happy to adjust the tests if needed.

pkg/plugins/gateway/gateway_test.go
ModiCodeCraftsman: Added the "single ready pod out of two" test case

pkg/plugins/gateway/gateway_test.go
ModiCodeCraftsman: Done.

pkg/plugins/gateway/gateway_req_body_test.go
```diff
@@ -0,0 +1,516 @@
...
+						assert.Contains(t, []string{"1.2.3.4:8000", "5.6.7.8:8000"}, targetPodIP, "Target pod IP should be one of the pod IPs from the mock setup")
+					}
+				}
+				assert.True(t, foundRoutingStrategy, "HeaderRoutingStrategy not found")
+				assert.True(t, foundTargetPod, "HeaderTargetPod not found")
+			},
+			checkStream: false,
+		},
+		{
+			name:        "no routable pods available - should return ServiceUnavailable",
```
varungup90, line 338: There are 4 tests here.
- pod len == 0 (missing).
- pod condition false (present)
- pod IP not set (present)
- pod in terminating state (missing).

Can you add the missing two cases.

pkg/plugins/gateway/gateway_test.go
varungup90: I think its ok to use mock router for most gateway tests. Later I can add one simple test e2e without mock router and mock cache.

varungup90:
lgtm. Nit comment to add two missing cases for service unavailable.

pkg/plugins/gateway/gateway_req_body_test.go
ModiCodeCraftsman: Sure, done.

ModiCodeCraftsman:
@varungup90 Could you please review and resolve the open discussions if you’re satisfied with the changes? I believe it’s best for the reviewer to close discussions once their requests have been addressed, to ensure nothing is missed. Let me know if there’s anything else you’d like me to update. Thanks!

Jeffwan:
@varungup90 any further comments for this change?

varungup90 (approved):

Issue #843, opened 2025-03-11
[RFC] Support inference engine SGLang


Belyenochi:
### Feature Description
I propose integrating the SGLang inference engine into AIBrix, as it has rapidly emerged as a leading high-performance inference framework in the community. According to our internal benchmarks and recent open-source data (e.g., [SGLang v0.4.3 release notes](https://github.com/sgl-project/sglang)), SGLang outperforms vLLM by approximately 10% in throughput, particularly in single-machine and small-cluster environments, with notable advantages in long-context scenarios (thanks to FlashInfer MLA optimizations). Furthermore, SGLang’s native support for models like DeepSeek-R1 (e.g., FP8 and multi-head latent attention MLA) makes it highly promising for resource-constrained private cloud settings.

### Motivation
AIBrix’s core design philosophy is to provide a modular, pluggable cloud-native inference infrastructure, empowering users to select optimal tools for their specific needs. While AIBrix currently integrates vLLM deeply (e.g., as seen in `config/overlays/release`), adding SGLang will further enrich the options, especially for the following scenarios:
- **Private Cloud Optimization**: In environments with limited hardware (e.g., 1-2 machines with 8 GPUs each), SGLang’s zero-overhead batch scheduling and cache-aware load balancing significantly boost efficiency.
- **Diverse Model Support**: SGLang’s optimizations for models like DeepSeek-R1 and LLaMA (e.g., tensor parallelism and data-parallel attention) align with AIBrix’s distributed inference goals.
- **Community Trends**: SGLang’s rapid iteration (e.g., support for DeepSeek V3/R1 in January 2025) and widespread adoption (e.g., by xAI, LMSYS) indicate it is becoming a mainstream choice for inference frameworks.

By integrating SGLang, AIBrix will offer private and hybrid cloud users a higher-performance inference option, reinforcing its position as a comprehensive LLM infrastructure platform. This not only complements the existing vLLM support but also responds to the community’s demand for flexibility and performance optimization.

---
### Real-World Scenario
[...]

PR #1293, opened 2025-07-16
[Feature] Adding raw metrics name conversion in metrics

happyandslow:
## Pull Request Description
Adding multi engine support by adapting different metrics exposed. 

## Related Issues
Resolves: #843


pkg/cache/cache_metrics.go
```diff
@@ -33,6 +33,11 @@ const (
 	// the request port and metrics port may differ, so a dedicated metrics port is required.
 	MetricPortLabel                     = "model.aibrix.ai/metric-port"
 	defaultPodMetricRefreshIntervalInMS = 50
+	engineLabel                         = "model.aibrix.ai/engine"
+	portLabel                           = "model.aibrix.ai/port"
+	modelLabel                          = "model.aibrix.ai/name"
+	defaultEngineLabel                  = "vllm"
```
Jeffwan: defaultEngineLabelValue

pkg/cache/cache_metrics.go
```diff
@@ -250,7 +254,7 @@ func (c *Store) updateQueryLabelMetricFromRawMetrics(pod *Pod, allMetrics map[st
 		}
 		rawMetricName := metric.RawMetricName
 		scope := metric.MetricScope
-		metricFamily, exists := allMetrics[fmt.Sprintf("vllm:%s", rawMetricName)]
```
Jeffwan, line 253: my point is we should always use full metrics name instead of `vllm:%s` which is confusing. Let's cut a short PR to fix this part first.

pkg/cache/cache_metrics.go
```diff
@@ -330,14 +334,37 @@ func (c *Store) queryUpdatePromQLMetrics(ctx context.Context, metric metrics.Met
 	return nil
 }
 
+func (c *Store) fetchMetrics(pod *Pod, allMetrics map[string]*dto.MetricFamily, labelMetricName string) (*dto.MetricFamily, bool) {
+	metric, exists := metrics.Metrics[labelMetricName]
```
Jeffwan, line 335: labelMetricName is the aibrix normalized metrics name, right?

pkg/cache/cache_metrics.go
```diff
@@ -330,14 +334,37 @@ func (c *Store) queryUpdatePromQLMetrics(ctx context.Context, metric metrics.Met
...
+	if !exists {
+		klog.V(4).Infof("Cannot find labelMetricName %v in collected metrics names", labelMetricName)
+		return nil, false
+	}
+	engineType := getPodLabel(pod, engineLabel, defaultEngineLabel)
+	rawMetricName, ok := metric.RawMetricNameMapping[engineType]
```
Jeffwan: here it tries to convert to engine specific name.

pkg/cache/cache_metrics.go
```diff
@@ -380,3 +407,12 @@ func getPodMetricPort(pod *Pod) int {
 	}
 	return defaultMetricPort
 }
+
+func getPodLabel(pod *Pod, labelName string, defaultValue string) string {
```
Jeffwan: do we have such utils in utils.go etc? can you search references?

pkg/metrics/types.go
```diff
@@ -77,12 +77,13 @@ const (
...
-	MetricType    MetricType
-	PromQL        string // Optional: Only applicable for PromQL-based metrics
-	RawMetricName string // Optional: Only applicable for QueryLabel-based metrics
-	Description   string
-	MetricScope   MetricScope
+	MetricSource         MetricSource
+	MetricType           MetricType
+	PromQL               string            // Optional: Only applicable for PromQL-based metrics
+	RawMetricName        string            // Optional: Only applicable for QueryLabel-based metrics
+	RawMetricNameMapping map[string]string // Optional: Mapping from engine type to raw metric name.
```
Jeffwan: with the mapping, the rawMetricsName should be logical metrics names? We need to consider the compatibility as well

pkg/plugins/gateway/algorithms/least_util.go
```diff
@@ -0,0 +1,81 @@
...
+		cache: c,
+	}, nil
+}
+
+func (r leastUtilRouter) Route(ctx *types.RoutingContext, readyPodList types.PodList) (string, error) {
+	var targetPod *v1.Pod
+	minBusyTimeRatio := math.MaxFloat64 // <= 1 in general
+
+	for _, pod := range readyPodList.All() {
+		busyTimeRatio, err := r.cache.GetMetricValueByPod(pod.Name, pod.Namespace, metrics.GPUBusyTimeRatio) // todo: replace mock
```
Jeffwan: this is same as least_busy_time?

pkg/plugins/gateway/algorithms/least_gpu_cache.go
```diff
@@ -0,0 +1,95 @@
...
+
+func (r leastGpuCacheRouter) Route(ctx *types.RoutingContext, readyPodList types.PodList) (string, error) {
+	var targetPod *v1.Pod
+	minGpuCache := math.MaxFloat64
+
+	for _, pod := range readyPodList.All() {
+		// Due to metric refactor (pull/543) to better support lora and multi models,
+		// we change to use PodModelMetrics instead of PodMetrics in some scenarios.
+		// This works but doesn't look very promising, we can revisit this part later.
+		gpuCache, err := r.cache.GetMetricValueByPodModel(pod.Name, pod.Namespace, ctx.Model, metrics.GPUCacheUsagePerc)
```
Jeffwan: this is ok

pkg/cache/cache_metrics.go
```diff
@@ -33,6 +33,11 @@ const (
 	// the request port and metrics port may differ, so a dedicated metrics port is required.
 	MetricPortLabel                     = "model.aibrix.ai/metric-port"
 	defaultPodMetricRefreshIntervalInMS = 50
+	engineLabel                         = "model.aibrix.ai/engine"
```
Jeffwan, line 33: do we have const in some common files?

pkg/cache/cache_metrics.go
happyandslow: Fixed

pkg/cache/cache_metrics.go
happyandslow: Yes

pkg/cache/cache_metrics.go
happyandslow: Yes -- right now `metric.RawMetricNameMapping` contains mapping between aibrix metrics name to engine-specific metrics name

pkg/cache/cache_metrics.go
happyandslow: Adding a utils.go in cache package

pkg/plugins/gateway/algorithms/least_util.go
happyandslow: Yes -- I'm now moving xllm's engine utilization metrics policy to least_utils. And the metrics names are separated as well.

pkg/cache/cache_metrics.go
happyandslow: Right now all metrics described at the top of the metrics.go are engine-agnostic (aibrix:) and only the metadata associated with each metrics name shows whether they have been adapted to different metrics.

pkg/cache/cache_metrics.go
happyandslow: It seems most usages of labels I've found are like [this](https://github.com/vllm-project/aibrix/blob/7fe0a91d2f7e82bdc054539aa22bf6aa98dd5e13/pkg/cache/informers.go#L37). If there should be a common file let's do a full refactor later.

pkg/metrics/types.go
happyandslow: Right now I'm only adding the EngineMetricsNameMapping field to query label metrics. If we find compatible metrics for other engine we can add those in the mappings.

pkg/cache/cache_metrics.go
```diff
@@ -28,11 +27,17 @@ import (
...
-	defaultMetricPort = 8000
 	// When the engine's HTTP proxy is separated from the engine itself,
 	// the request port and metrics port may differ, so a dedicated metrics port is required.
 	MetricPortLabel                     = "model.aibrix.ai/metric-port"
+	engineLabel                         = "model.aibrix.ai/engine"
+	portLabel                           = "model.aibrix.ai/port"
+	modelLabel                          = "model.aibrix.ai/name"
+	defaultMetricPort                   = 8000
+	defaultEngineLabelValue             = "vllm"
+	defaultModelName                    = ""
```
Jeffwan: do we need this? seems "" is weird. if there's no model found, we should ignore this pod or throw warning

pkg/cache/cache_metrics.go
```diff
@@ -44,6 +49,7 @@ var (
 		metrics.AvgGenerationThroughputToksPerS,
 		metrics.GPUCacheUsagePerc,
 		metrics.CPUCacheUsagePerc,
+		metrics.EngineUtilization,
```
Jeffwan, line 51: if you put the engine here, does vLLM deployment also try to find this metrics?

pkg/cache/cache_metrics.go
```diff
@@ -330,14 +332,37 @@ func (c *Store) queryUpdatePromQLMetrics(ctx context.Context, metric metrics.Met
...
+
 // Update `PodMetrics` and `PodModelMetrics` according to the metric scope
 // TODO: replace in-place metric update podMetrics and podModelMetrics to fresh copy for preventing stale metric keys
 func (c *Store) updatePodRecord(pod *Pod, modelName string, metricName string, scope metrics.MetricScope, metricValue metrics.MetricValue) error {
 	if scope == metrics.PodMetricScope {
 		pod.Metrics.Store(metricName, metricValue)
 	} else if scope == metrics.PodModelMetricScope {
 		if modelName == "" {
-			return fmt.Errorf("modelName should not be empty for scope %v", scope)
+			modelName = getPodLabel(pod, modelLabel, defaultModelName)
```
Jeffwan: let's update the method to return `name, bool`, defaultModleName ="" is kind of weird

pkg/cache/utils.go
```diff
@@ -0,0 +1,46 @@
+/*
+Copyright 2024 The Aibrix Team.
```
Jeffwan: 2025

pkg/metrics/metrics.go
```diff
@@ -40,6 +40,7 @@ const (
 	GPUCacheUsagePerc                    = "gpu_cache_usage_perc"
 	GPUBusyTimeRatio                     = "gpu_busy_time_ratio"
 	CPUCacheUsagePerc                    = "cpu_cache_usage_perc"
+	EngineUtilization                    = "engine_utilization"
```
Jeffwan, line 43: logical name is still engine utilization? we discussed to use some batch percetage etc.

pkg/metrics/metrics.go
```diff
@@ -88,6 +99,9 @@ var (
 			MetricType: MetricType{
 				Raw: Gauge,
 			},
+			EngineMetricsNameMapping: map[string]string{
```
Jeffwan, line 102: sglang doesn't have such metrics?

pkg/cache/utils.go
happyandslow: fixed

pkg/cache/cache_metrics.go
happyandslow: Fixed

pkg/cache/cache_metrics.go
happyandslow: removed

pkg/cache/cache_metrics.go
happyandslow: Check the current failing logic [here](https://github.com/vllm-project/aibrix/blob/61fd8fbf2d8a73fe829e04cbf6463d347f189b15/pkg/cache/cache_metrics.go#L343). If the mapping couldn't be found then it will output a log entry but won't throw any exceptions.

Jeffwan: @happyandslow please do not resolve comment without reply

pkg/metrics/metrics.go
happyandslow: adding sglang:gen_throughput but I can't find input throughput

pkg/metrics/metrics.go
happyandslow: I'm thinking we should stick to engine utilization for now since the decision of using batch percentages as measurement is completely internal decision of xllm folks (and they chose engine utilization to name this metrics so we should probably use this as well).

Jeffwan:
please wait me to merge this one. it has > 20 commits and directly merge will leads to messy commit logs

Issue #942, opened 2025-04-04
[RFC]: Sync KV cache from inference engine to gateway


varungup90:
### 🚀 Feature Description and Motivation

The gateway currently caches the input request along with the pod it was routed to. This cached information is then used to perform prefix matching for future requests, allowing them to be routed to the same pod that already holds the relevant KV cache locally.

However, gateway or pod restarts can potentially disrupt this mapping. The goal is to maintain near real-time synchronization of KV cache information from the pod to the gateway to ensure consistency.

For reference: [dynamo](https://github.com/ai-dynamo/dynamo/blob/main/docs/kv_cache_routing.md#dynamo-events) uses pub-sub to sync the state.

### Use Case

NA

Jeffwan:
@ae86zhizhi shows interest in this story. Feel free to pick this up

PR #1349, opened 2025-07-29
feat: Add KV cache event synchronization system

ae86zhizhi:
## Pull Request Description
Implement a comprehensive KV cache event synchronization system that enables real-time cache state sharing across vLLM instances for improved prefix caching efficiency.

Key features:
- ZMQ-based pub/sub system for low-latency event distribution
- MessagePack encoding for efficient binary serialization
- Thread-safe concurrent hash indexer for cache state management
- Integration with existing prefix cache routing algorithms
- Comprehensive test coverage including unit, integration, and E2E tests
- Performance benchmarks and chaos testing framework
- Migration scripts and deployment configurations

The system supports multiple deployment modes:
- Standalone mode without event sync (backward compatible)
- Event sync mode with configurable ZMQ connections
- Redis-based persistence for cache metadata

The two flags that control new KV event synchronization feature are:

  1. `AIBRIX_USE_REMOTE_TOKENIZER=true` - Must be enabled first
  2. `AIBRIX_KV_EVENT_SYNC_ENABLED=true` - Depends on remote tokenizer being enabled

  These flags have a dependency relationship where the remote tokenizer must be enabled before the KV event sync can be activated. This ensures tokenization consistency between the gateway and vLLM engines, which is critical for accurate hash computation and cache state synchronization.

If both flags are disabled (default state), AIBrix behavior remains completely identical to previous versions. The system operates in backward-compatible mode with no changes to existing functionality - prefix cache routing uses local tokenization, cache state management follows the original implementation, and there is no event synchronization between the gateway and vLLM engines. This ensures seamless upgrades and maintains full compatibility with existing deployments.

## Related Issues
Resolves: #942


helm/aibrix-model/Chart.yaml
```diff
@@ -0,0 +1,18 @@
+---
```
Jeffwan: let's remove `helm/aibrix-model` from this PR. we can file separate PR for this change

build/container/Dockerfile
```diff
@@ -4,6 +4,13 @@ ARG TARGETOS
 ARG TARGETARCH
 
 WORKDIR /workspace
+
+# Install ZMQ dependencies
```
Jeffwan: controller-manager dockerfile doesn't need to be changed as I know?

Jeffwan: zmq is only used in gateway-plugin, right?

build/container/Dockerfile.gateway
```diff
@@ -16,18 +23,27 @@ COPY cmd/ cmd/
...
+    ca-certificates \
+    && rm -rf /var/lib/apt/lists/*
 
-# Use distroless as minimal base image to package the manager binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
 WORKDIR /
 COPY --from=builder /workspace/gateway-plugins .
+
+# Create non-root user
```
Jeffwan: this is feature unrelated change? we should cut separate PR if needed

build/container/Dockerfile.metadata
```diff
@@ -16,18 +23,27 @@ COPY cmd/ cmd/
...
+# Build with CGO enabled for ZMQ support
 # the GOARCH has not a default value to allow the binary be built according to the host where the command
 # was called. For example, if we call make docker-build in a local env which has the Apple Silicon M1 SO
 # the docker BUILDPLATFORM arg will be linux/arm64 when for Apple x86 it will be linux/amd64. Therefore,
 # by leaving it empty we can ensure that the container and binary shipped on it will have the same platform.
-RUN CGO_ENABLED=0 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -a -o metadata-service cmd/metadata/main.go
+RUN CGO_ENABLED=1 GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} go build -tags zmq -a -o metadata-service cmd/metadata/main.go
+
+# Use debian-slim as base image since we need ZMQ libraries at runtime
+FROM debian:bookworm-slim
```
Jeffwan: I see you already install libzmq, why do you still need `debian:bookworm-slim`?

docs/kv-cache-events-guide.md
```diff
@@ -0,0 +1,79 @@
+# Enabling KV Cache Event Synchronization
```
Jeffwan: Can you check in the code to docs/source? 

https://aibrix.readthedocs.io/latest/ here is the doc site

docs/testing/e2e-test-guide.md
```diff
@@ -0,0 +1,296 @@
+# End-to-End Testing Guide for KV Event Synchronization
```
Jeffwan: Great work on the documentation

migration/enable-kv-events.sh
```diff
@@ -0,0 +1,79 @@
+#!/bin/bash
```
Jeffwan: move to `hack` folder.

pkg/cache/cache_init.go
```diff
@@ -381,3 +396,45 @@ func initTraceCache(redisClient *redis.Client, stopCh <-chan struct{}) {
...
+	remoteTokenizerEnabled, _ := strconv.ParseBool(remoteTokenValue)
+
+	// Create sync indexer only if KV sync is properly configured
+	if kvSyncEnabled && remoteTokenizerEnabled {
+		s.syncPrefixIndexer = syncindexer.NewSyncPrefixHashTable()
+
+		// Enable KV event sync
+		s.kvEventManager = NewKVEventManager(s)
+		if err := s.kvEventManager.Start(); err != nil {
+			return fmt.Errorf("failed to start KV event sync: %w", err)
```
DwyaneShi: need to reclaim s.syncPrefixIndexer and s.kvEventManager if failed to start

pkg/cache/kvcache/zmq_client.go
```diff
@@ -0,0 +1,475 @@
...
+	}
+
+	// Create DEALER socket for replay (to communicate with ROUTER)
+	replaySocket, err := zmq.NewSocket(zmq.DEALER)
+	if err != nil {
+		_ = subSocket.Close()
+		return fmt.Errorf("failed to create DEALER socket: %w", err)
+	}
+
+	replayEndpoint := fmt.Sprintf("tcp://%s:%d", c.config.PodIP, c.config.RouterPort)
```
DwyaneShi: better to leave a comment if IPv6 is not supported yet

pkg/cache/cache_init.go
ae86zhizhi: Thank you spotting that the previous implementation could fail to clean up resources if `kvEventManager.Start()` ran into an issue.

I've pushed a fix in commit `45b72c5` that addresses this. Here’s a quick rundown of the changes:

*   **Proactive Validation:** I've added a new `validateConfiguration()` method. This now checks if the manager can start successfully *before* we allocate the sync indexer, preventing the problem from happening in the first place.

*   **Guaranteed Cleanup with `defer`:** To make the cleanup process more robust, I've used a `defer` pattern. This ensures that even if initialization fails, our cleanup logic will always run.
    ```go
    var initialized bool
    defer func() {
        if !initialized {
            s.cleanupKVEventSync()
        }
    }()
    ```

*   **Centralized Cleanup Logic:** The `defer` now calls a single `cleanupKVEventSync()` method. This new function is responsible for properly cleaning up both the event manager and the sync indexer, keeping the logic tidy.

*   **Idempotent `Stop()` Method:** I've updated `KVEventManager.Stop()` to be idempotent. It now tracks if it has already been called and will simply return on any subsequent calls, which prevents errors from multiple cleanup attempts.

*   **Comprehensive Tests:** I added new unit tests to confirm that resources are properly released after a failure and to verify that the cleanup methods can be called multiple times safely.

The fix should ensure no resources are leaked, even if the initialization process fails at any point. All existing tests continue to pass, and the new tests specifically validate this new cleanup behavior.

helm/aibrix-model/Chart.yaml
ae86zhizhi: I've removed the `helm/aibrix-model/` directory from this PR (commit a62681b). The Helm chart will be submitted as a separate PR after this KV event sync feature is merged. This keeps the current PR focused solely on the core implementation.

migration/enable-kv-events.sh
ae86zhizhi: Thank you for the suggestion regarding the migration script location. I've addressed this in commit 6953f7c:
**Changes made:**
- Moved `enable-kv-events.sh` from `migration/` to `hack/upgrade/` following Kubernetes/Go community conventions
- Created `hack/upgrade/` subdirectory specifically for upgrade-related scripts
- Added README.md documenting the upgrade scripts directory and guidelines for future scripts

build/container/Dockerfile
ae86zhizhi: Addressed in commit d0f188f. You were correct - controller-manager doesn't need ZMQ. I've implemented a differential build strategy:

**Components without KV sync** (controller-manager, metadata-service):
- Built with `CGO_ENABLED=0` and `-tags="nozmq"`
- Use `gcr.io/distroless/static:nonroot`
- Result: Smaller, more secure images

**Components with KV sync** (gateway-plugins):
- Built with `CGO_ENABLED=1` and `-tags="zmq"`
- Use `debian:bookworm-slim` with libzmq5
- Result: Full KV event sync functionality

build/container/Dockerfile.metadata
ae86zhizhi: After analyzing the code, metadata-service doesn't actually use ZMQ at all - it was incorrectly configured with these dependencies.

**Fixed in commit 42ed6a8:**
- Removed all ZMQ dependencies (libzmq3-dev, libzmq5)
- Changed from `CGO_ENABLED=1` to `CGO_ENABLED=0`
- Switched from `debian:bookworm-slim` to `gcr.io/distroless/static:nonroot`
- Build with `-tags="nozmq"` instead of `-tags="zmq"`

To answer your original question: The libzmq3-dev in the builder was for compilation, while debian-slim was needed for the runtime library (libzmq5). But since metadata-service doesn't use ZMQ functionality, both were unnecessary.

docs/kv-cache-events-guide.md
ae86zhizhi: Fixed in commit 4afc4ed! I've migrated all KV event sync documentation to the proper Sphinx structure:

- `docs/kv-cache-events-guide.md` → `docs/source/features/kv-event-sync.rst`
- `docs/testing/e2e-test-guide.md` → `docs/source/testing/kv-event-sync-e2e.rst`

docs/testing/e2e-test-guide.md
ae86zhizhi: I've preserved all the detailed testing scenarios while converting to RST format. The guide is now at `docs/source/testing/kv-event-sync-e2e.rst`.

pkg/cache/kvcache/zmq_client.go
ae86zhizhi: I've implemented full IPv6 support in commit 76d36f5. 
**Implementation details:**
- Created `formatZMQTCPEndpoint()` helper that uses Go's `net.JoinHostPort()` to automatically handle IPv6 bracket notation
- Added IP address validation in `ValidateConfig()`
- Comprehensive test coverage for IPv4, IPv6, and hostname scenarios

The code now properly supports:
- IPv4: `tcp://192.168.1.1:5557`
- IPv6: `tcp://[2001:db8::1]:5557`
- IPv6 with zone: `tcp://[fe80::1%eth0]:5557`
- Hostnames: `tcp://localhost:5557

ae86zhizhi: I've implemented full IPv6 support in commit 76d36f5. 
**Implementation details:**
- Created `formatZMQTCPEndpoint()` helper that uses Go's `net.JoinHostPort()` to automatically handle IPv6 bracket notation
- Added IP address validation in `ValidateConfig()`
- Comprehensive test coverage for IPv4, IPv6, and hostname scenarios

The code now properly supports:
- IPv4: `tcp://192.168.1.1:5557`
- IPv6: `tcp://[2001:db8::1]:5557`
- IPv6 with zone: `tcp://[fe80::1%eth0]:5557`
- Hostnames: `tcp://localhost:5557

pkg/cache/cache_init.go
ae86zhizhi: Thank you spotting that the previous implementation could fail to clean up resources if `kvEventManager.Start()` ran into an issue.

I've pushed a fix in commit 45b72c5 that addresses this. Here’s a quick rundown of the changes:

*   **Proactive Validation:** I've added a new `validateConfiguration()` method. This now checks if the manager can start successfully *before* we allocate the sync indexer, preventing the problem from happening in the first place.

*   **Guaranteed Cleanup with `defer`:** To make the cleanup process more robust, I've used a `defer` pattern. This ensures that even if initialization fails, our cleanup logic will always run.
    ```go
    var initialized bool
    defer func() {
        if !initialized {
            s.cleanupKVEventSync()
        }
    }()
    ```

*   **Centralized Cleanup Logic:** The `defer` now calls a single `cleanupKVEventSync()` method. This new function is responsible for properly cleaning up both the event manager and the sync indexer, keeping the logic tidy.

*   **Idempotent `Stop()` Method:** I've updated `KVEventManager.Stop()` to be idempotent. It now tracks if it has already been called and will simply return on any subsequent calls, which prevents errors from multiple cleanup attempts.

*   **Comprehensive Tests:** I added new unit tests to confirm that resources are properly released after a failure and to verify that the cleanup methods can be called multiple times safely.

The fix should ensure no resources are leaked, even if the initialization process fails at any point. All existing tests continue to pass, and the new tests specifically validate this new cleanup behavior.

pkg/cache/kvcache/zmq_client.go
DwyaneShi: besides handling ipv6 format, explictly setting ipv6 flag (i.e., sock.SetIpv6(true)) is also required to avoid potential issues on dual-stack machines

.github/workflows/installation-tests.yml
```diff
@@ -3,7 +3,7 @@ name: Installation Tests
 on:
   workflow_dispatch:  # Allows manual trigger
   push:
-    branches: [ "main", "release-*" ]
+    branches: [ "main", "release-*", "feature/kv-event-pub-sub" ]
```
Jeffwan: remove all `feature/kv-event-pub-sub` in the github actions. 

I think you added them to reuse the testing workflow, we should remove them in upstream

.github/workflows/complete-testing.yml
```diff
@@ -0,0 +1,410 @@
+name: Complete Testing Pipeline
```
Jeffwan: any summary on the difference between `installation-test.yaml`?

.github/workflows/lint-and-tests.yml
```diff
@@ -37,6 +37,8 @@ jobs:
         uses: actions/setup-go@v4
         with:
           go-version: '1.22'
+      - name: Install ZMQ dependencies
```
Jeffwan: is this helpful? does lint and unit test rely on zero now?

.github/workflows/nightly-performance.yml
```diff
@@ -0,0 +1,287 @@
+name: Nightly Performance Tests
```
Jeffwan: let's keep this one as a separate PR. please back up the files and let's talk about this separately.remove this in current PR

.github/workflows/python-aibrix-kvcache-tests.yml
```diff
@@ -2,7 +2,7 @@ name: Python AIBrix KVCache Tests
 
 on:
   push:
-    branches: [ "main", "release-*" ]
+    branches: [ "main", "release-*", "feature/kv-event-pub-sub" ]
```
Jeffwan: same here. remove the branch

CLAUDE.md
```diff
@@ -0,0 +1,173 @@
+# CLAUDE.md
```
Jeffwan: I am thinking how to embrace such tools. does Claude requires the root level `CLAUDE.md`?

Makefile
```diff
@@ -142,6 +177,18 @@ lint-all: licensecheck lint
 build: manifests generate fmt vet ## Build manager binary.
 	go build -o bin/manager cmd/controllers/main.go
 
+.PHONY: build-controller-manager
```
Jeffwan, line 145: can we make `-tags="nozmq"` as default option?  in this case, we do not need to make anychanges to existing code

build/container/Dockerfile
Jeffwan: see above https://github.com/vllm-project/aibrix/pull/1349#discussion_r2243547052. it's better not to change the default code and behavior. only enable zmq build for router related components

CLAUDE.md
ae86zhizhi: It is a mistake. Now filtered.

pkg/cache/kvcache/zmq_client.go
ae86zhizhi: I've updated the implementation in 04e3815 to enable IPv6 on all ZMQ sockets unconditionally.

The approach follows modern best practices:
- All sockets now call SetIpv6(true) after creation for dual-stack support
- Works seamlessly with IPv4, IPv6, and hostnames
- Tests updated to use :: for wildcard binding

This ensures proper behavior in all network environments without requiring conditional logic.

pkg/utils/redis.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 func GetRedisClient() *redis.Client {
-	redisHost := LoadEnv("REDIS_HOST", "localhost")
+	redisHost := LoadEnv("REDIS_HOST", "127.0.0.1")
```
autopear: 这个改动的目的是什么？

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -43,31 +50,114 @@ var (
...
+			Help: "Status of prefix cache indexer (1=available, 0=unavailable)",
+		},
+		[]string{"model", "indexer_type"},
+	)
+
+	prefixCacheRoutingLatency = promauto.NewHistogramVec(
+		prometheus.HistogramOpts{
+			Name:    "aibrix_prefix_cache_routing_latency_seconds",
+			Help:    "Latency of prefix cache routing decisions",
+			Buckets: prometheus.ExponentialBuckets(0.00001, 2, 15), // 10μs to ~160ms
```
autopear: Do not include any unicode characters in the source file like `μ`

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -43,31 +50,114 @@ var (
...
-	if tokenizerType == "tiktoken" {
-		tokenizerObj = tokenizer.NewTiktokenTokenizer()
-	} else {
-		// Default to character tokenizer for backward compatibility
-		if tokenizerType != "character" {
-			klog.InfoS("unrecognized tokenizer type, defaulting to character", "type", tokenizerType)
+	var remoteTokenizerObj tokenizer.Tokenizer
+
+	// Check configuration dependencies
+	remoteTokenizerValue := utils.LoadEnv("AIBRIX_USE_REMOTE_TOKENIZER", "false")
```
autopear: You already have `LoadEnvBool`

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -43,31 +50,114 @@ var (
...
+	useRemoteTokenizer, _ := strconv.ParseBool(remoteTokenizerValue)
+	kvSyncValue := utils.LoadEnv("AIBRIX_KV_EVENT_SYNC_ENABLED", "false")
+	kvSyncEnabled, _ := strconv.ParseBool(kvSyncValue)
```
autopear: LoadEnvBool

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -224,3 +524,61 @@ func getTargetPodOnLoadImbalance(cache cache.Cache, readyPods []*v1.Pod) (*v1.Po
...
+	switch {
+	case matchPercent == 0:
+		bucket = "0"
+	case matchPercent <= 25:
+		bucket = "1-25"
+	case matchPercent <= 50:
+		bucket = "26-50"
+	case matchPercent <= 75:
+		bucket = "51-75"
+	case matchPercent <= 99:
```
autopear: Why not just 76-100 instead of 76-99 + 100?

pkg/utils/syncprefixcacheindexer/sync_hash.go
```diff
@@ -0,0 +1,718 @@
...
+
+	// Metadata
+	createTime    time.Time
+	lastAccess    atomic.Int64 // Unix timestamp (lock-free update)
+	totalPrefixes int64
+}
+
+// EngineHashMapping maintains unidirectional mapping
+type EngineHashMapping struct {
+	engineToAibrix map[int64]uint64 // engine block hash → aibrix prefix hash
```
autopear, line 75: is engine block hash int64 or uint64?

.github/workflows/python-aibrix-kvcache-tests.yml
autopear: removed

.github/workflows/nightly-performance.yml
autopear: removed

.github/workflows/lint-and-tests.yml
autopear: removed

.github/workflows/complete-testing.yml
autopear: removed

.github/workflows/installation-tests.yml
autopear: removed

Jeffwan (approved):
lgtm.

PR #1372, opened 2025-08-01
Link to system-installed zmq without building

autopear:
Do not build zmq locally, link to system-install zmq and its dependencies.


build/container/Dockerfile.gateway
```diff
@@ -26,21 +20,25 @@ COPY cmd/ cmd/
...
 
 RUN GOOS=${TARGETOS:-linux} GOARCH=${TARGETARCH} \
     go build -tags zmq \
-    -ldflags '-extldflags "-static"' \
+    -ldflags "-extldflags '${CGO_LDFLAGS}'" \
     -a -o gateway-plugins cmd/plugins/main.go
 
+# Gather shared libs
+RUN mkdir -p /workspace/deps
+RUN ldd /workspace/gateway-plugins | awk '{if ($3 ~ /^\//) print $3}' | xargs -I '{}' cp -v '{}' /workspace/deps/ || true
```
DwyaneShi, line 33: why do we prefer using copy rather than installing those dependencies in the distroledss image section?

autopear: The distroless image has no package manager like apt/yum, so it's very hard to install system dependencies.

DwyaneShi: I see, but this way will pollute the distroless image, we should avoid this. Please take a look at https://github.com/GoogleContainerTools/distroless/issues/863#issuecomment-986062361 and https://dev.to/sineaggi/python-and-distroless-a-simplified-guide-13je, this is a better solution relying on `apt download` and dpkg. The second post has more details (and some typos though).

autopear: @DwyaneShi I don't think this is a good method and it is essentially the same as what I did.

A deb package is just a tarball fill containing 2 tarballs, one is control.tar.gz, the other is data.tar.* where * depends on the compression algorithm like gzip, bz2, lzma, etc.

If you check the command the 2nd links uses:
```shell
        dpkg --ctrl-tarfile $deb | tar -Oxf - ./control > /dpkg/var/lib/dpkg/status.d/${package_name}; \
        dpkg --extract $deb /dpkg || exit 10; \
```
it extracts the control file (which defines the deb's content like package name, version, dependencies, etc.) into status.d, and the actual package content (like lib and include files) into /dpkg.

And later, to generate the image, it `COPY --from=build/dpkg /` which copies the whole deb contents to the image's / dir. This simulates the installing of deb packages via dpkg commands.

However, for this PR, what we need is just some .so libs which is used in the runtime with `LD_LIBRARY_PATH`, and we do not need the other files like include files, configs, etc. IMO, the method in this PR is even cleaner than the 2 links you provided.

DwyaneShi: I'm concerned that copying libs directly could overwrite common system libs like libc.so libstdc++.so etc. This is risky because it might break abi compatibility, cause some unknown issues and also complicate future maintenance. Using `apt download` might be able to avoid the overwrite issue from my understanding. And of course we can explicitly exclude those system libs from copying as well. I am good with any approach that can resolve the overwrite issue.

autopear: The current version copies everything except linux-vdso.so.1 and ld-linux-x86-64.so.2 to a non-standard location /gateway-plugins-lib. It does not overwrite any system lib files.

Also, I believe that at least libstdc++.so is needed as I remember this is missing in distroless image. This file is provided in package, which depends on gcc.

DwyaneShi: I wasn't aware of your recent changes, LGTM then.

PR #1431, opened 2025-08-10
Unified minimum value selection code

jiangxiaobin96:
Unify code for selecting minimum value code which is the same with least_util and least_gpu_cache and avoid two for loop.


Jeffwan (approved):
seem this is not reviewed yet. the change is very straightforward and it looks good to me.

Jeffwan:
@jiangxiaobin96 thanks for the improvement and apologies for the delay in reviewing. I changed the PR title a little bit to make it more self-explainable.

PR #1463, opened 2025-08-17
add least_request_test

jiangxiaobin96:
## Pull Request Description
add least_request_test.go

### DIFF
--- a/pkg/plugins/gateway/algorithms/least_request_test.go
+++ b/pkg/plugins/gateway/algorithms/least_request_test.go
@@ -1,0 +1,134 @@
+/*
+Copyright 2024 The Aibrix Team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package routingalgorithms
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/vllm-project/aibrix/pkg/cache"
+	"github.com/vllm-project/aibrix/pkg/metrics"
+	"github.com/vllm-project/aibrix/pkg/types"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestLeastRequest(t *testing.T) {
+	tests := []struct {
+		name       string
+		readyPods  []*v1.Pod
+		podMetrics map[string]map[string]metrics.MetricValue
+		expectErr  bool
+		expectMsgs []string
+	}{
+		{
+			name: "successful routing with least request",
+			readyPods: []*v1.Pod{
+				{ObjectMeta: metav1.ObjectMeta{Name: "p1"}, Status: v1.PodStatus{PodIP: "1.1.1.1",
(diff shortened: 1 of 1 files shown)


pkg/plugins/gateway/algorithms/least_request_test.go
```diff
@@ -0,0 +1,68 @@
...
+			targetPod, err := leastRequestRouter.Route(ctx, podList)
+			if tt.expectErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Equal(t, tt.expectMsg, targetPod)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
googs1025: I think these cases are reasonable and necessary. Can you help add them? 😄

Jeffwan:
@googs1025  I will leave the PR review to you.  Feel free to merge it after all feedback is addressed. BTW, I will merge this one first. https://github.com/vllm-project/aibrix/pull/1431

pkg/plugins/gateway/algorithms/least_request_test.go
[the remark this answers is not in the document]
jiangxiaobin96: done

jiangxiaobin96:
How to solve this UT fail? I run all `go test` in pkg/plugins/gateway/algorithms/ and all pass.

googs1025:
Yes, this doesn't seem to be introduced by this pull request. We've encountered a data race. We'll check it out sometime. 
cc @Jeffwan 

FYI: 
https://github.com/vllm-project/aibrix/actions/runs/17043154923/job/48311887540

pkg/plugins/gateway/algorithms/least_request_test.go
```diff
@@ -0,0 +1,134 @@
...
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+func TestLeastRequest(t *testing.T) {
+	tests := []struct {
+		name       string
+		readyPods  []*v1.Pod
+		podMetrics map[string]map[string]metrics.MetricValue
+		expectErr  bool
```
googs1025, line 36: Can expectErr be a detectable err, not just a bool value?

pkg/plugins/gateway/algorithms/least_request_test.go
```diff
@@ -0,0 +1,134 @@
...
+				cache: c,
+			}
+
+			ctx := types.NewRoutingContext(context.Background(), "test", "m1", "message", "request", "user")
+			targetPod, err := leastRequestRouter.Route(ctx, podList)
+			if tt.expectErr {
+				assert.Error(t, err)
+			} else {
+				assert.NoError(t, err)
+				assert.Contains(t, tt.expectMsgs, targetPod)
```
googs1025, line 130: When multiple requests have the same request count, will a random one be retrieved? That's why assert.Contains is used here. 🤔

pkg/plugins/gateway/algorithms/least_request_test.go
jiangxiaobin96: expectErr refers to `pd_disaggregation_test.go` which represent if error exist. Here the error is `no ready pods available for random selection`, need to confirm if need to present detectable error.

pkg/plugins/gateway/algorithms/least_request_test.go
jiangxiaobin96: Yes, in source code when request count is the same, that means we find more than one targetPod,  we will select a random one.
https://github.com/vllm-project/aibrix/blob/528fd8076a31f969c433e7aa20f7c04e4e67d421/pkg/plugins/gateway/algorithms/least_request.go#L93

pkg/plugins/gateway/algorithms/least_request_test.go
googs1025: sounds good

googs1025 (approved):

googs1025:
LGTM. thanks for this help ! 😄

PR #1508, opened 2025-08-26
[Misc] Add autoscaling validation of minReplicas and maxReplicas

jiangxiaobin96:
## Pull Request Description
Add validation of minReplicas and maxReplicas in HPA autoscaling

### DIFF
--- a/pkg/controller/podautoscaler/hpa_resources.go
+++ b/pkg/controller/podautoscaler/hpa_resources.go
@@ -38,10 +38,19 @@ var (
 // MakeHPA creates an HPA resource from a PodAutoscaler resource.
 func makeHPA(pa *pav1.PodAutoscaler) (*autoscalingv2.HorizontalPodAutoscaler, error) {
 	minReplicas, maxReplicas := pa.Spec.MinReplicas, pa.Spec.MaxReplicas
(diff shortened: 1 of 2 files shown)


pkg/controller/podautoscaler/hpa_resources.go
```diff
@@ -59,11 +68,10 @@ func makeHPA(pa *pav1.PodAutoscaler) (*autoscalingv2.HorizontalPodAutoscaler, er
 				Name:       pa.Spec.ScaleTargetRef.Name,
 			},
 			MaxReplicas: maxReplicas,
+			MinReplicas: minReplicas,
```
googs1025: I'm a bit curious that MinReplicas seems to be optional rather than required in HPA.

jiangxiaobin96: I get it! That's why `hpa.Spec.MinReplicas = minReplicas` is set after `hpa := &autoscalingv2.HorizontalPodAutoscaler`

googs1025 (approved):
I think this check is sufficient at this stage. If there are more complex validations or other validations in the future, we can use validate webhook to achieve

jiangxiaobin96:
cc @Jeffwan

PR #1608, opened 2025-09-27
Use EnqueueRequestsFromMapFunc for KVCache controller

baozixiaoxixi:
## Pull Request Description
The current scenario requires mapping Pod events to their associated KVCache resources

### DIFF
--- a/pkg/controller/kvcache/kvcache_controller.go
+++ b/pkg/controller/kvcache/kvcache_controller.go
@@ -103,7 +103,23 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
 		For(&orchestrationv1alpha1.KVCache{}).
 		Owns(&corev1.Service{}).
 		Owns(&appsv1.Deployment{}).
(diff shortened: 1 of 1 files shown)


pkg/controller/kvcache/kvcache_controller.go
```diff
@@ -103,7 +104,23 @@ func add(mgr manager.Manager, r reconcile.Reconciler) error {
...
+				cacheName, ok := obj.GetLabels()[constants.KVCacheLabelKeyIdentifier]
+				if !ok {
+					return nil
+				}
+
+				return []ctrl.Request{
+					{
+						NamespacedName: types.NamespacedName{
+							Namespace: obj.GetNamespace(),
+							Name:      cacheName,
```
googs1025, line 118: good catch. This seems correct,  kvcache controller we triggered needs to use the cacheName not pod name

googs1025:
Leaving final ack for @Jeffwan

Jeffwan:
LGTM as well

Jeffwan (approved):

PR #1624, opened 2025-10-02
feat: support metric label

baozixiaoxixi:
## Pull Request Description

I noticed that the multi-engine support for metrics has been consolidated into the updatePodMetricsFromTypedResult method, but label extraction is not yet supported. In gateway-plugins service of my env, I encountered some runtime errors in the logs related to failing to extract labels such as vllm:running_lora_adapters. 

I made some changes to the label metrics to address this.


Jeffwan:
@baozixiaoxixi I think the refactor makes sense, thanks for the clean up the deprecated code.

Jeffwan (approved):

Jeffwan:
@baozixiaoxixi could you sign off the commit to pass the DCO check?

baozixiaoxixi:
> @baozixiaoxixi could you sign off the commit to pass the DCO check?

Done~

PR #1631, opened 2025-10-06
Refining workload generator documentation

happyandslow:
## Pull Request Description
This PR refines benchmark documents / README


benchmarks/README.md
```diff
@@ -1,4 +1,24 @@
...
+5. [Workload Generation](#workload-generation)  
+6. [Dispatch Workload Using Client](#dispatch-workload-using-client)  
+7. [Benchmark Analysis](#benchmark-analysis)  
+
+
+---
+
+## Benchmark Overview
 
 AIBrix Benchmark contains the following components:
```
dczhu, line 23: Not specific to this PR but about README.md itself -- Since a "Benchmark Scenario" is related to an AIBrix component or a feature, it would include at least two parts: (a) a special workload that should be part of the "Workload Generation" step, e.g. an "Autoscaling" scenario may require traffic wave simulation, (b) a set of scripts to set up the benchmark environment for the scenario. Is it a correct understanding?

benchmarks/README.md
```diff
@@ -22,20 +42,29 @@ kubectl -n envoy-gateway-system port-forward ${service_name} 8888:80 &
 export API_KEY="${your_api_key}"
 ```
 
-## Run benchmark end-to-end
-To run all steps using the default setting, try
+## Using This Benchmark
```
dczhu, line 45: Some of the info below could be found by digging into the code, some not. It would be nice to give brief intro to the key parameters in config.yaml, like:


* synthetic_shared -> prompt_length: the meanings of the 3 fields separated by comma, how they are used.
* sharegpt -> target_dataset: How to obtain files like "/tmp/ShareGPT_V3_unfiltered_cleaned_split.json"
* client_trace -> trace: Where to obtain a client trace, how it looks like

benchmarks/README.md
```diff
@@ -77,7 +106,78 @@ The first two types generate synthetic prompts and the second two types convert
 
 For details of the dataset generator, check out [README](./generator/dataset-generator/README.md). All tunable parameters are set under ```config/dataset```.
```
dczhu: README.md file missing.

benchmarks/README.md
```diff
@@ -90,7 +190,7 @@ Below are the workload types that are currently being supported. The ```workload
     - Synthetic fluctuation load (**synthetic**): The loads (QPS/input length/output length) fluctuate based on configurable parameters.
```
dczhu, line 216: The constant load has stable load, what does "controllable fluctuation" mean in this case?

benchmarks/README.md
```diff
@@ -128,7 +228,104 @@ The workload generator will produce a workload file that looks like the followin
 
 Details of the workload generator can be found [here](generator/workload-generator/README.md).
```
dczhu: README.md not found

benchmarks/README.md
```diff
@@ -137,13 +334,66 @@ The benchmark client supports both batch and streaming modes. Streaming mode sup
...
+client_output: ...
+endpoint: ...
+```
+
+- **workload_file**: `str` — File path to the workload file. Default: `None`. Default: `./output/workload/${workload_type}/workload.jsonl`.
+- **client_output**: `str` - Path to the output log file produced by client. Default: `./output/client_output`.
+- **endpoint**: `str` — Endpoint URL for the target service. Default: `"http://localhost:8888"`.
+- **api_key**: `str` — API key to the service. Set through environment variable: `${API_KEY}`.
+- **target_model**: `str` — Default target model (used if workload does not contain a target model). Default: `None`.
+- **time_scale**: `float` — Scaling factor for workload’s logical time. Default: `1.0`.
```
dczhu: What does "scaling factor for workload's logical time" mean? What does it affect?

benchmarks/README.md
```diff
@@ -137,13 +334,66 @@ The benchmark client supports both batch and streaming modes. Streaming mode sup
...
+
+```yaml
+# ---------------
+# OPTIONAL: ANALYSIS
+# ---------------
+trace_output: ...
+goodput_target: ...
+```
+
+- **trace_output**: `str` — Directory that stores the analysis output. Default: `./output/trace_analysis`.
```
dczhu: How to use the trace analysis?

benchmarks/README.md
happyandslow: The scenario: https://github.com/vllm-project/aibrix/tree/main/benchmarks/scenarios directory is meant for this. I actually also have an open PR for the autoscaling scenario: https://github.com/vllm-project/aibrix/pull/862. These scenarios are not actively maintained but I think they should be revived.

benchmarks/README.md
happyandslow: Could you check out "Configuring Dataset Generation" section and see if this has addressed your comment? Some of the parameters we select (like prompt_length) was based on our own data and I'm hesitating whether we should claim that the data is derived from internal source.

happyandslow: Added some contents to address two other comments.

benchmarks/README.md
happyandslow: Fixed link.

benchmarks/README.md
happyandslow: "sampled explential distribution interval fluctuation". Updated.

benchmarks/README.md
happyandslow: Fixed.

benchmarks/README.md
happyandslow: Added.

benchmarks/README.md
happyandslow: Added.

benchmarks/README.md
```diff
@@ -1,4 +1,25 @@
...
+
+1. [Bencmark Overview](#benchmark-overview)  
+1. [Preliminary](#preliminary)  
+1. [Using This Benchmark](#using-this-benchmark)  
+1. [Dataset Generation](#dataset-generation)  
+2. [Workload Generation](#workload-generation)  
+3. [Dispatch Workload Using Client](#dispatch-workload-using-client)  
+4. [Analyze Section](#analyze-section)  
+5. [Notes & Usage](#notes--usage)  
+6. [Example Config](#example-config)
```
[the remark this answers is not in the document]
happyandslow: Fixed.

benchmarks/README.md
dczhu: > Could you check out "Configuring Dataset Generation" section and see if this has addressed your comment? Some of the parameters we select (like prompt_length) was based on our own data and I'm hesitating whether we should claim that the data is derived from internal source.

[dczhu] Yes, I did read that section. What I don't understand is why we need multiple configs (3 in this case), and in what way these configs work together (or how they are used, not how the numbers were derived).

benchmarks/README.md
dczhu: > "sampled explential distribution interval fluctuation". Updated.

[dczhu] Exponential?

Issue #1803, opened 2025-11-26
Add external-filter in Header for advanced routing


rayne-Li:
### 🚀 Feature Description and Motivation

AFAIK currently aibrix-gateway-plugins use pod.Label `model.aibrix.ai/name` filtering pods in cluster, I want to filter pods besides that, by using external filter in header or somewhere

### Use Case

my cloud-edge openyurt k8s cluster need filter pods by nodepools, which i usually label pods like: node-pool:a or other topo-types labels i need
by adding external-filters in headers, i can filter pod in exact area or for some tenant

### Proposed Solution

1. add external-filter(k8s label-selector type string) in selectTargetPod in gateway.go
```go
// gateway.go
func (s *Server) selectTargetPod(ctx *types.RoutingContext, pods types.PodList, externalFilterExpr string) (string, error) {
	// ......

	// filter pod by header ext-filter
	// k8s labelSelector format, eg: "k=v"、"env in (prod,stg)"
	if externalFilterExpr != "" {
		sel, err := labels.Parse(externalFilterExpr)
		if err != nil {
			return "", err
		}
		out := make([]*corev1.Pod, 0, len(readyPods))
		for _, p := range readyPods {
			klog.V(3).InfoS("filtering pod", "pod", p.Name)
			if sel.Matches(labels.Set(p.Labels)) {
				out = append(out, p)
				klog.V(3).InfoS("filter passed", "pod", p.Name)
			}
		}
		readyPods = out
	}

	// ......

	return router.Route(ctx, &utils.PodArray{Pods: readyPods})
}
```
2. get filter from header
```go
// pkg/plugins/gateway/gateway_req_body.go line:78
		// external filter in header
		externalFilter := routingCtx.ReqHeaders[HeaderExternalFilter]
		klog.InfoS("found external filter", "filter", externalFilter)
		targetPodIP, err := s.selectTargetPod(routingCtx, podsArr, externalFilter)
```
3. handle that header
```go
// pkg/plugins/gateway/gateway_req_headers.go
func (s *Server) HandleRequestHeaders(ctx context.Context, requestID string, req *extProcPb.ProcessingRequest) (*extProcPb.ProcessingResponse, utils.User, int64, *types.RoutingContext) {
         // ......
	for _, n := range h.RequestHeaders.Headers.Headers {
		if strings.ToLower(n.Key) == "user" {
			username = string(n.RawValue)
		}
		if strings.ToLower(n.Key) == ":path" {
[...]

PR #1804, opened 2025-11-26
[Feature]: Add external-filter in Header for advanced routing

rayne-Li:
## Pull Request Description

Add external-filter in Header for advanced routing

## Related Issues
Resolves: #1803

### DIFF
--- a/pkg/plugins/gateway/gateway_req_body.go
+++ b/pkg/plugins/gateway/gateway_req_body.go
@@ -75,7 +75,6 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 		klog.InfoS("request start", "requestID", requestID, "requestPath", requestPath, "model", model, "stream", stream)
 	} else {
 		externalFilter := routingCtx.ReqHeaders[HeaderExternalFilter]
-		klog.InfoS("found external filter", "filter", externalFilter)
 		targetPodIP, err := s.selectTargetPod(routingCtx, podsArr, externalFilter)
 		if targetPodIP == "" || err != nil {
 			klog.ErrorS(err, "failed to select target pod", "requestID", requestID, "routingStrategy", routingAlgorithm, "model", model, "routingDuration", routingCtx.GetRoutingDelay())
--- a/docs/source/features/gateway-plugins.rst
+++ b/docs/source/features/gateway-plugins.rst
@@ -222,7 +222,7 @@ https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
 
 .. note::
     1. Filtering happens **before** the routing strategy. It never changes which Pods are considered “optimal” by the routing strategy.
-    2. The ``external-filter`` only takes effect when a ``routing-strategy``` is set. If no routing strategy found, the external filter skipped.
+    2. The ``external-filter`` only takes effect when a ``routing-strategy``` is set.
     3. It only reduces the Pod selected by `model.aibrix.ai/name` and set by applying extra label constraints.
     4. Same as `no target pod`, If the filter eliminates all Pods, the request will fail with ``no ready pods for routing``.
     5. ``external-filter`` is optional. When omitted, no extra filtering is applied.


googs1025 (commented):
@rayne-Li Thank you for your proposal, but I'm not entirely sure if this breaks the separation of concerns principle between the gateway and workload. 🤔  Could we achieve this separation at the workload level instead? Or could this be accomplished within the existing model.aibrix.ai/name labeling strategy? Additionally, with this filtering approach, will the target pod we obtain still be optimal?

pkg/plugins/gateway/gateway_req_body.go
```diff
@@ -74,7 +74,10 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 		headers = buildEnvoyProxyHeaders(headers, HeaderModel, model)
 		klog.InfoS("request start", "requestID", requestID, "requestPath", requestPath, "model", model, "stream", stream)
 	} else {
-		targetPodIP, err := s.selectTargetPod(routingCtx, podsArr)
+		// external filter in header
+		externalFilter := routingCtx.ReqHeaders[HeaderExternalFilter]
```
Jeffwan, line 77: could you add brief doc here? a client example would be great! 

you can either add code here or in https://aibrix.readthedocs.io/latest/features/gateway-plugins.html


BTW, external filter support label list, right? something like `label1=value1,label2=value2`

rayne-Li:
> @rayne-Li Thank you for your proposal, but I'm not entirely sure if this breaks the separation of concerns principle between the gateway and workload. 🤔 Could we achieve this separation at the workload level instead? Or could this be accomplished within the existing model.aibrix.ai/name labeling strategy? Additionally, with this filtering approach, will the target pod we obtain still be optimal?

@googs1025 As shown in the code:
```go
func (s *Server) selectTargetPod(ctx *types.RoutingContext, pods types.PodList, externalFilterExpr string) {
    // incoming var: pods are all pod with exact model.name
    // so this won't change whether the pods are the best choices
    // FilterRoutablePods filters ready pods, so I extend that by adding an external filtering function
    readyPods := utils.FilterRoutablePods(pods.All())
    // ......
    for _, p := range readyPods {
        // filtering pod by externalFilter
    }
}
```

The pod filtering happens after utils.FilterRoutablePods (which only filters for ready pods),
so it does not change the optimal pod selected by the routing strategy.

What I want is a common and pluggable filtering method that does not affect any existing routing-strategy logic and can be easily injected or bypassed.
If we rely on another pod label like model.aibrix.ai/name, it may interfere with ModelCache and s.cache.HasModel, making the pod label-select logic more complicated.

pkg/plugins/gateway/gateway_req_body.go
rayne-Li: image

docs and comments are added, and sure label1=value1,label2=value2 are legal by https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/

pkg/plugins/gateway/gateway.go
```diff
@@ -170,6 +172,25 @@ func (s *Server) selectTargetPod(ctx *types.RoutingContext, pods types.PodList)
 		return "", fmt.Errorf("no pods for routing")
 	}
 	readyPods := utils.FilterRoutablePods(pods.All())
+
+	// filter pod by header ext-filter
+	// k8s labelSelector format, eg: "k=v"、"env in (prod,stg)"
+	if externalFilterExpr != "" {
```
googs1025: Could we make this part into a method and write test cases for that method? Because our current test cases don't cover all supported input methods, right?

pkg/plugins/gateway/gateway_req_body.go
```diff
@@ -74,7 +74,22 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 		headers = buildEnvoyProxyHeaders(headers, HeaderModel, model)
 		klog.InfoS("request start", "requestID", requestID, "requestPath", requestPath, "model", model, "stream", stream)
 	} else {
-		targetPodIP, err := s.selectTargetPod(routingCtx, podsArr)
+		// external-filter in header, value is k8s labelSelector format https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/
```
googs1025: we add in docs, so that can we remove this part?

rayne-Li: comments are removed

pkg/plugins/gateway/gateway.go
rayne-Li: add test case for: = != in notin !key/key in selectTargetPod
make a new util func FilterPodsByLabelSelector and also add test cases

googs1025:
> > @rayne-Li Thank you for your proposal, but I'm not entirely sure if this breaks the separation of concerns principle between the gateway and workload. 🤔 Could we achieve this separation at the workload level instead? Or could this be accomplished within the existing model.aibrix.ai/name labeling strategy? Additionally, with this filtering approach, will the target pod we obtain still be optimal?
> 
> @googs1025 As shown in the code:
> 
> ```go
> func (s *Server) selectTargetPod(ctx *types.RoutingContext, pods types.PodList, externalFilterExpr string) {
>     // incoming var: pods are all pod with exact model.name
>     // so this won't change whether the pods are the best choices
>     // FilterRoutablePods filters ready pods, so I extend that by adding an external filtering function
>     readyPods := utils.FilterRoutablePods(pods.All())
>     // ......
>     for _, p := range readyPods {
>         // filtering pod by externalFilter
>     }
> }
> ```
> 
> The pod filtering happens after utils.FilterRoutablePods (which only filters for ready pods), so it does not change the optimal pod selected by the routing strategy.
> 
> What I want is a common and pluggable filtering method that does not affect any existing routing-strategy logic and can be easily injected or bypassed. If we rely on another pod label like model.aibrix.ai/name, it may interfere with ModelCache and s.cache.HasModel, making the pod label-select logic more complicated.

thanks for the detailed explanation. For now, this simple filtering is sufficient for routing. 😄

pkg/plugins/gateway/gateway_req_body.go
```diff
@@ -74,7 +74,10 @@ func (s *Server) HandleRequestBody(ctx context.Context, requestID string, req *e
 		headers = buildEnvoyProxyHeaders(headers, HeaderModel, model)
 		klog.InfoS("request start", "requestID", requestID, "requestPath", requestPath, "model", model, "stream", stream)
 	} else {
-		targetPodIP, err := s.selectTargetPod(routingCtx, podsArr)
+		// external filter in header
+		externalFilter := routingCtx.ReqHeaders[HeaderExternalFilter]
+		klog.InfoS("found external filter", "filter", externalFilter)
```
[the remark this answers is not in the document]
googs1025: @rayne-Li can we take this comment?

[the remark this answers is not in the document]
rayne-Li: sure, it's been removed

googs1025 (approved):
LGTM. thanks for your implement 😄

Issue #1840, opened 2025-12-10
Add more httpRoute path from annotation


rayne-Li:
### 🚀 Feature Description and Motivation

from v0.4.0 to v0.5.0, httpRoute path for model is added from
```go
/v1/completions
/v1/chat/completions
```
to
```go
/v1/completions
/v1/chat/completions
/v1/embeddings
/generate
/generatevideo
```
since vllm has other api path like: /version /pooling /score, in the other hand some small-model not obey OPENAI-API, so a hardcode way to add path won't meet all requirements, so i want to an annotation to specified paths user want himself instead

why i choose annotation not label?

1. labels not support `/`
2. aribrix already control features by annotations like `model.aibrix.ai/sidecar-injection` 

### Use Case

```go
apiVersion: apps/v1
kind: Deployment
metadata:
  labels:
    model.aibrix.ai/name: lzy-qw-18bservice name.
    model.aibrix.ai/port: "8000"
  annotations:
    model.aibrix.ai/model-router-custom-paths: /version,/score # split by , 
```
matches will added in httpRoute like
```yaml
kind: HTTPRoute
spec:
  rules:
  - matches:
# ......
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /version
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /score
```
 

### Proposed Solution

```go
# pkg/controller/modelrouter/modelrouter_controller.go line 287
func (m *ModelRouter) createHTTPRoute(namespace string, labels map[string]string, annotations map[string]string) {
	// append matches if model-router-custom-paths  is set
	if annotations != nil && len(httpRoute.Spec.Rules) > 0 {
		if paths, ok := annotations[modelRouterCustomPathAnnotation]; ok {
			pathSlice := strings.Split(paths, ",")

			for _, path := range pathSlice {
				path = strings.TrimSpace(path)
				if path == "" {
					continue
				}
				httpRoute.Spec.Rules[0].Matches = append(
					httpRoute.Spec.Rules[0].Matches,
					gatewayv1.HTTPRouteMatch{
						Path: &gatewayv1.HTTPPathMatch{
							Type:  ptr.To(gatewayv1.PathMatchPathPrefix),
							Value: ptr.To(path),
						},
						Headers: []gatewayv1.HTTPHeaderMatch{
							modelHeaderMatch,
						},
					},
[...]

PR #1841, opened 2025-12-10
[Feature] add custom path for httpRoute by annotation (#1840)

rayne-Li:
## Pull Request Description
add custom httpRoute paths in annotation

## Related Issues
Resolves: #1840

### DIFF
--- a/docs/source/features/gateway-plugins.rst
+++ b/docs/source/features/gateway-plugins.rst
@@ -89,13 +89,7 @@ On a model or lora adapter deployment, their respective controllers create a HTT
 
 In v0.5.0, the default PathPrefix configuration within HTTPRoute.Spec.Rules.Matches only includes the following endpoints:
 
-* /v1/completions
-* /v1/chat/completions
-* /v1/embeddings
-* /generate
-* /generatevideo
-
-To add custom path prefixes, you must specify them using the model.aibrix.ai/model-router-custom-paths annotation in your deployment's manifest (or in the modelAdapter, RayClusterFleet, LWS).
+To add custom path prefixes, you must specify them using the model.aibrix.ai/model-router-custom-paths annotation in your deployment's manifest (or in the modelAdapter, RayClusterFleet).
 
 .. code-block:: bash
 


pkg/controller/modelrouter/modelrouter_controller.go
```diff
@@ -401,3 +407,45 @@ func (m *ModelRouter) deleteReferenceGrant(namespace string) {
...
+		}
+		if httpRoute.Spec.Rules == nil || len(httpRoute.Spec.Rules) == 0 {
+			httpRoute.Spec.Rules = []gatewayv1.HTTPRouteRule{
+				{
+					Matches: []gatewayv1.HTTPRouteMatch{match},
+				},
+			}
+		} else {
+			httpRoute.Spec.Rules[0].Matches = append(httpRoute.Spec.Rules[0].Matches, match)
+		}
```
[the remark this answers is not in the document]
googs1025: can you check this?

pkg/controller/modelrouter/modelrouter_controller_test.go
```diff
@@ -0,0 +1,133 @@
...
+			name: "no rules in httpRoute",
+			httpRoute: &gatewayv1.HTTPRoute{
+				Spec: gatewayv1.HTTPRouteSpec{
+					Rules: nil,
+				},
+			},
+			annotations: map[string]string{
+				modelRouterCustomPath: "/foo",
+			},
+			wantPaths:    []string{"/foo"}, // rules 为空
```
googs1025: no Chinese comment

pkg/controller/modelrouter/modelrouter_controller_test.go
```diff
@@ -0,0 +1,133 @@
...
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			appendCustomModelRouterPaths(tt.httpRoute, modelHeaderMatch, tt.annotations)
+
+			if tt.skipCheck {
+				return
+			}
+
+			// 没有规则的情况
```
googs1025: ditto

pkg/controller/modelrouter/modelrouter_controller_test.go
```diff
@@ -0,0 +1,133 @@
...
+										Value: ptr.To("/origin"),
+									},
+								},
+							},
+						},
+					},
+				},
+			},
+			annotations: map[string]string{
+				modelRouterCustomPath: "/foo, /bar , ,/baz/",
```
googs1025: this double `,` intentional ?

rayne-Li: yes, check if path added when path is empty

pkg/controller/modelrouter/modelrouter_controller.go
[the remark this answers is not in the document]
rayne-Li: solved in lastest commit

pkg/controller/modelrouter/modelrouter_controller_test.go
googs1025: ok so can you include it in the name? Because this looks like a typo error. 🤔

rayne-Li: solved, add a new test case

googs1025 (commented):
~~LGTM, over all. I don't have other suggestions for this change.~~
if docs needs to be updated, please remember to update it. 😄

rayne-Li:
> LGTM, over all. I don't have other suggestions for this change. if docs needs to be updated, please remember to update it. 😄

doc added

googs1025:
Have we miss the routing configuration on the gateway plugin? 🤔
If we only modify the httproute config, is it feasible for traffic to flow through the gateway to httproute?

rayne-Li:
> Have we miss the routing configuration on the gateway plugin? 🤔 If we only modify the httproute config, is it feasible for traffic to flow through the gateway to httproute?

according to my test, all traffic will routing to aibrix-eg first and decided to be transfer to gateway plugin for routing by model-name in payload(proceed later by envoypatchpolicy in original_destination_cluster) or directly to endpoint of service(according to httpRoute), so changing of httpRoute will not effect routing in gateway-plugin, they are two different way of routing. 

which also means through this way, request will lose ability of advanced routing by gateway-plugin, that also what i want at the begining, if some small model not use OPENAI API , how can they be routed with no model-name in payload? httpRoute may be the only way now?

googs1025:
> > Have we miss the routing configuration on the gateway plugin? 🤔 If we only modify the httproute config, is it feasible for traffic to flow through the gateway to httproute?
> 
> according to my test, all traffic will routing to aibrix-eg first and decided to be transfer to gateway plugin for routing by model-name in payload(proceed later by envoypatchpolicy in original_destination_cluster) or directly to endpoint of service(according to httpRoute), so changing of httpRoute will not effect routing in gateway-plugin, they are two different way of routing.
> 
> which also means through this way, request will lose ability of advanced routing by gateway-plugin, that also what i want at the begining, if some small model not use OPENAI API , how can they be routed with no model-name in payload? httpRoute may be the only way now?


Just want to clarify (I'm not sure if I understand correctly.):  
For now, request traffic (e.g., `/v1/chat/completions`) matches the `reserved-router` HTTPRoute and is **sent to `gateway-plugins`** for model-name-based routing.  

like this: 

```
Client 
  → Envoy Gateway (aibrix-eg) 
  → HTTPRoute: reserved-router (matches /v1/...) 
  → gateway-plugins (filter pod)
  → Inference Pod
```

With your change — adding custom paths for inference httproute via annotation — those new paths will create separate HTTPRoutes that point directly to the backend service, right? 🤔 

That means:

- Requests to custom paths (e.g., `/custom/infer/my-model`) will **bypass `gateway-plugins` entirely**.
- They go: Client → Envoy Gateway (`aibrix-eg`) → Model Pod (directly)?

```
Client 
  → Envoy Gateway (aibrix-eg) 
  → Custom HTTPRoute (e.g., matches /custom/infer/my-model) 
  → Model Pod (directly)
```
Bypasses `gateway-plugins`

googs1025:
> ~LGTM, over all. I don't have other suggestions for this change.~ if docs needs to be updated, please remember to update it. 😄

I'm not sure if this change is appropriate; if it bypasses the gateway plugin, it doesn't seem like a good option.

rayne-Li:
For now, in v0.5.0, when we create deployment rayclusterfleet or modelAdapter, a httpRoute is created by modelRouterController like `lzy-qw-18b-router `
```go
kubectl -n aibrix-system get httproute
NAME                                       HOSTNAMES   AGE
aibrix-reserved-router                                 2d17h
aibrix-reserved-router-metadata-endpoint               2d17h
lzy-qw-18b-router                                      22h
```

in `lzy-qw-18b-router `, the rules show that if traffic with both  Header: `model: lzy-qw-18b` and Path: `/v1/completions` or `/v1/chat/completions` or `/v1/embeddings`  or `/generate` or `/generatevideo` will directly transfered to endpoints of service `lzy-qw-18b`(the backendRefs)

```yaml
# without this PR
spec:
  parentRefs:
  - group: gateway.networking.k8s.io
    kind: Gateway
    name: aibrix-eg
    namespace: aibrix-system
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: lzy-qw-18b
      namespace: lzy-test
      port: 8000
      weight: 1
    matches:
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /v1/completions
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /v1/chat/completions
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /v1/embeddings
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /generate
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /generatevideo
    timeouts:
      request: 120s
```

this PR only extend more PathPrefix by user, they only effect with `model:lzy-qw-18b` in Header

```
  rules:
  - backendRefs:
    - group: ""
      kind: Service
      name: lzy-qw-18b
      namespace: lzy-test
      port: 8000
      weight: 1
    matches:
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /v1/completions
// ..... extend by custom-paths
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /score
    - headers:
      - name: model
        type: Exact
        value: lzy-qw-18b
      path:
        type: PathPrefix
        value: /version
```

those 2 way you mentioned above are correct, but reserved-router actived when no `model` in Header, if `model` in Header, traffic will go through HTTPRoute `lzy-qw-18b`(this logic exist already)
```
Client 
  → Envoy Gateway (aibrix-eg) 
  → HTTPRoute: reserved-router (matches /v1/...) 
  → gateway-plugins (filter pod)
  → Inference Pod

# HTTPRoute for model already exist, not created by this PR but extend by this PR
Client 
  → Envoy Gateway (aibrix-eg) 
  → Custom HTTPRoute (e.g., matches /custom/infer/my-model) 
  → Model Pod (directly)
```

Jeffwan:
@googs1025 @rayne-Li 

Technically, in AIBrix settings,

- HTTPRoute: backendRefs usually points to a Kubernetes Service. some special case like Lora's endpoint slice etc are fine grained controlled by AIBrix.  this is straightforward, we reuse kubernetes first party objects to manage traffic, underneath routing strategy is kubernetes service's round robin (IPVS mode) / random (iptables).
- Gateway-plugin: custom routing strategies, it directly connect to the pods. user needs to provide "--routing-strategy: xxx"

Earlier, we support http route because at that time, reliability and api compatibility is not that mature using ext-proc.  Now, if user still have needs, I think it's acceptable adding more endpoints. 

there're two apporaches in my mind
- user defined endpoints - exactly as PR proposed, user leverages annotation to control the http endpoint path.
- provide engine templates - check it's vllm or sglang or tensorrt etc, expose all the public facing endpoints. this is a fixed approach but simpler. 

either way works for me.  @rayne-Li have you considered 2nd approach when you propose this PR?

rayne-Li:
@Jeffwan 

sure, like 2nd way said, /generate and /generatevideo are added from v0.4.0 to v0.5.0, but there will always some paths be missed because vllm update first, in my thought adding all vllm/sglang api endpoints are a good way to solve openAI-api format LLM-service

but unlike LLM, in my scenario, i only have one gateway(or router) which is aibrix-eg, small models with custom endpoints often don't have model in payload(cannot route by gateway-plugin with --routing-strategy. model in payload is required), so there is no way routing those traffic without custom-path in httpRoute

docs/source/features/gateway-plugins.rst
```diff
@@ -87,6 +87,28 @@ On a model or lora adapter deployment, their respective controllers create a HTT
...
 
+In v0.5.0, the default PathPrefix configuration within HTTPRoute.Spec.Rules.Matches only includes the following endpoints:
+
+* /v1/completions
+* /v1/chat/completions
+* /v1/embeddings
+* /generate
+* /generatevideo
+
+To add custom path prefixes, you must specify them using the model.aibrix.ai/model-router-custom-paths annotation in your deployment's manifest (or in the modelAdapter, RayClusterFleet, LWS).
```
googs1025: LWS is not support yet.

docs/source/features/gateway-plugins.rst
```diff
@@ -87,6 +87,28 @@ On a model or lora adapter deployment, their respective controllers create a HTT
           Namespace:  aibrix-system
     Events:           <none>
 
+In v0.5.0, the default PathPrefix configuration within HTTPRoute.Spec.Rules.Matches only includes the following endpoints:
+
+* /v1/completions
+* /v1/chat/completions
+* /v1/embeddings
+* /generate
+* /generatevideo
```
googs1025: I'm not sure if this info is necessary? 🤔  The following description and examples seem clear enough.

rayne-Li: removed

docs/source/features/gateway-plugins.rst
rayne-Li: removed

Jeffwan:
> but unlike LLM, in my scenario, i only have one gateway(or router) which is aibrix-eg, small models with custom endpoints often don't have model in payload(cannot route by gateway-plugin with --routing-strategy. model in payload is required), so there is no way routing those traffic without custom-path in httpRoute

that's reasonable. I think the two approaches are not conflicts. We can merge this PR to support custom route. then let's gradually imporve the vllm/sglang compatibility. the custom route in future could be used to support non llm engines or other special case

Jeffwan:
@googs1025 the change looks good to me. Feel free to merge it if all your comments have been addressed

Issue #1913, opened 2026-01-30
APA will not work when using StormService without podGroupSize


rayne-Li:
### 🐛 Describe the bug

according to pkg/controller/podautoscaler/podautoscaler_controller.go, a labelSelector '"stormservice.orchestration.aibrix.ai/pod-group-index"=0' is added 
when using stormService and APA, so only stormService with podGroupSize > 2 will occur autoScaling 

```go
func (r *PodAutoscalerReconciler) computeMetricBasedReplicas(
	ctx context.Context,
	pa autoscalingv1alpha1.PodAutoscaler,
	scalingContext scalingctx.ScalingContext,
	scaleObject *unstructured.Unstructured,
	currentReplicas int32,
) (*ReplicaComputeResult, error) {

	// Stormservice only fetch metrics from pod group index 0 (rank 0)
	if scaleObject.GetAPIVersion() == orchestrationv1alpha1.GroupVersion.String() && scaleObject.GetKind() == StormService {
		requirement, err := labels.NewRequirement(constants.PodGroupIndexLabelKey, selection.Equals, []string{"0"})
		if err != nil {
			return nil, fmt.Errorf(
				"failed to add %s=0 requirement to label selector: %w",
				constants.PodGroupIndexLabelKey, err,
			)
		}
		labelsSelector = labelsSelector.Add(*requirement)
	}
```

maybe `"stormservice.orchestration.aibrix.ai/role-replica-index"` alse need to be considered in this PR #1704 

<img width="995" height="557" alt="Image" src="https://github.com/user-attachments/assets/d65338fb-e9eb-4ba7-92c4-b60bc0299819" />

### Steps to Reproduce

```
apiVersion: autoscaling.aibrix.ai/v1alpha1
kind: PodAutoscaler
metadata:
  name: vllm-replica-apa
  namespace: default
  annotations:
    autoscaling.aibrix.ai/storm-service-mode: "replica"
spec:
  scaleTargetRef:
    apiVersion: orchestration.aibrix.ai/v1alpha1
    kind: StormService
    name: vllm-replica-fix-xpyd

  # optional for replica mode, if role is specified, it will only calculate metrics for specific role, if not specified, it will calculate metrics for all roles
  subTargetSelector:
    roleName: decode

  minReplicas: 1
  maxReplicas: 5
  scalingStrategy: APA

  metricsSources:
    - metricSourceType: pod
      protocolType: http
      port: "8000"
      path: /metrics
[...]

PR #1917, opened 2026-02-02
[Bugfix] add APA support to stormService without podGroupSize (#1913)

rayne-Li:
## Pull Request Description

add support when use stormService + APA without podGroupSize, fix labelSelector for pods

## Related Issues
Resolves: #1913

### DIFF
--- a/build/container/Dockerfile.gateway
+++ b/build/container/Dockerfile.gateway
@@ -1,11 +1,10 @@
 # build gateway plugins
 ARG BUILDER_IMAGE=golang:1.22
 ARG RUNTIME_IMAGE=gcr.io/distroless/base-debian12:nonroot
+FROM ${BUILDER_IMAGE} AS builder
 
-ARG TARGETOS=linux
-ARG TARGETARCH=amd64
-
-FROM --platform=${TARGETOS}/${TARGETARCH} ${BUILDER_IMAGE} AS builder
+ARG TARGETOS
+ARG TARGETARCH
 
 WORKDIR /workspace
 
@@ -55,7 +54,7 @@ RUN set -eux; \
 
 # Use distroless as minimal base image to package the manager binary
 # Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM --platform=${TARGETOS}/${TARGETARCH} ${RUNTIME_IMAGE}
+FROM ${RUNTIME_IMAGE}
 WORKDIR /
 COPY --from=builder /etc/ssl/certs/ca-certificates.crt /etc/ssl/certs/
 COPY --from=builder /workspace/gateway-plugins .


pkg/controller/podautoscaler/workload_scale.go
```diff
@@ -280,21 +280,39 @@ func (s *workloadScale) getPodSelectorForRole(ctx context.Context, pa *autoscali
...
+
+		// add new selector for podGroup index 0, based on #1704
+		if role.PodGroupSize != nil && *role.PodGroupSize > 1 {
+			req, err := labels.NewRequirement(constants.PodGroupIndexLabelKey, selection.Equals, []string{"0"})
+			if err != nil {
+				return nil, err
+			}
+			labelSelector = labelSelector.Add(*req)
+		}
+	}
```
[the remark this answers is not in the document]
rayne-Li: according to the previous logic which i removed in this commit, labelSelector will be added if scaleObject is stormService, no matter what role it has. so i kept that
```
	// Stormservice only fetch metrics from pod group index 0 (rank 0)
	if scaleObject.GetAPIVersion() == orchestrationv1alpha1.GroupVersion.String() && scaleObject.GetKind() == StormService {
		requirement, err := labels.NewRequirement(constants.PodGroupIndexLabelKey, selection.Equals, []string{"0"})
		if err != nil {
			return nil, fmt.Errorf(
				"failed to add %s=0 requirement to label selector: %w",
				constants.PodGroupIndexLabelKey, err,
			)
		}
		labelsSelector = labelsSelector.Add(*requirement)
	}
```

rayne-Li:
@Jeffwan please  take a look

Jeffwan:
@rayne-Li I will take a look today

pkg/controller/podautoscaler/podautoscaler_controller.go
```diff
@@ -1083,17 +1082,6 @@ func (r *PodAutoscalerReconciler) computeMetricBasedReplicas(
 	if err != nil {
 		return nil, fmt.Errorf("failed to get pod selector: %w", err)
 	}
-	// Stormservice only fetch metrics from pod group index 0 (rank 0)
-	if scaleObject.GetAPIVersion() == orchestrationv1alpha1.GroupVersion.String() && scaleObject.GetKind() == StormService {
-		requirement, err := labels.NewRequirement(constants.PodGroupIndexLabelKey, selection.Equals, []string{"0"})
```
Jeffwan, line 1088: your point is stormservice without using podgroup size should not have this label selector, right?

Jeffwan: in that case, can we check the label key first instead of removing it?

rayne-Li: > your point is stormservice without using podgroup size should not have this label selector, right?

yes, only ss with podGroup has label `constants.PodGroupIndexLabelKey`

> in that case, can we check the label key first instead of removing it?

i'm not sure if you miss the last part of workload_scale.go 🤔, i moved and kept that func in workload_scale.go, and add a condition if podGroup is set

image

Jeffwan:
@rayne-Li 
image

could you help rebase the change?

pkg/controller/podautoscaler/podautoscaler_controller.go
Jeffwan: I did miss that part. :D

Issue #1793, opened 2025-11-23
Add Fine-Grained Server-Level Routing in ModelRouter for Multi-Server Pods


paranoidRick:
### 🚀 Feature Description and Motivation

Currently, Aibrix's ModelRouter routes requests at the Pod level, which means that requests are routed to an entire Pod rather than an individual server or inference service within the Pod. However, in scenarios where multiple inference services (such as DP or TP domains) are running within a single Pod, this approach does not offer the necessary granularity to effectively distribute load and ensure efficient utilization of resources.

Problem:

Pod-Level Routing: The ModelRouter only routes requests to the entire Pod, not individual inference servers within that Pod. This causes inefficiency in scenarios where there are multiple services within the same Pod (e.g., multiple GPUs or DP domains).

Multi-Server Pods: When a Pod contains multiple inference services (like DP domains or GPUs), routing requests to the entire Pod can lead to resource bottlenecks, inefficient load balancing, and suboptimal performance.

Lack of Granularity: There's no mechanism to route requests to specific servers within the Pod, resulting in potential overload of certain servers while others remain underutilized.

### Use Case

In scenarios where a Pod contains two different DP domains or multiple GPUs, we want the ModelRouter to route requests to specific servers within the Pod:

Example Pod with two servers (DP domains):
Pod 1:
  ├── dp0: http://pod1:8000
  ├── dp1: http://pod1:8001
Currently, ModelRouter would route requests to Pod 1 as a whole, potentially leading to inefficiencies. With server-level routing, ModelRouter would be able to route requests directly to either dp0 or dp1 based on factors like load and health.

I would like to submit a request for the implementation of server-level routing within Aibrix's ModelRouter to address the limitations mentioned above.

### Proposed Solution

I propose an enhancement to Aibrix's ModelRouter to support server-level routing. The key idea is to:

Track and register each inference server (e.g., dp0, dp1, etc.) within a Pod separately, enabling granular routing decisions.

Route requests to specific servers (rather than Pods) based on load, health, and other relevant factors.

Introduce health checks and failure management mechanisms at the server level to enhance fault tolerance and reliability.

jiangxiaobin96:
one vllm for multi-model or different vllm processes？

paranoidRick:
> one vllm for multi-model or different vllm processes？

If we want true server-level routing, we should run multiple vLLM processes inside one Pod.
[...]

PR #1970, opened 2026-03-03
[feat] Implement multi-port support for prefill pods

paranoidRick:
## Pull Request Description
[Please provide a clear and concise description of your changes here]

## Related Issues
Resolves: #1793

---

[API][Routing] Add Prefiller mulit DP-level API server routing

This PR enhances the least-request routing algorithm to support distributed data-parallel (DP) inference scenarios where pods expose multiple ports for different DP ranks. Previously, routing was performed at the pod level, but for distributed DP workloads, we need to route requests at the API server level, considering individual (pod + port) combinations.

Key Changes:
Extended routing logic in leastRequestRouter to detect when pods have multiple ports (DP > 1) and switch to distributed DP-level routing

Enhanced port discovery in GetPortsForPod with better validation and error handling

one routing strategy:

DP-level routing: When pods have multiple ports (DP>1), route to specific (pod + port) with least active requests

Improved metrics collection that tracks request counts per (pod, port) combination for accurate load balancing

it's only support multiple ports bind with simple pod

Benefits:

✅ Better resource utilization across all DP ranks within a pod

Usage Example:
For a pod configured with data-parallel-size=3 and model-port=8000, the router will now consider three endpoints: pod:8000, pod:8001, pod:8002, and route to the one with the least active requests.

dp == n+1 > 1:
d_set{d0, d1, ,,,dn} belongs to simple pod
exec.proc("cmd", env={export ascend_rt_devices = 0,1,2,3})
vllm serve $MODEL --data-parallel-size 2 --data-parallel-rank 0 --host pod_ip --port 8000
--data-parallel-address {pod_ip} --data-parallel-rpc-port 13345

    exec.proc("cmd", env={export ascend_rt_devices = 4,5,6,7})
    vllm serve $MODEL --data-parallel-size 2 --data-parallel-rank 1 --port 8001\
                      --data-parallel-address {pod_ip} --data-parallel-rpc-port 13345
Related Issues
Resolves: https://github.com/vllm-project/aibrix/issues/1793 (Add Fine-Grained Server-level routing for multi-server pods)

Important: Before submitting, please complete the description above and review the checklist below.


paranoidRick:
Hi, could you help me review this PR, or ask another committer to review it for me? thx 😊 @Jeffwan

Jeffwan:
@paranoidRick I will take a look tomorrow. sorry for the delay.

Jeffwan:
@paranoidRick there're some changes around trtllm support in pd_disaggregation.go which leads the code conflict. could you help rebase the changes?

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -166,7 +172,7 @@ func (r *pdRouter) Route(ctx *types.RoutingContext, readyPodList types.PodList)
 		return "", fmt.Errorf("engine validation failed for request %s: %w", ctx.RequestID, err)
 	}
 
-	prefillPod, decodePod, err := r.filterPrefillDecodePods(ctx, readyPodList.All())
+	prefillPod, prefillPort, decodePod, err := r.filterPrefillDecodePods(ctx, readyPodList.All())
```
Jeffwan: do we need to return prefillPort here? snce prefillPod is a pod object, we can get port from the pod in later steps? this makes the method signature more clean

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -108,6 +112,8 @@ type PrefillRequestTracker struct {
 	podRequestCounts sync.Map // map[string]*int32
 	// Map of request ID -> pod name for cleanup
 	requestToPod sync.Map // map[string]string
+	// Flag to track if any pod has multiple ports
+	isMultiPort bool
```
Jeffwan: Route()  is read concurrently in the methods. it would be great to make it thread safe

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -920,6 +997,30 @@ func (t *PrefillRequestTracker) RemovePrefillRequest(requestID string) {
...
+				counts[pod.Name] = 0
+			} else {
+				counts[pod.Name] = countInterface.(*atomic.Int32).Load()
+			}
+		}
+	}
+	return counts
+}
+
+func (t *PrefillRequestTracker) GetPrefillPortRequestCountsForPods(pods []*v1.Pod) map[string]int32 {
```
Jeffwan: Seems the new created method doesn't have reference? could you help check it?

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -214,31 +226,44 @@ func (r *pdRouter) filterPrefillDecodePods(routingCtx *types.RoutingContext, rea
...
 					"requestId", routingCtx.RequestID, "selectedCombinedPod", combinedPod.Name)
-				return nil, combinedPod, nil
+				return nil, 0, combinedPod, nil
+			}
+		}
+	}
+
+	if r.prefillRequestTracker != nil {
+		for _, pod := range prefillPods {
+			if _, exsit := pod.Labels[dpSizeIdentifier]; exsit {
```
Jeffwan: typo, -> exist

pkg/plugins/gateway/algorithms/pd_disaggregation.go
Jeffwan: is it possible to defer port selection to doPrefillRequest. Right before making the HTTP call, pick the  least-loaded port from the pod's labels

Jeffwan:
@paranoidRick please help take a look. I also like to know how do you deploy DP using aibrix. If you can share more details, we can better polish the PR.

paranoidRick:
> @paranoidRick there're some changes around trtllm support in pd_disaggregation.go which leads the code conflict. could you help rebase the changes?

sure, Thanks for the review. I'll go through your comments and get back to you shortly and rebase the changes.

paranoidRick:
> @paranoidRick please help take a look. I also like to know how do you deploy DP using aibrix. If you can share more details, we can better polish the PR.

```yaml
name: prefill
          replicas: 1
          stateful: true
          template:
            spec:
              schedulerName: volcano
              containers:
                # --- Data Parallel Rank 0 ---
                - name: prefill-dp-rank-0
                  image: quay.io/ascend/vllm-ascend:v0.10.0rc1
                  command: ["/bin/sh", "-c"]
                  args:
                    - |
                      MODEL_PATH="/home/xxx"
                      # 获取当前 Pod IP 用于 RPC 通信
                      POD_IP=$(hostname -i)
                      python3 -m vllm.entrypoints.openai.api_server \
                        --host "0.0.0.0" --port 8000 \
                        --model $MODEL_PATH --served-model-name qwen2-vl-7b-instruct \
                        --distributed-executor-backend mp \
                        --data-parallel-size 2 --data-parallel-rank 0 \
                        --data-parallel-address $POD_IP --data-parallel-rpc-port 13345
                  ports:
                    - name: http-8000
                      containerPort: 8000
                  resources:
                    limits:
                      huawei.com/Ascend910: "4"
                    requests:
                      huawei.com/Ascend910: "4"

                # --- Data Parallel Rank 1 ---
                - name: prefill-dp-rank-1
                  image: quay.io/ascend/vllm-ascend:v0.10.0rc1
                  command: ["/bin/sh", "-c"]
                  args:
                    - |
                      MODEL_PATH="/home/xxx"
                      POD_IP=$(hostname -i)
                      python3 -m vllm.entrypoints.openai.api_server \
                        --host "0.0.0.0" --port 8001 \
                        --model $MODEL_PATH --served-model-name qwen2-vl-7b-instruct \
                        --distributed-executor-backend mp \
                        --data-parallel-size 2 --data-parallel-rank 1 \
                        --data-parallel-address $POD_IP --data-parallel-rpc-port 13345
                  ports:
                    - name: http-8001
                      containerPort: 8001
                  resources:
                    limits:
                      huawei.com/Ascend910: "4"
                    requests:
                      huawei.com/Ascend910: "4"

```

paranoidRick:
> @paranoidRick please help take a look. I also like to know how do you deploy DP using aibrix. If you can share more details, we can better polish the PR.

paranoidRick:
hi, dear Jeffwan @Jeffwan, I have updated the code to support the multi-process DP port allocation we discussed. Could you please review the changes again? I look forward to your feedback. 
Your insights were very helpful! thx!!!

Jeffwan:
@paranoidRick thanks,I will take a look tomorrow

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -523,24 +526,113 @@ func (r *pdRouter) finalPDScore(routingCtx *types.RoutingContext,
...
+
+	// If multiple ports have the same load, randomly select one
+	if len(candidatePorts) > 1 {
+		selectedPort = candidatePorts[rand.Intn(len(candidatePorts))]
+	} else {
+		selectedPort = candidatePorts[0]
+	}
+
+	// Increase load count for selected port
+	portLoadsMap[selectedPort].Add(1)
```
Jeffwan: seems counter.inc() is embedded in the Port selection which is not a good practice.. `get` function shouldn't mutate state.

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -557,13 +649,17 @@ func (r *pdRouter) doPrefillRequest(routingCtx *types.RoutingContext, prefillPod
 		// For SGLang, use async prefill - the bootstrap mechanism (bootstrap_host/port/room)
 		// coordinates between prefill and decode pods, so we don't need to wait
 		go func() {
-			defer r.prefillRequestTracker.RemovePrefillRequest(routingCtx.RequestID)
+			defer func() {
+				r.prefillRequestTracker.RemovePrefillRequest(routingCtx.RequestID)
+				decreasePortLoad()
```
Jeffwan: this is not clean, let's do the defer way..

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -523,24 +526,113 @@ func (r *pdRouter) finalPDScore(routingCtx *types.RoutingContext,
...
 func (r *pdRouter) doPrefillRequest(routingCtx *types.RoutingContext, prefillPod *v1.Pod, llmEngine string) error {
 	// Prepare prefill request payload
 	payload, err := r.preparePrefillPayload(routingCtx, prefillPod, llmEngine)
 	if err != nil {
 		return fmt.Errorf("failed to prepare prefill payload for request %s: %w", routingCtx.RequestID, err)
 	}
 
+	// Select an available port from multiple ports with load balancing
+	port := r.getAvailablePortForPod(routingCtx.RequestID, prefillPod)
+
```
Jeffwan, line 583:
```
// Caller explicitly acquires and defers release
r.acquirePort(prefillPod.Name, port)
defer r.releasePort(prefillPod.Name, port)
```


- Acquire/release are explicit and paired
- defer handles all exit paths (including panics) — no need to manually call in a few places

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -523,24 +526,113 @@ func (r *pdRouter) finalPDScore(routingCtx *types.RoutingContext,
...
+		return utils.GetModelPortForPod(requestID, pod)
+	}
+
+	// If only one port, return it directly
+	if len(ports) == 1 {
+		return ports[0]
+	}
+
+	// Get pod port load information
+	podPortLoads, _ := r.portLoads.LoadOrStore(pod.Name, make(map[int64]*atomic.Int32))
```
Jeffwan: there's still race issue since inner map  is `make(map[int64]*atomic.Int32))` during the initialization?  let's probably use the sync.Map as well in this case or protect initialization with a mutex.

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -143,6 +145,7 @@ func NewPDRouter() (types.Router, error) {
 		httpClient:            httpClient,
 		prefixUpdateCh:        make(chan prefixUpdateJob, 1024),
 		selectionCounts:       make(map[string]int64),
+		portLoads:             sync.Map{},
```
Jeffwan: minor: this will grow forever, while most of the scale won't be that large, this might be a concern, let's leave a comment here.

pkg/plugins/gateway/algorithms/pd_disaggregation.go
paranoidRick: sure, I fix it

pkg/plugins/gateway/algorithms/pd_disaggregation.go
paranoidRick: I have refactored the portLoads structure to portsTracker using sync.Map.

pkg/plugins/gateway/algorithms/pd_disaggregation.go
paranoidRick: Updated. I've renamed it to portsTracker and implemented an Informer-based cleanup logic to prevent the map from growing indefinitely when Pods are deleted. Added comments to document this behavior and the resource reclamation flow as requested.

Looking forward to your feedback.

Jeffwan:
@paranoidRick there're some linter errors. could you run `make fmt` to fix it?

paranoidRick:
> @paranoidRick there're some linter errors. could you run `make fmt` to fix it?

Sure, fix it, looking forward to your feedback.

paranoidRick:
> > @paranoidRick there're some linter errors. could you run `make fmt` to fix it?
> 
> Sure, fix it, looking forward to your feedback.

@Jeffwan hi, could you check the code again

Jeffwan:
@paranoidRick sure. I will also invite @varungup90 to help take a look

pkg/plugins/gateway/algorithms/router.go
```diff
@@ -150,3 +150,27 @@ func (rm *RouterManager) Init() {
 func Init() {
 	defaultRM.Init()
 }
+
+// NotifyOnPodDelete notifies all routers that a pod has been deleted
```
varungup90, line 154: can you describe the rationale for it

pkg/plugins/gateway/gateway.go
```diff
@@ -104,6 +107,67 @@ func NewServer(redisClient *redis.Client, client kubernetes.Interface, gatewayCl
...
+
+// initPodEventListener initializes the pod event listener to listen for pod deletion events
+func (s *Server) initPodEventListener() {
+	if s.client == nil {
+		klog.Warning("Kubernetes client not provided, skipping pod event listener initialization")
+		return
+	}
+
+	var informerFactory informers.SharedInformerFactory
+	namespace := os.Getenv("AIBRIX_POD_NAMESPACE")
```
varungup90, line 123: who populates the namespace? This field is gateway level but user pods are not restricted to be in one namespace.

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -608,22 +634,114 @@ func (r *pdRouter) doPrefillRequest(routingCtx *types.RoutingContext, prefillPod
...
+	} else {
+		selectedPort = candidates[0]
+	}
+
+	return selectedPort
+}
+
+func (r *pdRouter) filterAvailablePortForPod(pod *v1.Pod) []int64 {
+	readyMap := make(map[string]bool)
+	for _, cs := range pod.Status.ContainerStatuses {
```
varungup90, line 697: In kubernetes, pod will only be ready is all containers are ready. Then what is the need to track ready status of individual containers?

varungup90:
PR Review: [#1970](https://github.com/vllm-project/aibrix/pull/1970) — Multi-port support for prefill pods
Summary: Adds port-level load balancing for prefill pods that expose multiple ports (one per DP rank). Overall the approach is sound, but there are several correctness bugs that need addressing before merge.

Bugs
1. filterAvailablePortForPod excludes ports when ContainerStatuses is absent ([pd_disaggregation.go:695-714](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/pd_disaggregation.go#L695))


if ready, ok := readyMap[container.Name]; ok && ready {
If pod.Status.ContainerStatuses is empty (e.g., pod just scheduled, or status not yet propagated), readyMap is empty and all containers are skipped — returning [], which silently falls back to utils.GetModelPortForPod. This means multi-port routing silently degrades to single-port without any warning. At minimum, log a warning when a non-empty port list is expected but none are found.

2. Test cases for "rpc-port excluded" are testing the wrong thing ([pd_disaggregation_test.go:1430-1530](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/pd_disaggregation_test.go#L1430))

Test cases like "pod with RPC port (should be excluded)" and "pod with data parallel port (should be excluded)" don't set ContainerStatuses, so all containers are skipped and filterAvailablePortForPod returns []. getAvailablePortForPod then returns the default port 8000, which happens to be in expectedPorts. The tests pass for the wrong reason — name-based filtering of rpc-port is not actually implemented in filterAvailablePortForPod.

Either:

Add name-based filtering to filterAvailablePortForPod to match the test intent, or
Fix the tests to set ContainerStatuses and adjust expectedPorts to {8000, 8001, 13345} with a comment that name-based exclusion is not yet supported.
3. Non-atomic decrement in releasePort ([pd_disaggregation.go:729-744](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/pd_disaggregation.go#L729))


if counter.Add(-1) < 0 {
    counter.Store(0)  // Race: another goroutine can observe -1 here
}
Add(-1) is atomic but the guard Store(0) is a separate operation — a concurrent reader can briefly observe a negative count. Use a CAS loop instead:


for {
    v := counter.Load()
    if v <= 0 {
        counter.Store(0)
        break
    }
    if counter.CompareAndSwap(v, v-1) {
        break
    }
}
4. acquirePort is called before any error handling ([pd_disaggregation.go:577-582](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/pd_disaggregation.go#L577))


port := r.getAvailablePortForPod(routingCtx.RequestID, prefillPod)
r.acquirePort(prefillPod.Name, port)

// Execute HTTP request
apiURL := fmt.Sprintf(...)
The releasePort defers are placed inside the switch branches. If anything between acquirePort and the switch panics (e.g., a nil deref in fields construction), the port is never released. Consider adding an immediate defer r.releasePort(prefillPod.Name, port) right after acquirePort, and remove the per-branch defers.

Design Concerns
5. portsTracker entry is never updated after pod creation

LoadOrStore is called with the port list discovered at request time. If a container restarts (port list changes) or new containers are added, the stale PodPortBalancer from the first request persists until the pod is deleted. Consider checking if ports have changed and re-initializing the balancer.

6. No port-name filtering for internal ports

filterAvailablePortForPod includes all TCP container ports — metrics ports, health check ports, RPC ports (13345), etc. For DP deployments (as shown in the PR description), only ports 8000, 8001, etc. should be used. Consider filtering by port name convention (e.g., http-*) or adding an annotation/label (like the original dpPortsIdentifier label from the earlier version mentioned in Gemini's summary). This prevents routing real traffic to internal ports.

7. getAvailablePortForPod vs acquirePort inconsistency for single-port

When len(ports) == 1, the code returns that port but never stores it in portsTracker, so acquirePort silently no-ops. This is fine for single-port pods, but the comment at [pd_disaggregation.go:658](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/pd_disaggregation.go#L658) says "For single-port pods, no entry exists in portsTracker" — make this explicit to avoid confusion. Consider using len(ports) <= 1 instead of separate 0 and 1 cases.

Minor
8. dummyCtx in NotifyOnPodDelete — the comment is correct (verified: Register wraps a singleton, so provider(dummyCtx) always returns the same router instance). But it's fragile — add a note that this only works because Register creates a singleton provider.

9. Missing test for acquirePort/releasePort concurrency — the port balancer's core correctness under concurrent acquire/release is not tested. A simple goroutine-concurrent test would increase confidence.

10. hasAibrixLabels checks || not && — any pod with either model.aibrix.ai/name or model.aibrix.ai/engine triggers cleanup. If labels are partially set on non-aibrix pods, spurious cleanups could occur. Verify this is intentional.

varungup90:
From design perspective I will think it through

paranoidRick:
> From design perspective I will think it through

Thank you for the detailed code review and insightful feedback. We will evaluate the feasibility of the proposed solutions and address the bugs as soon as possible.

PR #1979, opened 2026-03-04
fix: disable rate limiting when Redis backend is not configured

### LINKED ISSUE #1874: Ratelimiter is enforced even when Redis backend is not configured
### 🐛 Describe the bug

Ratelimiter is enforced even when the ratelimiter Redis backend is not configured.

### Steps to Reproduce

In our setup, ratelimiter Redis configuration is intentionally omitted, but AIBrix still performs ratelimiter validation and may reject or delay requests. 

### Expected behavior

When ratelimiter backend (e.g. Redis) is not configured, ratelimiter should be treated as disabled and no rate-limit check should be performed.

### Environment

AIBrix v0.5.0

NJX-njx:
## Summary

Fixes #1874

When Redis is not configured (`REDIS_HOST` is empty or Redis connection fails), the gateway plugin now gracefully falls back to a `NoOpRateLimiter` instead of crashing with a fatal error or rejecting all requests.

## Changes

- **`pkg/plugins/gateway/ratelimiter/noop.go`**: Add `NoOpRateLimiter` implementation that always allows requests through, effectively disabling rate limiting
- **`pkg/plugins/gateway/ratelimiter/noop_test.go`**: Add unit tests for `NoOpRateLimiter`
- **`pkg/utils/redis.go`**: Add `TryGetRedisClient()` that returns `nil` instead of calling `klog.Fatalf` when Redis is unavailable
- **`pkg/plugins/gateway/gateway.go`**: Modify `NewServer` to use `NoOpRateLimiter` when `redisClient` is nil
- **`cmd/plugins/main.go`**: Use `TryGetRedisClient()` and handle nil redis client in defer

## Behavior

| Scenario | Before | After |
|----------|--------|-------|
| Redis configured and running | Rate limiting works | Rate limiting works (no change) |
| Redis not configured (REDIS_HOST empty) | Fatal crash on startup | Starts normally, rate limiting disabled |
| Redis configured but unreachable | Fatal crash on startup | Starts normally, rate limiting disabled with warning |

### DIFF
--- a/cmd/plugins/main.go
+++ b/cmd/plugins/main.go
@@ -64,10 +64,12 @@ func main() {
 		klog.Fatal("--endpoints-config is required when running in standalone mode")
 	}
 
-	redisClient := utils.GetRedisClient()
+	redisClient := utils.TryGetRedisClient()
 	defer func() {
-		if err := redisClient.Close(); err != nil {
-			klog.Warningf("Error closing Redis client: %v", err)
+		if redisClient != nil {
+			if err := redisClient.Close(); err != nil {
+				klog.Warningf("Error closing Redis client: %v", err)
+			}
 		}
 	}()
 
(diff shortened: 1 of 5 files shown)


varungup90:
If redis backend is not configured then rate limiting is not applied.

PR #1989, opened 2026-03-05
feat: add state sync for in-memory cache of aibrix-gateway instances

### LINKED ISSUE #761: Stateful information sync for ext-proc Instances

varungup90:
## Pull Request Description

<html><head></head><body>🚧 Problem</h2><p>When the AIBrix gateway runs with multiple replicas, in-memory state (e.g., prefix cache blocks used for KV-cache-aware routing) is maintained independently per replica.</p><p>Over time, these states diverge, leading to <strong>suboptimal routing decisions</strong> — each replica lacks visibility into prefix cache blocks observed by others.</p><p><strong>Resolves:</strong> #761</p><hr><h2>✅ Solution</h2><p>This PR introduces a <strong>generic Redis-backed state synchronization layer</strong> for multi-replica gateway deployments:</p><ul><li><p>📦 <code inline="">pkg/plugins/gateway/redissync</code></p><ul><li><p>Handles cross-replica state sharing</p></li><li><p>Uses simple Redis primitives:</p><ul><li><p><code inline="">SETEX</code>, <code inline="">MGET</code>, <code inline="">SCAN</code>, <code inline="">DEL</code></p></li></ul></li><li><p>No reliance on Pub/Sub or Lua scripting</p></li></ul></li><li><p>🔌 <code inline="">pkg/utils/syncable/syncable.go</code></p><ul><li><p>Defines a <code inline="">Syncable</code> interface</p></li><li><p>Keeps the sync layer fully decoupled from component-specific logic</p></li></ul></li></ul><hr><h2>🧠 Key Design Decisions</h2>

* **Key layout**
  Keys follow the format `aibrix:{namespace}:e:{entityId}`.
  The namespace acts as a Redis Cluster hash tag to ensure keys are routed to the same shard. Each key is stored with an independent TTL using `SETEX`.

* **Pull-before-push**
  Each sync cycle first pulls and merges remote state, then pushes local updates. This ordering minimizes the risk of blind overwrites and ensures replicas reconcile before publishing their own view.

* **Delta sync**
  If a `Syncable` also implements `DeltaSyncable`, only modified entities are pushed instead of a full snapshot. This significantly reduces write amplification and improves efficiency.

* **Tombstones**
  Deletions are propagated as tombstone payloads. Other replicas consume these during the next pull cycle and remove the corresponding local entries.

* **Stale policy**
  Implementations can mark remote entries as stale. The sync layer will delete these entries from Redis and skip applying them locally, preventing outdated data from persisting.

* **Jitter & backoff**
  To avoid synchronization spikes, replicas introduce a random startup delay and apply ±20% jitter to each sync cycle. Errors trigger an exponential backoff strategy with a capped maximum delay.

<hr><h2>🔄 Extensibility</h2><p>This design is intentionally <strong>generic and pluggable</strong>:</p><ul><li><p>Any component can opt into cross-replica sync by:</p><ol><li><p>Implementing <code inline="">syncable.Syncable</code></p></li><li><p>Registering with <code inline="">redissync.Manager</code></p></li></ol></li><li><p>The <code inline="">redissync</code> package:</p><ul><li><p>Contains <strong>no component-specific dependencies</strong></p></li><li><p>Is reusable across different gateway features</p></li></ul></li></ul><p>👉 See <code inline="">pkg/plugins/gateway/redissync/README.md</code> for:</p><ul><li><p>Step-by-step integration guide</p></li><li><p>Example implementation (<code inline="">TokenTracker</code>)</p></li></ul><hr><h2>💡 Why This Approach?</h2><ul><li><p>Keeps coordination <strong>simple and predictable</strong> (no distributed locks or streaming)</p></li><li><p>Minimizes <strong>write amplification</strong> via delta sync</p></li><li><p>Ensures <strong>eventual consistency</strong> with low operational overhead</p></li><li><p>Maintains <strong>strong separation of concerns</strong></p></li></ul><hr><ul></ul></body></html>

## Related Issues
Resolves: #761

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>


googs1025:
will help to check in these day

varungup90:
> will help to check in these day

@googs1025 can you please help review today? Thanks

googs1025:
> > will help to check in these day
> 
> @googs1025 can you please help review today? Thanks

ok

pkg/utils/lrustore/lru_store.go
```diff
@@ -54,6 +56,10 @@ func NewLRUStore[K comparable, V any](cap int, ttl, interval time.Duration, f ge
 	store.lruList.tail.prev = store.lruList.head
 
 	go store.startEviction()
+	dumpInterval := 60
+	if dumpInterval > 0 {
+		go store.startDebugDump(time.Duration(dumpInterval) * time.Second)
```
googs1025: `startDebugDump` runs `for range ticker.C` with no stop mechanism. Every `LRUStore` instance spawns a goroutine that never exits.  is here will be goroutine leak ?

varungup90: I added this for debugging purpose, will remove it

pkg/plugins/gateway/redissync/redissync.go
```diff
@@ -0,0 +1,501 @@
...
+		return fmt.Errorf("redissync push SETEX %s: %w", namespace, err)
+	}
+	klog.V(4).InfoS("redissync pushed full snapshot", "namespace", namespace, "entries", n)
+	return nil
+}
+
+// Start starts the background goroutine: optional initial sync (with small random
+// delay), then periodic Pull-first then optional Push, with jitter and backoff on errors.
+// Register must be called before Start.
+func (r *RedisSync) Start() {
```
googs1025, line 449: `Stop()` closes `stopCh` but doesn't wait for the `syncLoop` goroutine to finish. During shutdown, an in-flight Redis operation could be interrupted mid-cycle.

pkg/plugins/gateway/redissync/redissync.go
```diff
@@ -0,0 +1,501 @@
...
+	}
+	if nDeleted > 0 {
+		chunkSize := r.setexChunkSize
+		if chunkSize <= 0 {
+			chunkSize = setexChunkSize
+		}
+		ts, hasTombstone := s.(syncable.TombstoneSupport)
+		pipe := r.client.Pipeline()
+		n := 0
+		flushPipe := func() {
```
googs1025: In `pushDelta`, `flushPipe()` logs pipeline errors at V(4) but doesn't propagate them.

pkg/plugins/gateway/redissync/README.md
```diff
@@ -0,0 +1,201 @@
...
+
+- **Periodic sync**: Each replica runs **Pull first** (load from Redis), then **Push** (delta or full) every sync period.
+- **Write-through**: After local updates, call `Put`/`Delete` so other replicas see changes on the next pull.
+- **Deletions/tombstones**: If a Syncable implements `TombstoneSupport`, deletions are written as tombstones (SETEX) and Pull deletes local entries upon detection; otherwise DEL is used.
+- **Jitter and backoff**: Random initial delay and ±20% jitter on interval; exponential backoff (capped) on errors.
+- **Lifecycle**: Register all Syncables **before** `Start()`; empty namespace is rejected.
+
+**Per-record TTL:** Each entity is stored as its own key `aibrix:{namespace}:e:{entityId}` with `SETEX` and a 5-minute TTL. So each record expires 5 minutes after it was last written; the TTL window does not slide for other records when one is updated.
```
googs1025: recordTTL = 2 * time.Minute or 5 minutes

varungup90: 2 mins.

cmd/plugins/main.go
```diff
@@ -158,6 +160,19 @@ func main() {
 
 	gatewayServer := gateway.NewServer(redisClient, k8sClient, gatewayK8sClient)
 
+	redissyncEnabled := utils.LoadEnvBool("AIBRIX_REDISSYNC_ENABLED", false)
+	var syncManager *redissync.RedisSync
```
Jeffwan: `redissync` package looks like a implementation driven approach. can we rename to something like `statesync`, and the redis is one of the implementations?

config/test/gateway/vtc-test-env-patch.yaml
```diff
@@ -17,3 +18,9 @@ spec:
           value: "100"
         - name: AIBRIX_ROUTER_VTC_TOKEN_TRACKER_MAX_TOKENS
           value: "800"
+        - name: AIBRIX_REDISSYNC_ENABLED
```
Jeffwan, line 21: please add the envs to gateway docs.

config/test/gateway/vtc-test-env-patch.yaml
```diff
@@ -17,3 +18,9 @@ spec:
...
+          value: "true"
+        - name: AIBRIX_REDISSYNC_SYNC_PERIOD
+          value: "1s"
```
Jeffwan, line 24: is there estimation on the load? I am not sure whether this is too short for the sync.  what's the best practice here?

config/test/gateway/vtc-test-env-patch.yaml
```diff
@@ -17,3 +18,9 @@ spec:
...
+        - name: AIBRIX_PREFIX_CACHE_BLOCK_SIZE
```
Jeffwan, line 25: what's the default value. should we configure to the make size as engine? let's add to docs

Jeffwan:
it seems the gateway instance on the push side flush the kv every 10s (by default), the pull side needs to dump all the data and merge the view. 
1. scan cost grows with keyspace. do we have some estimate on the space? I just probably just some rough performance data
2. did you compare the case that push periodically vs push for every prefix hash change? 

there's actually another way. for all instances, just watch the kv events from all engines, this is adopted by other solutions. it would be great for us to compare these first.

config/test/gateway/vtc-test-env-patch.yaml
varungup90: This is only for e2e test. Default value is 10s.

config/test/gateway/vtc-test-env-patch.yaml
varungup90: This is only for e2e test. Default value is 10s.

varungup90:
> 1. scan cost grows with keyspace. do we have some estimate on the space? I just probably just some rough performance data

- The default per-key TTL is 2 minutes, so the active keyspace stays proportional to the working set of the last 2 minutes rather than growing unboundedly.

- Users can also increase AIBRIX_PREFIX_CACHE_BLOCK_SIZE (default: 128 tokens/block) to reduce the number of Redis entries per request 

- Benchmark run:

```
python3 -m vllm.entrypoints.cli.main bench serve \
    --tokenizer /models/Qwen3-8B \
    --model qwen3-8B \
    --base-url http://124.174.21.2:80 \
    --endpoint /v1/completions \
    --backend openai \
    --metric_percentiles '50,90,95,99' \
    --goodput ttft:5000 tpot:250 \
    --random-prefix-len 10 \
    --dataset-name random \
    --seed 1 \
    --percentile-metrics "ttft,tpot,itl,e2el" \
    --random-input-len 4096 \
    --random-output-len 100 \
    --max-concurrency 32 \
    --request-rate 8 \
    --num-prompts 500 \
    --ready-check-timeout-sec 30
```

benchmark results

varungup90:
> 2\. did you compare the case that push periodically vs push for every prefix hash change?

I haven't benchmarked write-through vs. periodic, but I believe periodic is the better design for prefix cache sync for a few reasons:

1. Write amplification on the hot path
AddPrefix is called on every request and iterates over all prefix blocks — a single request can touch hundreds of blocks. Write-through would issue one Redis SETEX per block per request, multiplying Redis write load by blocks_per_request × QPS. At any meaningful throughput this can saturate the Redis connection pool and affect all replicas sharing the instance.

2. Lock contention on the routing critical path
AddPrefix holds PrefixHashTable.mu (write lock) while updating the LRU store. A synchronous Redis call inside that lock extends hold time from nanoseconds to a full network RTT (~1ms+). Since seqSearchPrefix acquires the same lock as a read lock during routing, this directly increases p99 routing latency under load. An async fire-and-forget approach avoids the lock issue but adds goroutine lifecycle complexity.

3. Redis failure isolation
Decoupling writes from the request path means a Redis slowdown or outage has zero impact on routing latency. The periodic sync loop handles retries and backoff independently.

That said, for use cases requiring sub-second propagation (one scenario to implement next is sharing request count per pod), EncodeBlockForSync + RedisSync.Put() is available as an opt-in write-through path for specific high-value, low-frequency events.

varungup90:
> there's actually another way. for all instances, just watch the kv events from all engines, this is adopted by other solutions. it would be great for us to compare these first.


Both approaches have their own trade-offs, and the right choice depends on the deployment priorities.

**Redis-based KV cache sync**

* The current routing latency percentiles observed in production are as follows (P50: 1ms, P90: 1.8ms, P99: 2.9ms), and since Redis-based synchronization operates in parallel with the routing path, it is not expected to have any impact on routing performance. 
* Enables a more **general mechanism**: we’re planning to support syncing arbitrary metadata across gateway instances, not just KV cache
* Provides a centralized and consistent view of shared state
* Comes with an external dependency (Redis) and additional operational overhead

**KV cache sync via watch events on the vLLM engine**

* No external system required; simpler deployment model
* More tightly coupled to the inference engine
* Currently **vLLM-specific** (not a general metadata sync solution)
* Higher observed latency in current measurements:

  * P50: ~17–19ms
  * P90: ~25–28ms
  * P99: up to ~116ms

In summary, Redis-based sync provides better performance and a more extensible foundation for cross-instance state sharing, while the watch-based approach keeps the system simpler but is more limited in scope and currently shows higher latency. Users can choose based on whether they prioritize performance and extensibility or operational simplicity.

Yang1032:
@varungup90 Hi, what's the status of this PR? If we combine it with #2159, is it ready to test multi-gateway instances?

varungup90:
> @varungup90 Hi, what's the status of this PR? If we combine it with #2159, is it ready to test multi-gateway instances?

If you enable KV cache sync then this PR is optional. Secondly even without this PR, you can test multi-instance gateway.

@Jeffwan Do you know when you can review PR again? Addressed your comments.

Yang1032:
> > @varungup90 Hi, what's the status of this PR? If we combine it with #2159, is it ready to test multi-gateway instances?
> 
> If you enable KV cache sync then this PR is optional. Secondly even without this PR, you can test multi-instance gateway.
> 
> @Jeffwan Do you know when you can review PR again? Addressed your comments.

Thanks for the reply!
We are not using the KV cache sync feature because our backend engine is not vLLM. We are strictly relying on the default prefix-cache routing algorithm.

Since the default prefix-cache uses a local in-memory PrefixHashTable, running multiple gateway instances leads to isolated states  and routing inconsistencies across replicas in our environment.

That's exactly why we are looking into this PR. I will pull this branch and test the prefix-cache multi-replica scenario on our end. Thanks again for your work!

XiaozanZhang:
@varungup90 
Hello, I'm glad that you have implemented a Redis synchronized multi-instance prefix cache instance to solve the cache consistency problem across multiple instances. I have a few questions regarding the implementation details and hope to get your answers.

1. Why not directly put all the KV cache into Redis for direct queries instead of having each instance maintain a copy? Is it because frequent Redis queries result in high IO and performance degradation? However, this approach would require each replica to allocate a large amount of memory to store all the KV cache, which may lead to strong consistency issues for the data?

2. The current code supports a standalone Redis , why not support Redis cluster?

Thank you,

PR #2002, opened 2026-03-13
Fix: Support Chat Template Tokenization with vLLM Parameters in Prefix Cache Router

penfree:
## Problem

There were two critical issues causing prefix cache mismatches between Gateway and vLLM Pods in kv-sync mode:

1. **Double JSON Serialization Issue**: `ChatMessage.Content` was defined as `string`, causing structured multimodal content (images, audio) to be double-encoded when sent to vLLM tokenization API. This resulted in Gateway tokenizing differently than vLLM Pods.

2. **Missing Chat Template Support**: The `kvSyncPrefixCacheRouter` used simple text tokenization (`TokenizeInputText`) for all requests, instead of applying chat templates for `/v1/chat/completions` endpoints. This caused prefix cache misses because Gateway wasn't tokenizing messages the same way vLLM does with its chat template.

### Example of the Issue

**Before this fix:**
- Client sends: `{"role": "user", "content": [{"type": "text", "text": "Hello"}, {"type": "image_url", ...}]}`
- Gateway tokenizes: `"[{\"type\":\"text\",\"text\":\"Hello\"}...]"` (double-encoded string)
- vLLM Pod tokenizes: `[{"type": "text", "text": "Hello"}...]` (structured array)
- Result: **Prefix cache miss** ❌

**After this fix:**
- Both Gateway and vLLM Pod tokenize the same way with chat template
- Result: **Prefix cache hit** ✅

### 1. Define vLLM-Compatible Request Structure

**File**: `pkg/types/request.go` (NEW)

Created `ChatCompletionRequest` that embeds `openai.ChatCompletionNewParams` and adds vLLM-specific parameters:
- `add_generation_prompt` - Controls whether to add generation prompt to chat template (default: true)
- `add_special_tokens` - Controls whether to add special tokens on top of chat template (default: false)
- `return_token_strs` - Returns token strings for debugging (default: false)

These parameters match vLLM's [chat completion protocol](https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/chat_completion/protocol.py#L217-244).

### 2. Fix ChatMessage Content Type

**File**: `pkg/utils/tokenizer/types.go`

Changed `ChatMessage.Content` from `string` to `json.RawMessage`:
```go
type ChatMessage struct {
    Role    string          `json:"role"`
    Content json.RawMessage `json:"content"` // Now supports string or structured array
}
```

This preserves the original JSON structure without double-encoding, supporting:
- Simple text: `"Hello world"`
- Multimodal arrays: `[{"type": "text", ...}, {"type": "image_url", ...}]`

### 3. Implement Chat Template Tokenization in Router

**File**: `pkg/plugins/gateway/algorithms/prefix_cache.go`

Added `buildTokenizeInputFromChatRequest()` helper that:
1. Converts OpenAI message format to tokenizer format
2. Preserves multimodal content structure
3. Extracts vLLM-specific parameters with correct defaults

Updated `kvSyncPrefixCacheRouter.Route()` to:
1. Check endpoint type (`/v1/chat/completions` vs `/v1/completions`)
2. For chat endpoints:
   - Parse request as `ChatCompletionRequest`
   - Build `TokenizeInput` with chat template parameters
   - Use `TokenizeWithOptions` with `ChatInput` type
3. For other endpoints: Keep existing text tokenization
4. Implement comprehensive fallback mechanism

### 4. Export Helper Functions

**File**: `pkg/utils/tokenizer/utils.go`

Exported `IntToByteArray` function for converting token IDs to byte array format needed by prefix cache indexer.

### Unit Tests Added

1. **`types_test.go`**: Tests for ChatMessage serialization/deserialization
   - Simple text content
   - Multimodal array content
   - Round-trip preservation

2. **`serialization_test.go`**: Tests for sonic/JSON compatibility
   - Verifies `sonic.Marshal` correctly handles `json.RawMessage`
   - Validates vLLM request format matches expected structure
   - Confirms multimodal content is not double-encoded

### Test Results

```bash
$ go test ./pkg/utils/tokenizer/...
ok      github.com/vllm-project/aibrix/pkg/utils/tokenizer    4.536s

$ go test -v ./pkg/utils/tokenizer -run TestVLLMRequestSerialization
Serialized request:
{"messages":[{"role":"user","content":[{"type":"text","text":"test"},
{"type":"image_url","image_url":{"url":"http://example.com/img.jpg"}}]}],
"add_special_tokens":false,"add_generation_prompt":true}
--- PASS: TestVLLMRequestSerialization (0.00s)
```

### Supported Engines

- ✅ **vLLM**: Fully compatible, enhanced with chat template support
- ✅ **SGLang**: Fully compatible (doesn't use tokenization API, no impact)

### Backward Compatibility

- ✅ Existing `/v1/completions` endpoints work unchanged
- ✅ Fallback to text tokenization if chat template fails
- ✅ Compatible with tokenizers that don't support `ExtendedTokenizer`
- ✅ Default parameter values match vLLM behavior

## Benefits

1. **Improved Prefix Cache Hit Rate**: Gateway now tokenizes chat messages the same way as vLLM Pods, significantly improving cache hits in kv-sync mode

2. **Multimodal Content Support**: Properly handles images, audio, and other structured content in chat messages

3. **vLLM Parameter Compatibility**: Gateway can now parse and respect vLLM-specific chat template parameters

4. **Better Debugging**: Logs detailed tokenization information at V(4) level

5. **Robust Error Handling**: Multiple fallback layers ensure the router continues working even if chat tokenization fails

## Files Changed

| File | Status | Changes |
|------|--------|---------|
| `pkg/types/request.go` | NEW | +45 | ChatCompletionRequest struct |
| `pkg/utils/tokenizer/types.go` | Modified | ChatMessage.Content → json.RawMessage |
| `pkg/utils/tokenizer/utils.go` | Modified | Export IntToByteArray |
| `pkg/plugins/gateway/algorithms/prefix_cache.go` | Modified | +112 | Add chat template support |
| `pkg/utils/tokenizer/types_test.go` | NEW | +106 | Unit tests |
| `pkg/utils/tokenizer/serialization_test.go` | NEW | +135 | Serialization tests |

**Total**: 6 files changed, 274 insertions(+), 12 deletions(-)

## Verification Steps

To verify this fix works correctly:

1. **Test simple chat completion**:
```bash
curl -X POST http://gateway/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "test-model",
    "messages": [{"role": "user", "content": "Hello"}]
  }'
```

2. **Test multimodal content**:
```bash
curl -X POST http://gateway/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "test-model",
    "messages": [{
      "role": "user",
      "content": [
        {"type": "text", "text": "What is in this image?"},
        {"type": "image_url", "image_url": {"url": "https://..."}}
      ]
    }]
  }'
```

3. **Test with vLLM parameters**:
```bash
curl -X POST http://gateway/v1/chat/completions \
  -H "Content-Type: application/json" \
  -d '{
    "model": "test-model",
    "messages": [{"role": "user", "content": "Hello"}],
    "add_generation_prompt": true,
    "add_special_tokens": false
  }'
```

4. **Verify tokenization consistency**:
Compare token IDs from Gateway tokenization with vLLM tokenization API:
```bash
# Gateway logs should show:
# tokenized using chat template: message_count=1, token_count=X, add_generation_prompt=true
```

## Related Issues

This fix addresses the root cause of prefix cache mismatches in kv-sync mode when:
- Using chat completion endpoints
- Sending multimodal content (images, audio)
- vLLM chat template parameters differ from defaults

## Breaking Changes

**None** - This is a backward-compatible enhancement. All existing functionality remains unchanged.

### DIFF
--- a/pkg/plugins/gateway/algorithms/prefix_cache.go
+++ b/pkg/plugins/gateway/algorithms/prefix_cache.go
@@ -473,7 +473,11 @@ func buildTokenizeInputFromChatRequest(chatReq *types.ChatCompletionRequest) (*t
 
 		// Directly marshal the content to preserve its structure
 		content := msg.GetContent()
-		contentJSON, err := json.Marshal(content.AsAny())
+		contentAny := content.AsAny()
+		if contentAny == nil {
+			return nil, fmt.Errorf("message at index %d has nil content", i)
+		}
+		contentJSON, err := json.Marshal(contentAny)
 		if err != nil {
 			return nil, fmt.Errorf("failed to marshal message content at index %d: %w", i, err)
 		}


pkg/types/request.go
```diff
@@ -0,0 +1,45 @@
...
+// ChatCompletionRequest extends OpenAI's ChatCompletionNewParams with vLLM-specific parameters
+// Reference: https://github.com/vllm-project/vllm/blob/main/vllm/entrypoints/openai/chat_completion/protocol.py
+type ChatCompletionRequest struct {
+	openai.ChatCompletionNewParams
+
+	// vLLM-specific parameters
+
+	// AddGenerationPrompt controls whether to add the generation prompt to the chat template.
+	// Default: true (vLLM default)
+	// Reference: protocol.py line 217-224
```
Jeffwan: it would be great to link the file link directly.

pkg/utils/tokenizer/types.go
```diff
@@ -49,9 +50,10 @@ type TokenizeResult struct {
 }
 
 // ChatMessage represents a single message in a chat conversation
+// Content can be either a string (simple text) or a structured array (multimodal)
 type ChatMessage struct {
-	Role    string `json:"role"`
-	Content string `json:"content"`
+	Role    string          `json:"role"`
+	Content json.RawMessage `json:"content"` // Can be string or array for multimodal
```
Jeffwan, line 56: looks good

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package routingalgorithms
 
 import (
+	"encoding/json"
```
Jeffwan, line 20: the project adopts `github.com/bytedance/sonic` over `encoding/json` to optimize the performance, let's move to sonic if the scenario is supported.

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -455,6 +456,115 @@ func (p *prefixCacheRouter) Cleanup() error {
...
+	// Convert OpenAI messages to tokenizer messages, preserving content structure
+	messages := make([]tokenizer.ChatMessage, len(chatReq.Messages))
+	for i, msg := range chatReq.Messages {
+		role := msg.GetRole()
+		if role == nil {
+			return nil, fmt.Errorf("message at index %d has no role", i)
+		}
+
+		// Directly marshal the content to preserve its structure
+		content := msg.GetContent()
```
Jeffwan, line 475: If a message variant isn't recognized by the SDK, content.AsAny()  may returns nil? i would suggest to add a nil check here. 

```
  if contentAny == nil {
      return nil, fmt.Errorf("message at index %d has nil content", i)
  }
```

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -455,6 +456,115 @@ func (p *prefixCacheRouter) Cleanup() error {
...
+		contentJSON, err := json.Marshal(content.AsAny())
```
Jeffwan: use sonic?

Jeffwan: I notice it might be have some limitation. If the SDK uses standard json.Unmarshaler interface, sonic handles it fine. If it uses reflect-based registration or other non-standard mechanisms (which apijson.RegisterUnion suggests), sonic might not parse the union types correctly. 

To be safe, we can keep `encoding/json` for open ai marshall and unmarshall here. Let's make the function works as expected first.

Jeffwan:
@penfree  thanks for driving this effort! super helpful. Please help take a look at the comment.

Jeffwan:
overall looks good to me just some minor feedback

penfree:
@Jeffwan i've already followed your suggestion except encoding/json, PTAL

Issue #2003, opened 2026-03-13
race condition test failure from FAIL: TestLRUStore_TTL


Jeffwan:
### 🐛 Describe the bug

<img width="1530" height="1130" alt="Image" src="https://github.com/user-attachments/assets/d4113a88-0ee8-49b9-97d0-59bdb3266037" />

https://github.com/vllm-project/aibrix/actions/runs/23065542496/job/67002818054

### Steps to Reproduce

seems this is flaky test, `make test`  may help reproduce the issue

### Expected behavior

N/A

### Environment

main branch

xvchris:
I can take this. The race is in `TestLRUStore_TTL`: it relies on `time.Sleep` but `Put` calls `time.Now()` directly instead of the injectable `getCurrentTime`, so the mock clock never controls the entries. I'll fix `Put` to use `e.getCurrentTime()` and rewrite the test to use a fake clock with manual eviction.

PR #2007, opened 2026-03-14
[Bug]: fix flaky TestLRUStore_TTL by using injectable clock in Put

xvchris:
## Pull Request Description
`Put()` was calling `time.Now()` directly instead of `e.getCurrentTime()`, so the injectable clock had no effect on entry timestamps. This made `TestLRUStore_TTL` rely on real wall time (`time.Sleep`), which is inherently flaky under slow CI.

Changes:
- Use `e.getCurrentTime()` in `Put()` for consistent time injection
- Rewrite `TestLRUStore_TTL` with a fake clock and manual `evict()` call, removing the 3-second sleep
- Fix `TestLRUStore_ConcurrentEvictions` to use `store.Len()` instead of directly accessing `freeTable` without a lock (avoids race detector)

All tests pass with `go test -race`.

## Related Issues
Resolves: #2003

### DIFF
--- a/pkg/utils/lrustore/lru_store.go
+++ b/pkg/utils/lrustore/lru_store.go
@@ -37,6 +37,7 @@ type LRUStore[K comparable, V any] struct {
 	getCurrentTime
 	interval time.Duration
 	ttl      time.Duration
+	stop     chan struct{}
 }
 
 func NewLRUStore[K comparable, V any](cap int, ttl, interval time.Duration, f getCurrentTime) *LRUStore[K, V] {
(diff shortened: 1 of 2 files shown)


Jeffwan:
thanks for the quick fix. @xvchris could you help sign off the commit and also take a look at the gemini's comment?

xvchris:
Done — added sign-off and addressed the goroutine leak. Added a `Close()` method with a `stop` channel so `startEviction` exits cleanly, and called `t.Cleanup(func() { store.Close() })` in all four tests.

PR #2024, opened 2026-03-18
Feat: Support RequestTracker callback & add power of two router

penfree:
## Summary                                                                                                                
                  
  This PR introduces a flexible **RequestTracker** callback mechanism to the cache system and  a new **Power of Two** load balancing router that  Redis for distributed request tracking.

  ## Motivation

  1. **Extensible Request Tracking**: Different routing algorithms may need custom request lifecycle tracking (e.g.,
  Redis-based counting, database logging, custom ). The existing cache implementation lacked a mechanism for routers  to register custom tracking logic.

  2. **Need for Distributed Load Balancing**: The existing `least-request` router only works within a single Gateway
  instance. In multi-Gateway deployments, we need distributed request tracking to make  optimal routing decisions.

  3. **Power of Two Algorithm Benefits**:
     - **O(1) complexity** with near-optimal load distribution
     - **Distributed support** via Redis
     - **Multi-port support** for Data Parallel (DP) mode
     - Better than pure random, simpler than maintaining perfect least-request across instances

  ## Changes

  ### 1. RequestTracker Callback Mechanism (`pkg/cache/`)

  **Added `RequestTracker` interface for lifecycle tracking:**
  ```go
  type RequestTracker interface {
      AddRequestCount(ctx *types.RoutingContext, requestID, modelName string) int64
      DoneRequestCount(ctx *types.RoutingContext, requestID, modelName string, traceTerm int64)
      DoneRequestTrace(ctx *types.RoutingContext, requestID, modelName string,
                       inputTokens, outputTokens, traceTerm int64)
  }

  Added CanRegisterRequestTracker interface for cache:
  type CanRegisterRequestTracker interface {
      RegisterRequestTracker(tracker RequestTracker)
  }

  Key Features:
  - Routers can  RequestTracker to receive lifecycle callbacks
  - Cache  CanRegisterRequestTracker to accept tracker 
  - Registered via cache.RegisterRequestTracker(tracker)
  - Callbacks are  at request start, completion, and failure
  - Maintains backward  (existing routers unaffected)

  2. Power of Two Router (pkg/plugins/gateway/algorithms/power_of_two.go)

  Implementation Details:

  Core Algorithm:
  - Randomly selects two candidates (pod:port combinations)
  -  Redis for current request count of each candidate
  - Routes to the candidate with fewer running requests
  - Falls back to single candidate if only one available

  Multi-Port Support:
  - Fully supports Data Parallel (DP) mode
  - Tracks request count at pod:port granularity
  - Example: pod with 4 ports (8000-8003) → 4 independent candidates

  Redis Key Design:
  po2_req_count:<modelName>:<podName>_<port>:<rotatedTimestamp>

  Key Rotation Mechanism:
  rotatedTimestamp = currentTimestamp - (currentTimestamp % keyRotationSec)
  - Default rotation interval: 3600 seconds (1 hour)
  - Prevents stale count accumulation
  - Hardcoded in defaultKeyRotationSec constant

  Request Lifecycle:
  1. Route(): Select target using power of two
  2. AddRequestCount():  Redis counter, return traceTerm
  3. DoneRequestCount(): Decrement Redis counter when request completes/fails
  4. DoneRequestTrace():  to DoneRequestCount (simplified for this router)

  Error Handling:
  - Graceful handling of Redis connection failures (returns 0 count)
  -  cache check to prevent panics during initialization

  3.  (cmd/plugins/main.go)

  redisClient := utils.GetRedisClient()
  routingalgorithms.RegisterPowerOfTwoRouter(redisClient)

  The router automatically registers itself as a RequestTracker during initialization:
  func NewPowerOfTwoRouterWithRedis(redisClient *redis.Client, keyRotationSec ...int64) *PowerOfTwoRouter {
      // ... router setup ...
      c,  := cache.Get()
      if err ==  {
          c.RegisterRequestTracker(router)  // Self-
      }
      return router
  }

  Testing

  Unit Tests Added (power_of_two_test.go):
  - ✅ TestPowerOfTwoRouter_buildCandidates - Single/multi-port pod handling
  - ✅ TestPowerOfTwoRouter_podServerKey - Key formatting
  - ✅ TestPowerOfTwoRouter_buildRedisKey - Redis key generation with rotation
  - ✅ TestPowerOfTwoRouter_keyRotation - Time-based key rotation
  - ✅ TestPowerOfTwoRouter_Integration - End-to-end with Redis (skipped in short mode)

  All tests pass:
  === RUN   TestPowerOfTwoRouter
  --- PASS: TestPowerOfTwoRouter (0.00s)
  ok      github.com/vllm-project/aibrix/pkg/plugins/gateway/algorithms

  Usage

  1. :
  # Set routing algorithm to power-of-two
  export AIBRIX_ROUTING_ALGORITHM=power-of-two

  #  Redis connection (required)
  export REDIS_ADDR=redis:6379
  export REDIS_PASSWORD=your-

 2. Example Scenario:

  Multi-Gateway deployment with DP mode:
  - 2 Gateways (distributed)
  - 2 Pods, each with 4 API servers (ports 8000-8003)
  - Total 8 candidates across the 
  - Redis tracks request count for all 8 candidates
  - Each Gateway makes locally optimal decisions based on  state

  Performance Characteristics

  ┌──────────────────────────────┬─────────────────────────────────────────────┐
  │                        │                    Value                    │
  ├──────────────────────────────┼─────────────────────────────────────────────┤
  │ Time Complexity              │ O(1) - Two Redis                     │
  ├──────────────────────────────┼─────────────────────────────────────────────┤
  │ Space Complexity             │ O(P×N) - P pods × N ports per pod           │
  ├──────────────────────────────┼─────────────────────────────────────────────┤
  │ Redis Operations per Request │ 3 (2 GET + 1 INCR on route, 1 DECR on done) │
  ├──────────────────────────────┼─────────────────────────────────────────────┤
  │ Load Distribution            │ O(log log n) max load (proven by research)  │
  ├──────────────────────────────┼─────────────────────────────────────────────┤
  │ Key Rotation Interval        │ 3600 seconds (1 hour, hardcoded)            │
  ├──────────────────────────────┼─────────────────────────────────────────────┤
  │ Distributed Support          │ ✅ Yes                                      │
  ├──────────────────────────────┼─────────────────────────────────────────────┤
  │ Multi-Port Support           │ ✅ Yes                                      │
  └──────────────────────────────┴─────────────────────────────────────────────┘

  Comparison with Existing Routers

  ┌───────────────┬──────────────┬────────────┬─────────────┬────────────┐
  │    Router     │ Load Balance │ Multi-Port │ Distributed │ Complexity │
  ├───────────────┼──────────────┼────────────┼─────────────┼────────────┤
  │ random        │ ⭐           │ ❌         │ ✅          │ O(1)       │
  ├───────────────┼──────────────┼────────────┼─────────────┼────────────┤
  │ least-request │ ⭐⭐⭐⭐⭐   │ ✅         │ ❌          │ O(n)       │
  ├───────────────┼──────────────┼────────────┼─────────────┼────────────┤
  │ power-of-two  │ ⭐⭐⭐⭐     │ ✅         │ ✅          │ O(1)       │
  ├───────────────┼──────────────┼────────────┼─────────────┼────────────┤
  │ prefix-cache  │ ⭐⭐⭐       │ ❌         │ ❌          │ O(k)       │
  └───────────────┴──────────────┴────────────┴─────────────┴────────────┘

  Breaking Changes

  None. This PR is fully backward :
  - CanRegisterRequestTracker interface is -in
  - Existing routers continue to work unchanged
  - No changes to public APIs

  Files Changed

  - cmd/plugins/main.go (+3 lines)
  - pkg/cache/cache_api.go (+10 lines)
  - pkg/cache/cache_impl.go (+21 lines)
  - pkg/cache/cache_init.go (+3 lines)
  - pkg/plugins/gateway/algorithms/power_of_two.go (+293 lines, new file)
  - pkg/plugins/gateway/algorithms/power_of_two_test.go (+257 lines, new file)

### DIFF
--- a/pkg/plugins/gateway/algorithms/power_of_two.go
+++ b/pkg/plugins/gateway/algorithms/power_of_two.go
@@ -303,7 +303,7 @@ func (p *PowerOfTwoRouter) getRequestCountRedisKey(ctx *types.RoutingContext) st
 // AddRequestCount implements [cache.RequestTracker].
 func (p *PowerOfTwoRouter) AddRequestCount(ctx *types.RoutingContext, requestID string, modelName string) (traceTerm int64) {
 	// Check whether routing is done and target pod is set
-	if ctx.TargetPod() == nil {
+	if !ctx.HasRouted() {
 		return 0
 	}
 
@@ -356,7 +356,7 @@ func (p *PowerOfTwoRouter) AddRequestCount(ctx *types.RoutingContext, requestID 
 // DoneRequestCount implements [cache.RequestTracker].
 func (p *PowerOfTwoRouter) DoneRequestCount(ctx *types.RoutingContext, requestID string, modelName string, traceTerm int64) {
 	// Check whether target pod is set
-	if ctx.TargetPod() == nil {
+	if !ctx.HasRouted() {
 		return
 	}
 


Jeffwan:
image
seems unrelated.  Let's do the code review first

Jeffwan:
Let me take a look at it tomorrow

pkg/plugins/gateway/algorithms/power_of_two.go
```diff
@@ -0,0 +1,399 @@
...
+		return NewPowerOfTwoRouterWithRedis(redisClient)
+	})
+}
+
+type PowerOfTwoRouter struct {
+	redisClient    *redis.Client
+	keyRotationSec int64 // Time interval in seconds for key rotation, default 3600 (1 hour)
+
+	// lastRoutingTime is the timestamp of the last routing of a model. Used to avoid unnecessary redis request
+	lastRoutingTime       map[string]time.Time
```
Jeffwan, line 70: seem go map are not safe for concurrent read here?

pkg/plugins/gateway/algorithms/power_of_two.go
```diff
@@ -0,0 +1,399 @@
...
+		// No routing history for this model, skip counting
+		return 0
+	}
+
+	// Check whether it is called the first time
+	added := ctx.Value(po2RequestCountAddedKey)
+	if added != nil {
+		return 0
+	}
+	ctx.Context = context.WithValue(ctx.Context, po2RequestCountAddedKey, true)
```
Jeffwan, line 331: the routing context is pooled, mutating the context is a little bit risky here..

pkg/cache/cache_api.go
```diff
@@ -143,6 +144,15 @@ type RequestTracker interface {
 	DoneRequestTrace(ctx *types.RoutingContext, requestID string, modelName string, inputTokens, outputTokens, traceTerm int64)
 }
 
+// RequestTrackerRegistry can register multiple RequestTracker implementations
+// When Track operations are called, it will be passed to the registered RequestTracker
+type CanRegisterRequestTracker interface {
```
Jeffwan: can you rename it to `RequestTrackerRegistry` or something different? `Canxxx` is not good convention

pkg/plugins/gateway/algorithms/power_of_two.go
penfree: good point, i'll replace it with sync.Map

pkg/cache/cache_api.go
penfree: ok

pkg/plugins/gateway/algorithms/power_of_two.go
penfree: ok，i'll add mutex around it

pkg/plugins/gateway/algorithms/power_of_two.go
penfree: seems Context is reset for every request, so i think it's not a problem?

func NewRoutingContext(ctx context.Context, algorithms RoutingAlgorithm, model, message, requestID, user string) *RoutingContext {
	request := requestPool.Get().(*RoutingContext)
	request.reset(ctx, algorithms, model, message, requestID, user)
	return request
}

penfree: @Jeffwan i've fix other comments as you suggested, pls take a look at this

Jeffwan: sounds good. let me take a look. I think it should not be an issue

pkg/plugins/gateway/algorithms/power_of_two.go
```diff
@@ -0,0 +1,407 @@
...
+		pod:  ctx.TargetPod(),
+		port: port,
+	}
+	return p.buildRedisKey(ctx.Model, serverKey, ctx.RequestTime)
+}
+
+// AddRequestCount implements [cache.RequestTracker].
+func (p *PowerOfTwoRouter) AddRequestCount(ctx *types.RoutingContext, requestID string, modelName string) (traceTerm int64) {
+	// Check whether routing is done and target pod is set
+	if ctx.TargetPod() == nil {
```
Jeffwan: the failure test is from here. TargetPod() is a blocking call, not a simple nil check. Per its documentation: "TargetPod blocks until the target pod is set or an error is set."

this leads to gateway request timeout? Did you find this issue in your local testing?

pkg/plugins/gateway/algorithms/power_of_two.go
```diff
@@ -0,0 +1,407 @@
...
+		"new_count", newCount)
+
+	// Use current count as trace term
+	return newCount
+}
+
+// DoneRequestCount implements [cache.RequestTracker].
+func (p *PowerOfTwoRouter) DoneRequestCount(ctx *types.RoutingContext, requestID string, modelName string, traceTerm int64) {
+	// Check whether target pod is set
+	if ctx.TargetPod() == nil {
```
Jeffwan: same here.  use `!ctx.HasRouted()`

pkg/plugins/gateway/algorithms/power_of_two.go
Jeffwan: use `!ctx.HasRouted()` instead

Jeffwan:
@penfree please check my comments here.

Jeffwan:
BTW, there are some design pattern issues. 

Registering a router as a global RequestTracker that gets called for every request is problematic, not a good pattern. while, this probably need some design refactor and not in this PR's scope.

@penfree please help fix the issue and we can merge this PR. the failing integration test comes from some misuse issue. check my comments please

pkg/plugins/gateway/algorithms/power_of_two.go
```diff
@@ -0,0 +1,407 @@
...
+var po2RequestCountAddedKey = po2RequestCountAddedKeyStruct{}
+var po2RequestCountDoneKey = po2RequestCountDoneKeyStruct{}
+
+/*
+redis key design:
+
+po2_req_count:<modelName>:<podName>_<port>:timestamp -> realtime running request
+*/
+
+const (
```
Jeffwan, line 47: can you also check other algorithms, we need to `Register()` your algorithms when it starts up

pkg/plugins/gateway/algorithms/power_of_two.go
penfree: ok, it does not happen in my localtest, maybe AddRequestCount happened to be called before routing in some cases.

pkg/plugins/gateway/algorithms/power_of_two.go
penfree: since redisClient is needed by power_of_two， so it is registered in main.go.  do you have a better suggestion

penfree:
> BTW, there are some design pattern issues.
> 
> Registering a router as a global RequestTracker that gets called for every request is problematic, not a good pattern. while, this probably need some design refactor and not in this PR's scope.
> 
> @penfree please help fix the issue and we can merge this PR. the failing integration test comes from some misuse issue. check my comments please

@Jeffwan 
i think it's necessary， because routing strategy can be set by http header.  so requests of the same model can use different routing strategy. if RequestTracker is not called for every requests, it may not have accurate metrics。

to mitigate this problem,  i added lastRoutingTime to trace the last call of current router, and stop doing RequestTrack if it's not been called for a long time.

pkg/plugins/gateway/algorithms/power_of_two.go
Jeffwan: right, if other issues are addressed, we can merge this one. I can do some minor refactor to make pattern consistent

PR #2032, opened 2026-03-20
Multi-strategy routing support

DaveLi8086:
Content moved to PR 2124 to fix DCO issue

## Summary
This PR adds **multi-strategy routing** to AIBrix Gateway using a **batch soft-scoring** pipeline. Multiple routing algorithms can be combined in a single request/config string with integer **weight coefficients**. Each strategy scores all ready pods in one batch (`ScoreAll`), the router normalizes scores to `[0,1]` based on polarity, aggregates by weight, and picks the top pod.

## Motivation
The existing router selects a pod using a single strategy at a time. In practice we often want to blend multiple signals (e.g. load + latency + cache affinity) without “hard filtering” pods early. Soft scoring keeps the full candidate set and makes routing decisions smoother.

## User-facing configuration
Multi-strategy is configured via the existing header/env value (comma-separated list of `strategy[:coefficient]`):

- Header: `routing-strategy: least-request:2,throughput:1`
- Env: `AIBRIX_ROUTING_ALGORITHM="least-request:2,throughput:1"`

Rules:
- the weight `coefficient` is **integer only** in `[0, 1000000]` (following [gateway-api](https://gateway-api.sigs.k8s.io/reference/spec/#backendref))
- omitted coefficient defaults to `1`
- coefficient `0` means the strategy is skipped
- If routign strategy: [a:2,b,c], then final weights are **[0.5, 0.25, 0.25]** after normalization

## Design overview (Batch Soft Scoring)
Pipeline:
1. **Parse config** → `[]RouterItem{Name,Coefficient}`
2. **Batch score** each strategy over the same `readyPodList`
3. **Normalize** each strategy’s raw scores to `[0,1]` using strategy polarity
4. **Weighted aggregation** using `coefficient / sum(coefficients)`
5. **Pick top-1** (deterministic tie-break)

### Handling missing metrics
If a pod cannot be scored for a strategy (`scored=false` or `NaN`), its normalized score for that strategy is **0.0** (worst score).

## Unit Test and E2E Test Added
Extensive unit tests and E2E tests have been added to ensure the reliability and correctness of the multi-strategy routing architecture:
- **Unit Tests (`router_test.go`)**:
  - Validated single strategies (both `HigherIsBetter` and `LowerIsBetter` polarities).
  - Validated multi-strategy weighted aggregation with varying coefficients.
  - Handled missing metrics (`scored=false` or `NaN`) gracefully, assigning a normalized score of `0.0` without discarding the pod.
  - Handled edge cases like `max == min` scores (assigning a normalized score of `1.0` to all scored pods).
- **E2E Tests (`routing_strategy_test.go`)**:
  - `ValidMultiStrategy_Partial_Weights`: Tested partial coefficient configs (e.g., `least-request,throughput:2`).
  - `ValidMultiStrategy_No_Weights`: Tested configs with omitted coefficients, defaulting to `1`.
  - `ExclusiveStrategy_FallbackToSelf_*`: Ensured exclusive strategies (like `pd` and `slo-*`) safely strip out unsupported strategies and fallback to their own legacy paths.
  - `InvalidStrategy_FallbackToRandom`: Verified that invalid or unregistered strategies gracefully degrade to `random` routing, preventing 500 errors.
- **Diagnostic Logging (`klog`)**: Added detailed, tabular `request_start` logging that breaks down the raw score, normalized score, and weighted contribution of every candidate pod for each strategy, making debugging transparent.

## Notes
- Single-strategy behavior is preserved: when the config resolves to a single strategy name, gateway routing follows the existing path.
- Multi-strategy pick tie-break is deterministic (stable w.r.t. original pod order), while some single-strategy routers may still use randomized tie-break.


pkg/plugins/gateway/algorithms/router.go
```diff
@@ -31,13 +35,275 @@ const (
...
+			klog.Infof("Exclusive routing strategy '%s' found in multi-strategy config. Other strategies will be ignored.", item.Name)
+			return &MultiRouterConfig{Items: []RouterItem{{Name: item.Name, Coefficient: 1}}}, nil
+		}
+	}
+
+	return &MultiRouterConfig{Items: items}, nil
+}
+
+// multiStrategyRouter coordinates multiple sub-routers and selects a pod via soft-scoring
+type multiStrategyRouter struct {
```
varungup90, line 140: do we need to create multiStrategyRouter? can it encapsulate current router.

- I prefer not to distinguish between routers for single strategy and multi-strategy.
- For single strategy, weight can be just defaulted to 1

pkg/plugins/gateway/algorithms/router.go
```diff
@@ -31,13 +35,275 @@ const (
...
+}
+
+// MultiRouterConfig holds the parsed routing algorithms and their weight coefficients.
+type MultiRouterConfig struct {
+	Items []RouterItem
+}
+
+// ParseMultiRouterConfig parses a multi-router string into a MultiRouterConfig.
+// Format example: "prefix-cache:2,least-latency:1,least-request"
+// - Weight coefficients must be integers in range [0, 1000000].
```
varungup90, line 71: I will be easier to reason if weights add up to 1, each routing strategy has a weight between 0 to 1.

If user provides weight where total is more than 1, lets handle the scenario and not reject it, but lets keep weights to add up to 1.

pkg/plugins/gateway/algorithms/router.go
```diff
@@ -71,6 +348,22 @@ func Validate(algorithms string) (types.RoutingAlgorithm, bool) {
...
+	
+	// Check if it's a multi-strategy (contains ',' or ':')
+	if strings.Contains(algStr, ",") || strings.Contains(algStr, ":") {
+		cfg, err := ParseMultiRouterConfig(algStr)
+		if err == nil {
+			multiRouter, err := newMultiStrategyRouter(cfg, rm, ctx)
+			if err == nil {
+				return multiRouter, nil
+			}
+			klog.Warningf("Failed to initialize multi-strategy router for %s: %v, fallback to %s", algStr, err, RouterRandom)
```
varungup90: ctx has a field RequestID, include that in log message without that it is not debug-able.

pkg/plugins/gateway/algorithms/router.go
```diff
@@ -31,13 +35,275 @@ const (
...
+	}
+
+	return &multiStrategyRouter{
+		config:  config,
+		scorers: scorers,
+	}, nil
+}
+
+// Route executes the multi-strategy scoring pipeline and returns the best pod IP
+func (m *multiStrategyRouter) Route(ctx *types.RoutingContext, readyPodList types.PodList) (string, error) {
```
varungup90, line 177: With this change, run benchmarks to compare on prefix-cache vs prefix-cache:0.75,least-request:0.125,throughput:0.125

pkg/plugins/gateway/algorithms/router.go
DaveLi8086: Yes, the actually weight in this design is still [0,1]. 

The reason using this `weight coefficient` as wrapper to the weight, is to follow your suggestion and support more cases when user forgets to set weights to some strategies.

i.e `a:0.2,b,c`, it becomes ambiguous how to handle omitted weights of b/c, as we don't know what user want the ratio b/c to be (and also we don't know if user was planning to make the total sum > 1 if we allow it)

To solve it in an intuitive way, gateway-api/llm-d choose to use weight coefficient, and the default coefficient is `1`:

i.e `a:2,b,c`, in this case it implies `a:2,b:1,c:1. After normalization, actual weights of a/b/c are [0,5,0,25,0,25]
(if all weights are omitted, they will share same weight). Integer coefficients let us apply a clean default rule that is predictable.

pkg/plugins/gateway/algorithms/throughput.go
```diff
@@ -49,39 +50,70 @@ func NewThroughputRouter() (types.Router, error) {
...
+		klog.V(4).Infof("pod: %v, podIP: %v, totalThroughput: %v",
+			pod.Name, pod.Status.PodIP, scores[i])
+
+		if scores[i] < minCount {
+			minCount = scores[i]
+			targetPods = []string{pod.Name}
+		} else if scores[i] == minCount {
+			targetPods = append(targetPods, pod.Name)
 		}
 	}
```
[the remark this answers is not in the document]
DaveLi8086: According to the definition listed in [here](https://aibrix.readthedocs.io/latest/designs/aibrix-router.html#supported-routing-strategies), the `throughput` strategy means to route request to a pod which has processed the `lowest total weighted tokens`. Based on that, the polarity should be `PolarityLeast`, which means lower accumulated throughput is better. The previous unit test was incorrect on that, and got updated in this [commit](https://github.com/vllm-project/aibrix/pull/2032/changes/2e0ef293e40142af1bc6d49f8c4a78158d25ab60#diff-f441978a2571e2585c18b807c75f9d73f1678b4d63e19022c0a1ad97f2b55644)

pkg/plugins/gateway/algorithms/README.md
```diff
@@ -184,3 +184,65 @@ func doPrefillRequest(prefillPods) {
...
+* **Pod A**: High load (10 requests) but High throughput (100 tok/s).
+* **Pod B**: Low load (2 requests) but Low throughput (10 tok/s).
+
+#### Example 1: `least-request` Dominates
+```yaml
+AIBRIX_ROUTING_ALGORITHM: "least-request:10,throughput:1"
+```
+* `least-request` (weight 10): Pod B is much better (2 vs 10). Normalized: Pod A = 0.0, Pod B = 1.0. Score contribution: A = 0, B = 10.
+* `throughput` (weight 1): Pod A is much better (100 vs 10). Normalized: Pod A = 1.0, Pod B = 0.0. Score contribution: A = 1, B = 0.
+* **Total Score**: Pod A = 1, Pod B = 10. **Pod B wins!**
```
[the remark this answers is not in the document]
DaveLi8086: This has been fixed in the latest[ README.md](https://github.com/DaveLi8086/aibrix/blob/feat/multi-filter-g/pkg/plugins/gateway/algorithms/README.md).

pkg/plugins/gateway/algorithms/router.go
DaveLi8086: Yes, `ctx.RequestID` has been added here in this [commit](https://github.com/vllm-project/aibrix/pull/2032/changes/abb8525de75e81788bcee57feb6c6986fb071fd3#diff-2dc37dc9539c9d27bea5823c7812eb58bcce1e1887b3cebc7f2673847263ac2c)

Issue #2050, opened 2026-03-27
Make Redis optional in local mode


jasonlee-1024:
## Description
Redis should be not required when starting AIBrix in local mode. This introduces unnecessary dependency for local development and testing.

## Problem
When running AIBrix in local mode without Redis, the system fails to start due to missing Redis dependency.

## Proposed Change
- Allow Redis initialization to fail gracefully in local mode without blocking system startup  
- Add fallback logic for components that depend on Redis to ensure core functionalities remain available  

## Impact
- Simplifies local development setup  
- Reduces external dependencies for testing  
- Improves developer experience  

## Status
We are working on a fix and will submit a PR shortly.

jasonlee-1024:
I've created the issue about makeing redis optional. @Lucas-Qian6 @Jeffwan

Jeffwan:
Thanks for driving the efforts.

PR #2058, opened 2026-03-31
[Feat] Make Redis optional in local mode

Lucas-Qian6:
## Pull Request Description

Make Redis a soft dependency so the gateway can start and serve requests
without a running Redis server. This is needed for local development/testing
where Redis adds unnecessary setup friction.

### Changes

- **`pkg/utils/redis.go`**: Replace `klog.Fatalf` with `klog.Warningf` and
  return `nil` when Redis is unreachable, instead of crashing the process.
- **`cmd/plugins/main.go`**: Handle `nil` Redis client gracefully at startup;
  skip defer close and log a warning that rate limiting and user auth are disabled.
- **`pkg/plugins/gateway/gateway.go`**: Use a `NoopRateLimiter` when Redis
  client is nil, instead of creating a `RedisAccountRateLimiter` that would panic.
- **`pkg/plugins/gateway/gateway_req_headers.go`**: Skip user lookup via Redis
  when `redisClient` is nil.
- **`pkg/plugins/gateway/ratelimiter/noop.go`** (new): Add a no-op
  `RateLimiter` implementation that allows all requests through.

### Behavior

- **With Redis**: No change — everything works as before (rate limiting, user auth, po2 routing).
- **Without Redis**: Gateway starts normally; rate limiting and user auth are
  disabled.

## Related Issues
Resolves: #2050

### DIFF
--- a/cmd/plugins/main.go
+++ b/cmd/plugins/main.go
@@ -77,11 +77,19 @@ func main() {
 	}
 
 	redisClient := utils.GetRedisClient()
-	defer func() {
-		if err := redisClient.Close(); err != nil {
-			klog.Warningf("Error closing Redis client: %v", err)
+	if redisClient == nil {
+		if standalone {
+			klog.Warning("Running without Redis: rate limiting and user auth disabled")
+		} else {
+			klog.Fatal("Redis is required in Kubernetes mode")
 		}
-	}()
+	} else {
+		defer func() {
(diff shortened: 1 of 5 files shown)


Lucas-Qian6:
@Jeffwan Please take a look when you have time. Thanks.

Jeffwan (approved):
overall looks good to me.

PR #2089, opened 2026-04-08
Feature/prefix cache with redis reqcnt

penfree:
## Summary

  This PR includes critical bugfixes for KV sync prefix cache routing and adds
  Redis-based request counting support for multi-instance gateway deployments.

  ## Changes

  ### ✨ Feature: Redis-Based Request Counter for Multi-Instance Gateway Deployments

  **Motivation:**
  When multiple AIBrix gateway instances are deployed, each instance only tracks local
  request counts. This causes incorrect load imbalance detection since no single
  instance has the complete view of pod request distribution.

  **Implementation:**
  - Added `RequestCounter` interface with two implementations:
    - `localRequestCounter`: Uses local metrics (existing behavior)
    - `redisRequestCounter`: Uses Redis hash to track distributed request counts
  - Modified prefix cache router to accept request counter via dependency injection
  - Refactored load balancing functions to accept `map[string]int` instead of
  `cache.Cache`
  - Environment variable `AIBRIX_PREFIX_CACHE_USE_REDIS_REQCNT=true` enables
  Redis-based counting

  **Key Features:**
  - Redis key design: `aibrix:prefix-cache-reqcnt:{model_name}` (hashmap: pod_key →
  count)
  - Supports Redis Cluster via hashtag `{model_name}` for same-slot operations
  - Automatic cleanup when request count reaches zero
  - Timeout mechanism to stop tracking inactive models (5 minutes)
  - Returns all ready pods with count=0 for pods not in Redis

  **Files Changed:**
  - `pkg/plugins/gateway/algorithms/request_counter.go` (new file): RequestCounter
  interface and implementations
  - `pkg/plugins/gateway/algorithms/prefix_cache.go`: Integration with request counter
  - `pkg/plugins/gateway/algorithms/power_of_two.go`: Updated `podServerKey.String()`
  to include namespace

  ## Testing

  - ✅ All existing tests pass
  - ✅ Updated test fixtures to use `requestCounter`
  - ✅ Fixed test expectations for namespace-aware pod keys

  ## Breaking Changes

  None. Changes are backward compatible:
  - Default behavior remains unchanged (uses local request counter)
  - Redis-based counting is opt-in via environment variable

  ## Related Issues

  Closes: [Issue number if applicable]

  ## Deployment Notes

  To enable Redis-based request counting:
  1. Ensure Redis client is configured
  2. Set environment variable: `AIBRIX_PREFIX_CACHE_USE_REDIS_REQCNT=true`
  3. Multiple gateway instances will now share request count information

### DIFF
--- a/pkg/plugins/gateway/algorithms/prefix_cache.go
+++ b/pkg/plugins/gateway/algorithms/prefix_cache.go
@@ -22,7 +22,6 @@ import (
 	"fmt"
 	"math"
 	"math/rand"
-	"os"
 	"sort"
 	"strconv"
 	"sync"
@@ -171,8 +170,8 @@ func RegisterPrefixCacheRouterWithRedis(redisClient *redis.Client) {
 // kvSyncPrefixCacheRouter handles routing when KV sync is enabled
 type kvSyncPrefixCacheRouter struct {
 	cache          cache.Cache
-	tokenizerPool  TokenizerPoolInterface // Add TokenizerPool reference
-	syncIndexer    *syncindexer.SyncPrefixHashTable
+	tokenizerPool  TokenizerPoolInterface           // Add TokenizerPool reference
+	syncIndexer    *syncindexer.SyncPrefixHashTable // Use concrete type
 	metricsEnabled bool
 	requestCounter RequestCounter // Request counter for load balancing
 }
@@ -290,10 +289,10 @@ func NewPrefixCacheRouterWithRedis(redisClient *redis.Client) (types.Router, err
 
 	// Initialize request counter
 	var requestCounter RequestCounter
-	useRedisReqCnt := redisClient != nil && os.Getenv("AIBRIX_PREFIX_CACHE_USE_REDIS_REQCNT") == "true"
+	useRedisReqCnt := redisClient != nil && utils.LoadEnvBool(constants.EnvPrefixCacheUseRedisReqCnt, false)
 	if useRedisReqCnt {
 		// Use Redis-based request counter
-		requestCounter = NewRedisRequestCounter(redisClient, podRunningRequestImbalanceAbsCount)
+		requestCounter = NewRedisRequestCounter(redisClient)
 		klog.Info("Using Redis-based request counter for prefix cache router")
 	} else {
 		// Use local metrics-based request counter
@@ -326,22 +325,36 @@ func NewPrefixCacheRouterWithRedis(redisClient *redis.Client) (types.Router, err
 
 	// Only create KV sync router if enabled
 	if kvSyncEnabled && useRemoteTokenizer && tokenizerPool != nil {
-		kvSyncRouter := &kvSyncPrefixCacheRouter{
-			cache:          c,
-			tokenizerPool:  tokenizerPool, // Pass the pool reference
-			syncIndexer:    syncindexer.GetSharedSyncPrefixHashTable(),
-			metricsEnabled: true,
-			requestCounter: requestCounter, // Use the same request counter
-		}
+		// Get the shared sync indexer instance - this is the same instance
+		// that receives KV events from vLLM pods via the KV event subscriber
+		syncIdx := syncindexer.GetSharedSyncPrefixHashTable()
+		if syncIdx == nil {
+			klog.Warning("KV sync enabled but sync indexer not initialized, falling back to local indexer")
+			// Update metrics to reflect fallback to local indexer
+			if metrics := getPrefixCacheMetrics(); metrics != nil {
+				metrics.prefixCacheIndexerStatus.WithLabelValues("", "local").Set(1)
+				metrics.prefixCacheIndexerStatus.WithLabelValues("", "sync").Set(0)
+			}
+		} else {
+			kvSyncRouter := &kvSyncPrefixCacheRouter{
+				cache:          c,
+				tokenizerPool:  tokenizerPool, // Pass the pool reference
+				syncIndexer:    syncIdx,       // Use the shared sync indexer
+				metricsEnabled: true,
+				requestCounter: requestCounter, // Use the same request counter
+			}
 
-		router.kvSyncRouter = kvSyncRouter
+			router.kvSyncRouter = kvSyncRouter
 
-		// Set initial metrics only if KV sync is enabled
-		if metrics := getPrefixCacheMetrics(); metrics != nil {
-			metrics.prefixCacheIndexerStatus.WithLabelValues("", "sync").Set(1)
-			metrics.prefixCacheIndexerStatus.WithLabelValues("", "local").Set(0)
+			// Set initial metrics only if KV sync is enabled
+			if metrics := getPrefixCacheMetrics(); metrics != nil {
+				metrics.prefixCacheIndexerStatus.WithLabelValues("", "sync").Set(1)
+				metrics.prefixCacheIndexerStatus.WithLabelValues("", "local").Set(0)
+			}
 		}
-	} else {
+	}
+
+	if router.kvSyncRouter == nil {
 		// Set local indexer metrics only if metrics are enabled
 		metricsEnabled := utils.LoadEnvBool(constants.EnvPrefixCacheLocalRouterMetricsEnabled, false)
 		if metricsEnabled {
@@ -403,11 +416,13 @@ func (p prefixCacheRouter) routeOriginal(ctx *types.RoutingContext, readyPodList
 	readyPods := readyPodList.All()
 	readyPodsMap := map[string]struct{}{}
 	for _, pod := range readyPods {
-		readyPodsMap[pod.Name] = struct{}{}
+		podKey := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+		readyPodsMap[podKey] = struct{}{}
 	}
 
 	// Get request counts once for consistency across all subsequent operations
-	podRequestCount := p.requestCounter.GetRequestCounts(ctx.Context, readyPods, false)
+	// Pass ctx directly so RequestTime can be extracted for key rotation
+	podRequestCount := p.requestCounter.GetRequestCounts(ctx, readyPods)
 
 	leastReqPodList, isLoadImbalanced := getTargetPodListOnLoadImbalance(podRequestCount, readyPods)
 	if isLoadImbalanced {
@@ -420,7 +435,8 @@ func (p prefixCacheRouter) routeOriginal(ctx *types.RoutingContext, readyPodList
 		readyPodsMap = map[string]struct{}{}
 		// filter the readyPodsMap by leastReqPodList used by below codes
 		for _, pod := range leastReqPodList {
-			readyPodsMap[pod.Name] = struct{}{}
+			podKey := fmt.Sprintf("%s/%s", pod.Namespace, pod.Name)
+			readyPodsMap[podKey] = struct{}{}
 		}
 		klog.V(4).InfoS("prefix_cache_load_imbalanced",
 			"request_id", ctx.RequestID,
@@ -432,7 +448,7 @@ func (p prefixCacheRouter) routeOriginal(ctx *types.RoutingContext, readyPodList
 	klog.V(4).InfoS("prefix_hashes", "request_id", ctx.RequestID, "prefix_hashes", prefixHashes)
 
 	if len(matchedPods) > 0 {
-		targetPod = getTargetPodFromMatchedPods(podRequestCount, readyPods, matchedPods)
+		targetPod = getTargetPodFromMatchedPodsWithKeys(podRequestCount, readyPods, matchedPods)
 		if targetPod != nil {
 			klog.InfoS("prefix_cache_matched_pods",
 				"request_id", ctx.RequestID,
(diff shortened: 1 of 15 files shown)


penfree:
@Jeffwan hi, please review this pr

varungup90:
@penfree Thank you for the PR, is it possible to split the PR into two. Remove the redis counter into different PR, also I am started a PR to sync in-memory state across gateway instances (https://github.com/vllm-project/aibrix/pull/1989) -- in that PR, it syncs in-memory prefix cache stats, but can be extended to share request counter as well.

varungup90:
Summary
This PR fixes a significant bug in kvSyncPrefixCacheRouter (it was querying an empty indexer instead of the shared one from the cache store) and adds a RequestCounter interface with Redis-backed distributed request counting across multiple gateway instances.

Critical Issues
1. KV sync key format mismatch — [pkg/plugins/gateway/algorithms/prefix_cache.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/prefix_cache.go) ~line 715

GetRequestCounts() now returns keys in namespace/name format, but the loop that builds podRequestCountWithKeys looks up podRequestCount[pod.Name] (bare name). This silently fails — every lookup misses, podRequestCountWithKeys is always empty, and the router falls back to least-request selection with no valid input.

2. filterPodsByKey is dead code — [pkg/plugins/gateway/algorithms/prefix_cache.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/prefix_cache.go) ~lines 635–643

The function is defined but never called. This suggests the refactoring was left incomplete. Go won't fail to compile on an unused function, but the routing code that should call it likely has the bug described in #1.

3. KV sync nil-indexer fallback doesn't update metrics — [pkg/plugins/gateway/algorithms/prefix_cache.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/prefix_cache.go) ~lines 254–285

When c.GetSyncPrefixIndexer() returns nil, the code logs a warning and skips creating kvSyncRouter, but prefixCacheIndexerStatus metrics are never set. The fallback to local-indexer behavior goes unrecorded.

High Priority
4. Breaking Redis key format change for Power-of-Two router — [pkg/plugins/gateway/algorithms/utils.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/utils.go), [power_of_two.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/power_of_two.go)

podServerKey was moved to utils.go and its format changed from name_port → namespace/name/port. Existing in-flight Redis keys under the old format will be silently ignored on deploy, causing a transient load spike. This needs deployment notes.

5. lastModelRouteTime map grows unbounded — [pkg/plugins/gateway/algorithms/request_counter.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/request_counter.go) ~line 1291

Updated on every GetRequestCountsWithPort call but never pruned. Long-running instances routing many different models will leak memory.

Medium Priority
6. Lua script re-evaluated on every call — [request_counter.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/request_counter.go) ~line 1521

The Lua decrement script is a string literal inside DoneRequestCount(). In high-traffic environments, define it as a package-level redis.NewScript(...) and use EvalSha for Redis script caching.

7. Expire resets TTL on every increment — [request_counter.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/plugins/gateway/algorithms/request_counter.go) ~line 1467

A single ongoing request continuously refreshes the TTL. Consider ExpireNX (set only if no TTL exists) or ExpireGT to avoid preventing cleanup of active keys.

8. SyncIndexerProvider returns a concrete type — [pkg/cache/cache_api.go](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/pkg/cache/cache_api.go) ~line 71

GetSyncPrefixIndexer() *syncindexer.SyncPrefixHashTable tightly couples the interface to the implementation. An interface type would enable test injection and better separation of concerns.

9. redismock in production require block — [go.mod](vscode-webview://0kdgpvfq859mk3f0rvkgmj7it0r8i79qht6i3eet0akp78bl0llg/go.mod)

github.com/go-redis/redismock/v9 is a test-only library but placed in the main require block, adding it as a production binary dependency.

penfree:
@varungup90 Thanks for your patience。

> Summary This PR fixes a significant bug in kvSyncPrefixCacheRouter (it was querying an empty indexer instead of the shared one from the cache store) and adds a RequestCounter interface with Redis-backed distributed request counting across multiple gateway instances.
> 
> Critical Issues
> 
> 1. KV sync key format mismatch — pkg/plugins/gateway/algorithms/prefix_cache.go ~line 715
> 
> GetRequestCounts() now returns keys in namespace/name format, but the loop that builds podRequestCountWithKeys looks up podRequestCount[pod.Name] (bare name). This silently fails — every lookup misses, podRequestCountWithKeys is always empty, and the router falls back to least-request selection with no valid input.
> 

**i haven't found any code that used bare pod name to look up podRequestCount**

```bash
(.venv) ➜  pkg git:(feature/prefix-cache-with-redis-reqcnt) ack podRequestCount | grep Name           
plugins/gateway/algorithms/pd_disaggregation_benchmark_test.go:122:                             podRequestCounts[pod.Name] = requestCount
plugins/gateway/algorithms/request_counter.go:137:                      podRequestCount[keyName] = count
plugins/gateway/algorithms/prefix_cache.go:719:         // For KV sync router, we need podRequestCount to use pod keys (namespace/podName)
plugins/gateway/algorithms/pd_disaggregation.go:425:            podRequestCounts[pod.Name] = requestCount
plugins/gateway/algorithms/pd_disaggregation.go:483:            score := podRequestCounts[pod.Name] / math.Max(drainRate.GetSimpleValue(), defaultDrainRateEpsilon)
plugins/gateway/algorithms/pd_disaggregation.go:618:            reqCnt := float64(podRequestCount[pod.Name])
plugins/gateway/algorithms/pd_disaggregation.go:650:            normalizedRunningReqs := podRequestCounts[pod.Name] / maxRequestCount
plugins/gateway/algorithms/pd_disaggregation.go:667:                    "running_reqs", podRequestCounts[pod.Name], "max_running_reqs", maxRequestCount,
plugins/gateway/algorithms/pd_disaggregation.go:1168:   countInterface, _ := t.podRequestCounts.LoadOrStore(podName, &atomic.Int32{})
plugins/gateway/algorithms/pd_disaggregation.go:1191:   countInterface, exists := t.podRequestCounts.Load(podName)
plugins/gateway/algorithms/pd_disaggregation.go:1215:           countInterface, exists := t.podRequestCounts.Load(pod.Name)
plugins/gateway/algorithms/pd_disaggregation.go:1242:   countInterface, _ := t.podRequestCounts.LoadOrStore(podName, &atomic.Int32{})
plugins/gateway/algorithms/pd_disaggregation.go:1257:   countInterface, exists := t.podRequestCounts.Load(podName)
plugins/gateway/algorithms/pd_disaggregation.go:1281:   countInterface, exists := t.podRequestCounts.Load(podName)
plugins/gateway/algorithms/least_request.go:186:                podRequestCount[pod.Name] = int(runningReq.GetSimpleValue())
plugins/gateway/algorithms/least_request.go:217:                        podRequestCount[keyName] = count
```

> 2. filterPodsByKey is dead code — pkg/plugins/gateway/algorithms/prefix_cache.go ~lines 635–643
> 
> The function is defined but never called. This suggests the refactoring was left incomplete. Go won't fail to compile on an unused function, but the routing code that should call it likely has the bug described in #1.
> 

**removed**

> 3. KV sync nil-indexer fallback doesn't update metrics — pkg/plugins/gateway/algorithms/prefix_cache.go ~lines 254–285
> 
> When c.GetSyncPrefixIndexer() returns nil, the code logs a warning and skips creating kvSyncRouter, but prefixCacheIndexerStatus metrics are never set. The fallback to local-indexer behavior goes unrecorded.
> 

**Fixed**

> High Priority 4. Breaking Redis key format change for Power-of-Two router — pkg/plugins/gateway/algorithms/utils.go, power_of_two.go
> 
> podServerKey was moved to utils.go and its format changed from name_port → namespace/name/port. Existing in-flight Redis keys under the old format will be silently ignored on deploy, causing a transient load spike. This needs deployment notes.
> 

**Since redis key in power of two will rotate every hour, i think this is not a problem**

> 5. lastModelRouteTime map grows unbounded — pkg/plugins/gateway/algorithms/request_counter.go ~line 1291
> 
> Updated on every GetRequestCountsWithPort call but never pruned. Long-running instances routing many different models will leak memory.
> 

**the max size of this map is the amount of model, i don't think there can have so many models that cause a memory issue, this every entry has only a string and a int64. however, pruning it needs many code**


> Medium Priority 6. Lua script re-evaluated on every call — request_counter.go ~line 1521
> 
> The Lua decrement script is a string literal inside DoneRequestCount(). In high-traffic environments, define it as a package-level redis.NewScript(...) and use EvalSha for Redis script caching.
> 

**Fixed**

> 7. Expire resets TTL on every increment — request_counter.go ~line 1467
> 
> A single ongoing request continuously refreshes the TTL. Consider ExpireNX (set only if no TTL exists) or ExpireGT to avoid preventing cleanup of active keys.
> 

**i've added key rotation, and set ttl in lua only when the key does not exist**

> 8. SyncIndexerProvider returns a concrete type — pkg/cache/cache_api.go ~line 71
> 
> GetSyncPrefixIndexer() *syncindexer.SyncPrefixHashTable tightly couples the interface to the implementation. An interface type would enable test injection and better separation of concerns.
> 

**like PrefixHashTable,  a GetSharedSyncPrefixHashTable is added to get singleton of SyncPrefixHashTable, GetSyncPrefixIndexer is removed from cache_api interface.  

> 9. redismock in production require block — go.mod
> 
> github.com/go-redis/redismock/v9 is a test-only library but placed in the main require block, adding it as a production binary dependency.

**seems this is not an issue, the redismock is only used in test, so it's not built into the production binary**

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -155,15 +156,24 @@ func getPrefixCacheMetrics() *PrefixCacheMetrics {
...
 
 // kvSyncPrefixCacheRouter handles routing when KV sync is enabled
 type kvSyncPrefixCacheRouter struct {
 	cache          cache.Cache
-	tokenizerPool  TokenizerPoolInterface // Add TokenizerPool reference
-	syncIndexer    *syncindexer.SyncPrefixHashTable
+	tokenizerPool  TokenizerPoolInterface           // Add TokenizerPool reference
+	syncIndexer    *syncindexer.SyncPrefixHashTable // Use concrete type
 	metricsEnabled bool
+	requestCounter RequestCounter // Request counter for load balancing
```
varungup90, line 176: Local counter is not needed. This local counter is only incrementing requests for prefix-cache, compare to RealtimeNumRequestsRunning keep track of requests across all routing policies. 

RealtimeNumRequestsRunning: is calculated via gateway and it is not fetched from the pod.

```
metricName = metrics.RealtimeNumRequestsRunning
if val, err := cache.GetMetricValueByPod(pod.Name, pod.Namespace, metricName); err == nil && val != nil {
				count = int(val.GetSimpleValue())
			}
```

pkg/plugins/gateway/algorithms/prefix_cache.go
```diff
@@ -176,6 +186,9 @@ type prefixCacheRouter struct {
 
 	// KV sync router - only created when needed
 	kvSyncRouter *kvSyncPrefixCacheRouter
+
+	// Request counter for load balancing
+	requestCounter RequestCounter
```
varungup90, line 191: Lets simplify the PR, split the request counter into a different PR. This one to be more focussed on fixing bug with kv event sync.

pkg/plugins/gateway/algorithms/prefix_cache.go
penfree: @varungup90  i don't think so, localCounter implements the same interface as redisCounter, so kvSyncPrefixCacheRouter can use either of them as user's need.  
When only one gateway  instance is deployed， they don't need redisCounter with extra latency cost.  localCounter can be a dropin replacement in such cases, without using many local util functions.

pkg/plugins/gateway/algorithms/prefix_cache.go
penfree: @varungup90 already split the bugfix to #2096 , i'll rebase this pr to main branch after it's merged

penfree: @varungup90  rebased,  PTAL

pkg/plugins/gateway/algorithms/prefix_cache.go
varungup90: How to extend the counter to all the routing algorithms?

penfree: > How to extend the counter to all the routing algorithms?

@varungup90 different algorithms may record the counter in different ways, for example:
1) power of two router:  just need a key for each pod to record the count, get by key and compare them.
2) prefix cache:  the current algorithym needs to fetch the request counts from all pods, so a hashmap is needed to store them.  

so maybe they should have diffent implementations.

varungup90:
Addressed by this PR: https://github.com/vllm-project/aibrix/pull/2159

PR #2108, opened 2026-04-14
 [Bug] Modify the way container environment variables are rendered

### LINKED ISSUE #2107: The environment variables defined as interdependent in a role do not take effect inside the container.
### 🐛 Describe the bug

According to the official Kubernetes guide (https://kubernetes.io/docs/tasks/inject-data-application/define-interdependent-environment-variables/), interdependent environment variables are defined in a roleset as follows:
```
- name: MASTER_RAY_HOST_IP
  value: "$(ROLESET_NAME)-leader"
```
 
Inside the container, MASTER_RAY_HOST_IP cannot resolve the value of ROLESET_NAME, and still remains as `$(ROLESET_NAME)-leader`.

**Reason**
When rendering the pod, stormservice sorts all environment variables in alphabetical order (in injectContainerEnvVars) to ensure deterministic output and prevent unnecessary pod updates. In the example above, the result is that MASTER_RAY_HOST_IP is defined first inside the container, while the built-in ROLESET_NAME is defined later. Consequently, MASTER_RAY_HOST_IP cannot resolve the value of ROLESET_NAME.

Environment

### Steps to Reproduce

Define interdependent environment variables in role, for example
```
    - name: SERVICE_PORT
      value: "80"
    - name: SERVICE_IP
      value: "172.17.0.1"
    - name: MY_REFERENCE
      value: "http://$(SERVICE_IP):$(SERVICE_PORT)
```

Because the alphabetical order of MY_REFERENCE comes before SERVICE_PORT and SERVICE_IP, StormService places MY_REFERENCE ahead of SERVICE_PORT and SERVICE_IP when rendering the pod. As a result, MY_REFERENCE cannot resolve the correct values of SERVICE_IP and SERVICE_PORT inside the container.

### Expected behavior

- It is desired that the order of environment variable definitions inside the container matches the order defined in the workload YAML, making it easier to define interdependent environment variables.

- It is also desired that the environment variables built into StormService appear before the environment variables defined in the workload YAML, so that they can be used in the workload.

- Finally, still ensure deterministic output and prevent unnecessary pod updates.

### Environment

AIBrix 0.6.0

Peakpine:
## Pull Request Description
Modify the way environment variables are rendered:
- Do not sort. 
- Place the built-in environment variables first.
- Followed by the environment variables defined in the container
- With a fixed order.

## Related Issues
Resolves: #2107

### DIFF
--- a/pkg/controller/roleset/utils.go
+++ b/pkg/controller/roleset/utils.go
@@ -175,6 +175,10 @@ func renderStormServicePod(roleSet *orchestrationv1alpha1.RoleSet, role *orchest
 }
 
 // injectContainerEnvVars injects env variables into container.
+// Note: Built-in env variables are added first to ensure they're available for expansion
+// in user-defined env variables. User-defined env variables maintain their original order
+// from the container spec, which should be stable across reconcile loops if the upstream
+// RoleSpec preserves order (e.g., through YAML unmarshalling).
 func injectContainerEnvVars(
 	container *v1.Container,
 	roleSet *orchestrationv1alpha1.RoleSet,
@@ -182,12 +186,8 @@ func injectContainerEnvVars(
 	roleIndex *int,
 	templateHash string,
 ) {
-	// Use map to quickly check if env variable already exists
-	existingEnvs := make(map[string]bool, len(container.Env)+6)
 	// Use slice to maintain env variable order
 	envs := make([]v1.EnvVar, 0, len(container.Env)+6)
-
-	// First add built-in env variables
 	builtInEnvs := []v1.EnvVar{
 		{
 			Name:  constants.StormServiceNameEnvKey,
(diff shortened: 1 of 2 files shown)


varungup90:
```
1. The existingEnvs[env.Name] = true line in the user-env loop is unnecessary


for _, env := range container.Env {
    if !existingEnvs[env.Name] && !ContainerInjectEnv.Has(env.Name) {
        envs = append(envs, env)
        existingEnvs[env.Name] = true  // <-- only useful if duplicates in container.Env
    }
}
This guards against duplicate names within container.Env itself. That's a valid edge case but the original code didn't protect against it either (last write wins in the map). The behavior change here is: first occurrence wins instead of last. This is acceptable but worth noting.

2. Determinism concern with user-defined env vars

The original code sorted alphabetically to ensure "deterministic output and prevent unnecessary pod updates." The new code relies on the order of container.Env being stable across reconcile loops. If the upstream source (the RoleSpec) preserves order (which it should, since YAML unmarshalling into a slice preserves order), this is fine. But if something upstream ever reorders these, you'd get spurious pod updates. This is an acceptable trade-off given the benefit, but the comment should be updated.

3. Missing test for the interdependent var scenario

The PR fixes a real bug but I don't see a test that verifies the ordering contract — specifically that a user var like value: "$(ROLESET_NAME)-leader" comes after ROLESET_NAME in the final container env. The existing tests may cover basic injection but not order-sensitive expansion.

4. Minor: sort import is now unused

The sort.Strings() call was removed, so the sort import in [pkg/controller/roleset/utils.go](vscode-webview://17uqges1ak917c2njjj9sktbpdk5uiv86vrs1jnq2otfk0vmu75l/pkg/controller/roleset/utils.go) should be removed if it's no longer used elsewhere in the file.
```

Peakpine:
@varungup90 Thank you for your review.
1. The `existingEnvs` variable and its related code logic have been removed.
2. Added comments regarding stable ordering preservation for container environment variables.
3. Added unit tests covering: built-in environment variables come first, built-in environment variables cannot be overridden, and container environment variables maintain stable ordering.
4. Other functions in the file still use `sort.Slice`, so the `sort` import is not removed.

varungup90:
can you run make lint-fix, and update the PR.

Peakpine:
@varungup90 ok，lint error has been fixed.

varungup90 (approved):

PR #2112, opened 2026-04-16
feat:ignore pods with label podGroupIndex > 0 (#2111)

### LINKED ISSUE #2111: multi-node routing support

we found it hard to route to rank_0 pod when building a multi-node inference cluster with stormservice, so we use  `ROLE_REPLICA_INDEX` in starup commands and probe detection, it works fine by far if we create a service directly to the index0 pod even when roleset is scaled

```
apiVersion: v1
kind: Service
metadata:
  name: my-model-service
  namespace: default
spec:
  type: NodePort
  selector:
    app: my-model-worker
    storm-service-name: my-model
    role: worker
    stormservice.orchestration.aibrix.ai/role-replica-index: '0'
```

```yaml
apiVersion: orchestration.aibrix.ai/v1alpha1
kind: StormService
metadata:
  name: my-model
  namespace: default
spec:
  replicas: 1
  stateful: false
  template:
    spec:
      roles:
      - name: worker
        replicas: 2
        stateful: true
        template:
          metadata:
            labels:
              app: my-model-worker
              role: worker
          spec:
            containers:
            - name: my-model-head
              command: ["/bin/bash", "-c"]
              args:
              - |
                if [ "$ROLE_REPLICA_INDEX" != "0" ]; then
                  export MASTER_RAY_HOST_IP="${ROLESET_NAME}-${ROLE_NAME}-${ROLE_TEMPLATE_HASH}-0.${STORM_SERVICE_NAME}"
                  echo "MASTER_RAY_HOST_IP: $MASTER_RAY_HOST_IP"
                fi
                opentelemetry-instrument python3 llmservice.py
              env:
              - name: RAY_NODES_NUM
                value: '2'
              - name: TENSOR_PARALLEL_SIZE
                value: '4'
              - name: DATA_PARALLEL_SIZE
                value: '4'
              - name: DATA_PARALLEL_SIZE_LOCAL
                value: '2'
              - name: DATA_PARALLEL_BACKEND
                value: 'ray'
              - name: ENABLE_EXPERT_PARALLEL
                value: 'True'
              - name: WORKER_0_DOMAIN
                value: "$(ROLESET_NAME)-$(ROLE_NAME)-$(ROLE_TEMPLATE_HASH)-0.$(STORM_SERVICE_NAME)"
              ports:
              - containerPort: 8899
                name: http
                protocol: TCP
              - containerPort: 20000
                name: distributed
                protocol: TCP
              startupProbe:
                failureThreshold: 3
                initialDelaySeconds: 500
                periodSeconds: 10
                successThreshold: 1
                timeoutSeconds: 1
                exec:
                  command:
                  - sh
                  - -c
                  - |
                  - curl -f --connect-timeout 4 --max-time 5 "http://${WORKER_0_DOMAIN}:8899/qc"
              readinessProbe:
                failureThreshold: 3
                initialDelaySeconds: 5
                periodSeconds: 10
                successThreshold: 1
                timeoutSeconds: 1
                exec:
                  command:
                  - sh
                  - -c
                  - |
                  - curl -f --connect-timeout 4 --max-time 5 "http://${WORKER_0_DOMAIN}:8899/qc"
              livenessProbe:
                failureThreshold: 3
                initialDelaySeconds: 5
                periodSeconds: 10
                successThreshold: 1
                timeoutSeconds: 10
                exec:
                  command:
                  - sh
                  - -c
                  - |
                  - curl -f --connect-timeout 4 --max-time 5 "http://${WORKER_0_DOMAIN}:8899/qc"
```

but how to access from aibrix-gateway-plugin? we should use external-filter to filter pod with label `stormservice.orchestration.aibrix.ai/role-replica-index: '0'`

then we found a func in pd strategy, 

- in pd routing-strategy, if p/d role is a multi-node leader-worker arch which only node_rank=0 runs httpServer, a n accurate target is filtered by  `PodGroupIndex`,
```
// isPodWithHTTPServer checks if a pod should be selected for routing.
// In multi-node tensor parallelism setups (e.g., TP=16 with node_rank=0 and node_rank=1),
// only pods with stormservice.orchestration.aibrix.ai/pod-group-index="0" (corresponding to node_rank=0) run the HTTP server.
// Pods without the label are also selected for backward compatibility.
func isPodWithHTTPServer(pod *v1.Pod) bool {
	podGroupIndex, exists := pod.Labels[PodGroupIndex]
	if !exists {
		// No PodGroupIndex label means single-node or old setup - include it
		return true
	}
	// Only include pods from node_rank=0 which have the HTTP server
	return podGroupIndex == "0"
}
```

should we extend this feature to all routing-strategy? 

### Use Case

if we want to use index routing, we need to set podGroupSize > 1 in stormService,  then gateway-plugins will filter pod by PodGroupIndex label first

### Proposed Solution

all strategy should filter rank0 pod with `PodGroupIndex` label if exist

rayne-Li:
Signed-off-by: rayne-Li <rayne.lyvn@gmail.com>

## Pull Request Description

addPod ignore if podGroupIndex in label and value  > 0

## Related Issues
Resolves: #2111

### DIFF
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -169,6 +169,13 @@ var _ = Describe("Cache", func() {
 		_, exist = cache.metaPods.Load("default/p1")
 		Expect(exist).To(BeFalse())
 
+		// Ignore pods with podGroupIndex > 0
+		podGroupWorker := getReadyPod("p1", "default", "m1", 0)
+		podGroupWorker.ObjectMeta.Labels[podGroupIndex] = "1"
+		cache.addPod(podGroupWorker)
+		_, exist = cache.metaPods.Load("default/p1")
+		Expect(exist).To(BeFalse())
+
 		pod := getReadyPod("p1", "default", "m1", 0)
 		cache.addPod(pod)
 
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -362,6 +362,39 @@ var _ = Describe("Cache", func() {
 		Expect(utils.CountRoutablePods(pods.All())).To(Equal(1))
 	})
 
+	It("should updatePod after podIndex updated", func() {
+		oldPod := getNewPod("p1", "default", "m1", 0)
+		oldPod.ObjectMeta.Labels[podGroupIndex] = "1"
+		cache.addPod(oldPod)
+		_, exist := cache.metaPods.Load("default/p1")
+		Expect(exist).To(BeFalse())
+
+		newPod := getReadyPod("p1", "default", "m1", 0) // IP may changed due to migration
+		newPod.ObjectMeta.Labels[podGroupIndex] = "1"
+		cache.updatePod(oldPod, newPod)
+		_, exist = cache.metaPods.Load("default/p1")
+		Expect(exist).To(BeFalse())
+
+		newPod.ObjectMeta.Labels[podGroupIndex] = "0"
+		cache.updatePod(oldPod, newPod)
+		_, exist = cache.metaPods.Load("default/p1")
+		Expect(exist).To(BeTrue())
+	})
+
+	It("should not delete pod after podIndex updated", func() {
+		oldPod := getNewPod("p1", "default", "m1", 0)
+		oldPod.ObjectMeta.Labels[podGroupIndex] = "0"
+		cache.addPod(oldPod)
+		_, exist := cache.metaPods.Load("default/p1")
+		Expect(exist).To(BeTrue())
+
+		newPod := getReadyPod("p1", "default", "m1", 0) // IP may changed due to migration
+		newPod.ObjectMeta.Labels[podGroupIndex] = "1"
+		cache.updatePod(oldPod, newPod)
+		_, exist = cache.metaPods.Load("default/p1")
+		Expect(exist).To(BeFalse())
+	})
+
 	It("should deletePod clear pod, model, and modelAdapter entrys", func() {
 		pod := getReadyPod("p1", "default", "m1", 0)
 		cache.addPod(pod)
--- a/pkg/cache/informers.go
+++ b/pkg/cache/informers.go
@@ -123,15 +123,9 @@ func (c *Store) addPod(obj interface{}) {
 		klog.V(4).InfoS("ignored pod without model label or annotation", "name", pod.Name)
 		return
 	}
-	// ignore worker pods
-	nodeType, ok := pod.Labels[nodeType]
-	if ok && nodeType == nodeWorker {
-		klog.V(4).InfoS("ignored ray worker pod", "name", pod.Name)
-		return
-	}
 
-	// ignore podGroup index > 0
-	if isWorkerPodGroupIndex(pod) {
+	// ignore worker pod
+	if isWorkerPod(pod) {
 		return
 	}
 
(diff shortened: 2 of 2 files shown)


varungup90:
## 🔍 Code Quality Review

### `isWorkerPodGroupIndex` Helper — [[pkg/cache/informers.go](https://github.com/vllm-project/aibrix/pull/2112/files)](https://github.com/vllm-project/aibrix/pull/2112/files)

* **Clean Scope:** The helper function is well-scoped and helps keep the logic modular.
* **Bug in Error Handling:** When `strconv.Atoi` fails (e.g., a malformed label value), the function currently returns `true` and logs `"podGroupIndex > 0"`. This conflates a parsing error with a valid index check, which could mask configuration issues.
    * **Suggestion:** Separate the error case with its own log message or update the log to state "invalid or > 0".
* **Logging Consistency:** `addPod` uses `klog.V(4).Infof` (printf-style), whereas the surrounding codebase primarily uses `klog.V(4).InfoS` (structured logging).
    * **Recommendation:** Match the existing style:
      ```go
      klog.V(4).InfoS("ignored pod: podGroupIndex > 0", "name", pod.Name, "index", pgIndex)
      ```

---

## ⚙️ Logic Concerns

### `updatePod` Placement & Ordering
The new filter is currently placed after the old pod deletion block:

```go
// Step 4: delete old mappings (already executed)
if oldOk || existed {
    c.deletePodLocked(...)
}

// Step 5: ray worker check
if oldNodeType == nodeWorker || newNodeType == nodeWorker { return }

// Step 6: NEW podGroupIndex check (added here)
if isWorkerPodGroupIndex(oldPod) || isWorkerPodGroupIndex(newPod) { return }
```

* **Complexity:** An index-0 pod being updated (e.g., status change) is deleted from the cache before the `podGroupIndex` check runs on the new pod. While it gets re-added if it remains index 0, this creates unnecessary churn.
* **Edge Case Risk:** If an `oldPod` has an index > 0 but the `newPod` is updated to index 0 (a transition from worker to head), the `||` logic will trigger an early return, and the new head pod will **never** be added to the cache. Using `&&` might be more appropriate for the "add new" path.

### `deletePod` Consistency
* `deletePod` has not been updated. If a pod with `podGroupIndex > 0` somehow enters the cache (via race conditions or manual injection), `deletePod` will not filter it. While low risk, updating this would ensure total consistency across the cache lifecycle.

---

## 🧪 Test Coverage Gap

The current tests only cover `addPod`. To ensure stability, I recommend adding cases for:

- [ ] **`updatePod` (Stay Ignored):** New pod has `podGroupIndex > 0` (should not be re-added).
- [ ] **`updatePod` (Transition):** Old pod has `podGroupIndex = 0` and new pod has `podGroupIndex > 0` (should delete the old entry but not add the new one).
- [ ] **Malformed Labels:** Verify behavior when `podGroupIndex` is set to a non-integer value (e.g., `"abc"`).

varungup90:
Please help address gemini and my comments, overall look good.

rayne-Li:
> Edge Case Risk: If an oldPod has an index > 0 but the newPod is updated to index 0 (a transition from worker to head), the || logic will trigger an early return, and the new head pod will never be added to the cache. Using && might be more appropriate for the "add new" path.

should we add newPod if oldPod is worker but newPod is not worker?

rayne-Li:
1. separate log for err and index > 0
2. add 2 test cases on updatePod, oldPod>0 && newPod =0(should add), oldPod=0 && newPod >0(should delete)
3. combine nodeType and podGroupIndex judgement into isWorkerPod
4. skip early if old&&new are both worker to avoid unnecessary lock
5. logic ensure: add pod to cache if newPod is not woker
6. maybe no need to add judgement in deletePod?

varungup90 (approved):

PR #2130, opened 2026-04-23
fix(roleset): cleanup orphan resources when podGroupSize changes

DebugSy:
## Summary
- Add orphan Pod cleanup in `PodSetRoleSyncer.Scale()` when `podGroupSize` switches from `<=1` to `>1`
- Add orphan PodSet cleanup in `StatefulRoleSyncer.Scale()` and `StatelessRoleSyncer.Scale()` when `podGroupSize` switches from `>1` to `<=1`
- Cleanup is idempotent (ignores NotFound errors) and executes before creating new resources

## Problem
When `podGroupSize` crosses the boundary between `<=1` and `>1`, the RoleSet controller switches between different syncers. The new syncer doesn't recognize resources created by the old syncer, leaving orphan Pods or PodSets behind that consume cluster resources indefinitely.

## Solution
Each syncer now checks for and cleans up resources that belong to the other syncer type at the start of its `Scale()` method:
- `PodSetRoleSyncer.cleanupOrphanPods()`: detects Pods with OwnerReference pointing to RoleSet (not PodSet) and deletes them
- `cleanupOrphanPodSets()`: detects PodSets belonging to the same Role and deletes them

## Test plan
- [x] Unit tests for `PodSetRoleSyncer.Scale()` with orphan Pods scenario
- [x] Unit tests for `StatefulRoleSyncer.Scale()` with orphan PodSets scenario  
- [x] Unit tests for idempotent cleanup behavior

### DIFF
--- a/pkg/controller/roleset/orphan_cleanup_test.go
+++ b/pkg/controller/roleset/orphan_cleanup_test.go
@@ -1,0 +1,370 @@
+/*
+Copyright 2025 The Aibrix Team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package roleset
+
+import (
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/utils/ptr"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/fake"
+
+	orchestrationv1alpha1 "github.com/vllm-project/aibrix/api/orchestration/v1alpha1"
+	"github.com/vllm-project/aibrix/pkg/controller/constants"
+)
+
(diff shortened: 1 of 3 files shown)


varungup90:
### 🔴 Critical: Correctness Issues

* **Asymmetric Ownership Checking**
    There is a significant discrepancy between how Pods and PodSets are handled. While `cleanupOrphanPods` correctly filters by UID, `cleanupOrphanPodSets` in `rolesyncer.go` deletes all PodSets returned by `getRolePodSets` without an ownership check:
    ```go
    allPodSets, err := getRolePodSets(ctx, cli, roleSet.Namespace, roleSet.Name, role.Name)
    // Deletes every PodSet matching the label; missing UID/Kind guard.
    ```
    **Risk:** If a `RoleSet` is deleted and recreated with the same name (new UID), any lingering PodSets from the old instance would be deleted by the new controller. 
    **Recommended Fix:** Apply an `isOwnedByRoleSet` guard within `cleanupOrphanPodSets` before deletion.

* **Partial-Deletion State Reporting**
    If a mid-list deletion fails (e.g., deleting the 3rd of 5 PodSets), the function returns `(false, err)`. While functionally safe for eventual consistency, it incorrectly reports that no cleanup occurred. 
    **Recommendation:** Consider returning `(true, err)` if at least one resource was successfully targeted, ensuring the caller is aware of the partial state change.

---

### 🧪 Test Coverage Gaps

* **StatelessRoleSyncer Coverage:** Both `StatefulRoleSyncer` and `StatelessRoleSyncer` now call the cleanup logic, but only the Stateful variant has a dedicated scale test. Please add an equivalent test for `StatelessRoleSyncer`.
* **Ownership Filter Validation:** There is currently no test case verifying that `cleanupOrphanPodSets` ignores resources owned by a different `RoleSet` (same name, different UID). This is necessary to validate the fix for the critical issue mentioned above.
* **Scale Path Idempotency:** While the helper itself has an idempotent test, the `Scale()` path for `StatefulRoleSyncer` should be tested to ensure the integration works as expected across multiple reconciles.

---

### 🟢 Minor & Cosmetic Issues

* **Log Verbosity:** `klog.Infof` will trigger on every reconcile where orphans are present. To avoid log spam in production environments, consider moving this to `klog.V(4).Infof`.
* **Style & Linting:**
    * `orphan_cleanup_test.go`: Missing newline at end of file (will likely trigger CI linters).
    * `cleanupOrphanPodSets`: Remove the trailing blank line before the closing brace for consistency with the codebase.

varungup90:
Also sign the PR

PR #2165, opened 2026-05-01
[Misc] Add brixbench benchmark module

whalepark:
## Pull Request Description
This PR imports the initial `brixbench` benchmark module into the `aibrix` repository.

Included in this import:
- benchmark runner, resolver, deployer, driver, and observability packages under `brixbench/`
- benchmark scenarios, deployment manifests, and benchmark configuration files
- walkthrough documentation for the benchmark module
- module/import path updates for the nested `brixbench` module inside `aibrix`

Additional cleanup included in this PR:
- removed local-only and AI workflow metadata from the imported tree
- updated the module path to `github.com/vllm-project/aibrix/brixbench`
- removed bundled versioned release YAMLs from the repository
- added on-demand release artifact download and local cache logic under `.tmp/releases/<version>/`
- added resolver tests for release artifact download and cache reuse

Notes:
- Version-based release artifact resolution now downloads from GitHub releases at runtime instead of relying on checked-in release YAML files.

## May 21 2026

Addressed the latest review feedback and pushed follow-up commits.

Key updates:
- Removed the hardcoded gateway fallback and now require either a detected endpoint or an explicit BENCHMARK_GATEWAY_ENDPOINT override.
- Cleaned up the resolver schema to require provider explicitly, reject deprecated deployer, and preserve provider: null as the canonical plain vLLM baseline path.
- Added focused regression tests for gateway endpoint resolution, explicit provider: null handling, deprecated deployer rejection, and missing-provider rejection.

Verification:
- go test -v ./internal/resolver -count=1
- go test -v ./benchmark -run 'Test(FallbackGatewayEndpoint(Missing|Override)|ResolveGatewayEndpoint(UsesDetectedEndpoint|UsesOverrideOnDetectionFailure|FailsWithoutDetectedEndpointOrOverride)|ConfiguredMetricExporter|ExportMetricsIfConfigured|FormatScenarioRunIDUsesUTC|NowInUTCUsesUTCLocation)$' -count=1

## May 27 2026

Pushed follow-up updates for the latest review comments.

Changes:
- Moved the benchmark walkthrough into the brixbench README entry point.
- Added documentation for benchmark workload configs, especially vllmArgs such as request-rate, concurrency, num-prompts, dataset-name, goodput, and routing-specific options.
- Removed the large generated custom AIBrix manifests from testdata and kept the benchmark flow on release-artifact/workspace Helm-based deployment paths.

Verification:
- Confirmed the removed custom manifests are not referenced by active brixbench scenarios.
- Re-ran YAML/scenario reference validation.
- Re-ran the relevant Go tests.


varungup90:
@whalepark can you sign the commit and address gemini comments?

---

### ⚠️ Review Feedback: Issues and Suggestions

Thanks for the comprehensive import of the `brixbench` module. While the scope is impressive, there are several areas—ranging from security leaks to reliability—that need to be addressed before this is merged.

---

#### 🔴 Critical / Must Fix

1.  **Internal URL Leak:**
    In `benchmark/runner_test.go`, `defaultGatewayEndpoint` is set to `http://aibrix-gateway.bytedance.net`. This internal hostname is non-functional for external contributors and should be replaced with `localhost` or a configurable default.
2.  **No-op Observability Exporter:**
    In `internal/observability/exporter.go`, `PrometheusPushExporter` currently "simulates" exports by iterating over keys without doing anything. This silently drops metrics. Please implement the export logic or return `ErrNotImplemented` to avoid misleading users.

---

#### 🟡 Significant

3.  **Shell Invocation Overload:**
    Throughout `benchmark/preflight_stormservice.go` and `vllm.go`, `bash -lc` is used to wrap `kubectl` and `helm` commands. The `-l` (login) flag introduces unnecessary environment variability.
    * **Recommendation:** Use direct binary invocations via `exec.CommandContext` or, at minimum, drop the `-l` flag.
4.  **Legacy Field Ambiguity:**
    In `internal/resolver/resolver.go`, the presence of both `Provider` and `Deployer` (legacy alias) is confusing for an initial import. Please document the deprecation timeline or clean this up now to avoid technical debt.
5.  **Initialization Panics:**
    `mustLoadPacificLocation()` in `benchmark/runner_support_test.go` will panic if `tzdata` is missing (common in minimal CI containers).
    * **Fix:** Use a lazy initializer or provide a fallback like `time.FixedZone("PT", -8*3600)`.
6.  **Hardcoded Timezone in Logs:**
    `formatScenarioRunID` embeds "PT" in filenames. This is problematic for global teams. Please switch to **UTC** for unambiguous, reproducible log naming.

---

#### 🔵 Minor / Improvements

7.  **URL Construction:**
    In `internal/resolver/artifact_resolution.go`, replace manual string joining with `url.JoinPath` (available since Go 1.19) for better slash normalization.
8.  **Modern Atomic Usage:**
    Since the project uses Go 1.22.5, prefer `atomic.Int32` over `atomic.AddInt32` in `artifact_resolution_test.go` for cleaner syntax.
9.  **Dependency Transparency:**
    `brixbench/go.mod` is quite sparse because it relies heavily on shelling out to `kubectl`. Consider explicitly documenting the requirement for a pre-configured K8s cluster in `docs/walkthrough.md`.
10. **Module Wiring:**
    Ensure `brixbench/` is added to the root `go.work` or CI pipeline so it doesn't bitrot immediately after merging.

brixbench/benchmark/runner_test.go
```diff
@@ -0,0 +1,245 @@
...
+	if err := deployer.DeployControlPlane(ctx); err != nil {
+		return deployer, "", fmt.Errorf("failed to deploy control plane: %w", err)
+	}
+	if err := deployer.DeployEngine(ctx); err != nil {
+		return deployer, "", fmt.Errorf("failed to deploy engine: %w", err)
+	}
+	if err := deployer.WaitForReady(ctx); err != nil {
+		return deployer, "", fmt.Errorf("engine not ready: %w", err)
+	}
+	time.Sleep(20 * time.Second)
```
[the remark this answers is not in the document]
whalepark: Addressed. Removed the fixed post-ready sleep and now rely on WaitForReady() as the readiness gate. The real hello-world benchmark path was re-run successfully after this change.

brixbench/internal/deployers/command_logs.go
```diff
@@ -0,0 +1,109 @@
...
+	"time"
+)
+
+var commandLogNameSanitizer = regexp.MustCompile(`[^a-z0-9._-]+`)
+
+func (d *AIBrixDeployer) runLoggedCommand(ctx context.Context, stage string, command string) (string, error) {
+	startedAt := time.Now()
+	fmt.Printf("[cmd] START %s\n", command)
+
+	cmd := exec.CommandContext(ctx, "bash", "-c", command)
```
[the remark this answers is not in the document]
whalepark: Addressed. Refactored the command logger so simple commands use direct exec arguments via exec.CommandContext(ctx, name, args...). Shell-required flows are kept on a separate explicit shell helper.

brixbench/.gitignore
```diff
@@ -0,0 +1,56 @@
...
+# Personal work logs
+.worklog/
+
+# Spec-Kit auto-generated memory and rule files
+.specify/memory/
+.trae/rules/speckit.*.md
+
+# Keep only essential Spec-Kit files
+!.trae/rules/project_rules.md
+benchmark/__pycache__/plot_summary_vllm_bench.cpython-311.pyc
```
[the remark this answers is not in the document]
whalepark: Addressed. Replaced the version-specific generated .pyc ignore entry with a generic **/__pycache__/ rule.

brixbench/benchmark/preflight_stormservice.go
```diff
@@ -0,0 +1,104 @@
...
+		return nil
+	}
+	return fmt.Errorf(
+		"existing StormService resources found outside the reset benchmark namespace:\n%s",
+		formatStormServices(services),
+	)
+}
+
+func listStormServices(ctx context.Context) ([]stormServiceRef, error) {
+	cmd := exec.CommandContext(ctx, "bash", "-lc", "kubectl get stormservice -A -o json")
```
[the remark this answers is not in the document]
whalepark: Addressed. Replaced the StormService preflight shell invocation with a direct kubectl call:
exec.CommandContext(ctx, "kubectl", "get", "stormservice", "-A", "-o", "json").

brixbench/internal/drivers/vllm.go
```diff
@@ -0,0 +1,601 @@
...
+	if podPhase != "Succeeded" {
+		return fmt.Errorf("benchmark pod finished with phase %s", podPhase)
+	}
+
+	resultsFilename := config.Artifacts.ResultFilename
+	d.resultsPath = filepath.Join(runDir, resultsFilename)
+	fmt.Printf("[vllm-bench] Persisting benchmark results to: %s\n", d.resultsPath)
+	if err := persistBenchResultsFromLog(logPath, d.resultsPath); err != nil {
+		return fmt.Errorf("failed to persist benchmark results: %v (check %s for details)", err, logPath)
+	}
```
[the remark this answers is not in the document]
whalepark: I tested the suggested kubectl cp path, but in this cluster the benchmark pod is already in Succeeded phase when artifact collection runs, and kubectl cp fails because it requires exec into the completed container.

I kept the existing log-marker extraction path for this PR and documented this as an intentional WONTFIX/follow-up. A file-copy based transport would need a different artifact handoff design.

whalepark:
> @whalepark can you sign the commit and address gemini comments?
> 
> ### ⚠️ Review Feedback: Issues and Suggestions
> Thanks for the comprehensive import of the `brixbench` module. While the scope is impressive, there are several areas—ranging from security leaks to reliability—that need to be addressed before this is merged.
> 
> #### 🔴 Critical / Must Fix
> 1. **Internal URL Leak:**
>    In `benchmark/runner_test.go`, `defaultGatewayEndpoint` is set to `http://aibrix-gateway.bytedance.net`. This internal hostname is non-functional for external contributors and should be replaced with `localhost` or a configurable default.
> 2. **No-op Observability Exporter:**
>    In `internal/observability/exporter.go`, `PrometheusPushExporter` currently "simulates" exports by iterating over keys without doing anything. This silently drops metrics. Please implement the export logic or return `ErrNotImplemented` to avoid misleading users.
> 
> #### 🟡 Significant
> 3. **Shell Invocation Overload:**
>    Throughout `benchmark/preflight_stormservice.go` and `vllm.go`, `bash -lc` is used to wrap `kubectl` and `helm` commands. The `-l` (login) flag introduces unnecessary environment variability.
>    
>    * **Recommendation:** Use direct binary invocations via `exec.CommandContext` or, at minimum, drop the `-l` flag.
> 4. **Legacy Field Ambiguity:**
>    In `internal/resolver/resolver.go`, the presence of both `Provider` and `Deployer` (legacy alias) is confusing for an initial import. Please document the deprecation timeline or clean this up now to avoid technical debt.
> 5. **Initialization Panics:**
>    `mustLoadPacificLocation()` in `benchmark/runner_support_test.go` will panic if `tzdata` is missing (common in minimal CI containers).
>    
>    * **Fix:** Use a lazy initializer or provide a fallback like `time.FixedZone("PT", -8*3600)`.
> 6. **Hardcoded Timezone in Logs:**
>    `formatScenarioRunID` embeds "PT" in filenames. This is problematic for global teams. Please switch to **UTC** for unambiguous, reproducible log naming.
> 
> #### 🔵 Minor / Improvements
> 7. **URL Construction:**
>    In `internal/resolver/artifact_resolution.go`, replace manual string joining with `url.JoinPath` (available since Go 1.19) for better slash normalization.
> 8. **Modern Atomic Usage:**
>    Since the project uses Go 1.22.5, prefer `atomic.Int32` over `atomic.AddInt32` in `artifact_resolution_test.go` for cleaner syntax.
> 9. **Dependency Transparency:**
>    `brixbench/go.mod` is quite sparse because it relies heavily on shelling out to `kubectl`. Consider explicitly documenting the requirement for a pre-configured K8s cluster in `docs/walkthrough.md`.
> 10. **Module Wiring:**
>     Ensure `brixbench/` is added to the root `go.work` or CI pipeline so it doesn't bitrot immediately after merging.

Thanks for the detailed review. I pushed follow-up commits addressing the feedback.

Summary:
- Removed the hardcoded internal gateway fallback. The runner now uses the detected endpoint, allows an explicit BENCHMARK_GATEWAY_ENDPOINT override, and fails fast otherwise.
- Made the Prometheus exporter return ErrNotImplemented and disabled metrics export by default unless BENCHMARK_PUSHGATEWAY_URL is set.
- Reduced shell wrapping in the driver/deployer paths by moving simple kubectl/helm calls to direct exec helpers, while keeping shell-required flows explicit.
- Removed the legacy deployer schema field from the resolver, now require provider explicitly, and kept provider: null as the canonical plain vLLM baseline.
- Replaced Pacific-time initialization/run IDs with UTC.
- Switched release artifact URL construction to url.JoinPath.
- Modernized resolver test atomics to atomic.Int32.
- Added Kubernetes/kubectl prerequisite documentation to the walkthrough.
- Checked root workspace/CI wiring and confirmed brixbench is not currently wired there; I left that as a separate follow-up instead of widening this PR.

Verification:
- go test -v ./internal/resolver -count=1
- go test -v ./internal/drivers -count=1
- go test -v ./benchmark -run 'Test(FallbackGatewayEndpoint(Missing|Override)|ResolveGatewayEndpoint(UsesDetectedEndpoint|UsesOverrideOnDetectionFailure|FailsWithoutDetectedEndpointOrOverride)|ConfiguredMetricExporter|ExportMetricsIfConfigured|FormatScenarioRunIDUsesUTC|NowInUTCUsesUTCLocation)$' -count=1
- YAML validation for the new Qwen3-8B routing scenario: parsed 30 YAML files, verified 24 scenario test cases, and confirmed benchmark/engine/gateway resource references.
- Re-ran the real cluster hello-world path as documented in the tracker.

whalepark:
Pushed the latest updates to feature/add-brixbench.

New commits:
- 6145393 brixbench: address review feedback
- aa3c94c brixbench: add qwen3 routing scenario

What changed:
- Addressed the review feedback around gateway endpoint fallback, observability export behavior, shell invocation cleanup, provider/deployer schema ambiguity, UTC run IDs, artifact URL construction, atomic usage, and Kubernetes prerequisites documentation.
- Added a Qwen3-8B routing scenario with 24 test cases, including default routing and PD routing variants across 4p8d, 8p4d, and 8p8d configurations.
- Added benchmark YAMLs, model manifests, and gateway routing overrides for the new scenario.
- Updated the vLLM benchmark driver so an explicit dataset-name in scenario YAML is preserved instead of being overwritten by the default random dataset.

Verification:
- go test -v ./internal/resolver -count=1
- go test -v ./internal/drivers -count=1
- go test -v ./benchmark -run 'Test(FallbackGatewayEndpoint(Missing|Override)|ResolveGatewayEndpoint(UsesDetectedEndpoint|UsesOverrideOnDetectionFailure|FailsWithoutDetectedEndpointOrOverride)|ConfiguredMetricExporter|ExportMetricsIfConfigured|FormatScenarioRunIDUsesUTC|NowInUTCUsesUTCLocation)$' -count=1
- Parsed 30 new YAML files and verified all 24 scenario test cases reference existing benchmark, engine manifest, and gateway resource files.

brixbench/docs/walkthrough.md
```diff
@@ -0,0 +1,452 @@
+# AIBrix Benchmark Suite - Walkthrough Guide
```
Jeffwan: change to README.md? 

Feel free to add docs under project_root/docs as well

brixbench/docs/walkthrough.md
```diff
@@ -0,0 +1,452 @@
...
+Scenario: aibrix-recent-commits
+Tests:
+  - name: aibrix-main-d41653b-disaggregated
+    provider: aibrix
+    fullstack: false
+    commit: d41653b26cf1257157e34ff35ee5f028644473a8
+    engine:
+      type: vllm
+      manifest: testdata/deployments/aibrix/models/pd-model.yaml
+    benchmark: testdata/benchmarks/vllm-chat-smoke-pd.yaml
```
Jeffwan: I checked this file, seems it's different from what I expected. looks like it still describe how to deploy a model. I am looking for the benchmark data, how we send request to gateway (rps, concurrency, arrival rate etc). do you have such configs?

```
kind: vllm-bench
execution: cluster
image: aibrix-container-registry-cn-beijing.cr.volces.com/autodash/vllm-bench:v0.10.2-20260118
namespace: brixbench-adhoc
podName: vllm-bench-client
modelHostPath: /data01/models
rootHostPath: /root

artifacts:
  resultFilename: bench_results.json
  logDir: testdata/logs

vllmArgs:

```

brixbench/benchmark/testdata/deployments/aibrix/custom/aibrix-core-custom.yaml
Jeffwan, line 1: brixbench/benchmark/testdata/deployments/aibrix/custom/aibrix-dependency-custom.yaml 
and
brixbench/benchmark/testdata/deployments/aibrix/custom/aibrix-core-custom.yaml

are too heavy, do not check in and let's get this removed. Instead, I highly suggest to use production recommended way to deploy it. please check whether helm has some sdk so we can use helm. If not, we can use kustomize to generate the right configuration.

brixbench/docs/walkthrough.md
whalepark: Sounds valid. Will do.

brixbench/docs/walkthrough.md
whalepark: You are right, the walkthrough was focused on deployment/scenario wiring and did not clearly explain the request workload config. The request-side benchmark settings live in the benchmark YAML referenced by each test case, under vllmArgs. (`brixbench/benchmark/testdata/benchmarks/vllm-chat-smoke-pd.yaml`) I will update the README to add a dedicated benchmark workload section covering request rate, concurrency, num prompts, endpoint, dataset name, and routing specific options.

brixbench/benchmark/testdata/deployments/aibrix/custom/aibrix-core-custom.yaml
whalepark: Let me remove them. Thanks for pointing out.

varungup90:
Thanks for the hard work on this PR! I’ve reviewed the changes and found a few items that need to be addressed before we merge. I've broken them down into correctness issues, architecture concerns, and general code quality improvements.

### 🔴 Correctness Issues

**1. Incomplete `llmd`/`dynamo` deployer branches (`runner_test.go:1545-1548`)**

```go
case "llmd":
    t.Log("Using LLM-d deployer")
    // return nil, "", fmt.Errorf("LLM-d deployer not implemented")
case "dynamo":
    t.Log("Using Dynamo deployer")
    // return nil, "", fmt.Errorf("Dynamo deployer not implemented")

```

The commented-out error returns fall through to the `deployer.Initialize()` call with `deployer == nil`, which will panic at runtime. Either uncomment the error returns or initialize with placeholder deployers (`NewLLMdDeployer()` / `NewDynamoDeployer()` already exist).

**2. `Initialize` sets fields twice when `TestCase != nil` (`aibrix.go:4686-4708`)**
The method first assigns from `config.*` then immediately overwrites with `config.TestCase.*`. The first assignment block (lines 4687-4697) is dead code when `TestCase != nil`.

**3. `PlainVLLMDeployer.Teardown` silently swallows errors (`plain_vllm.go`)**

```go
func (d *PlainVLLMDeployer) Teardown(ctx context.Context) error {
    d.deleteNamespace(ctx, d.namespace)
    return nil
}

```

The return value of `deleteNamespace` is discarded and the method always returns `nil`, breaking teardown error reporting.

**4. `capturePodLogs` builds a shell command string using `fmt.Sprintf` then passes it to `bash -c` (`runner_logs_test.go:1121-1122`)**

```go
command := fmt.Sprintf("kubectl logs -n %q %q --all-containers=true --timestamps=true", namespace, podName)
cmd := exec.CommandContext(ctx, "bash", "-c", command)

```

`%q` adds Go-style quoting (with backslash escapes) that bash may not handle correctly for pod names with unusual characters. Use `exec.CommandContext` with separate args instead of `bash -c`.

**5. `generateVLLMBenchFigures` uses `runtime.Caller(0)` to locate the Python script (`runner_artifacts_test.go:847`)**
This resolves the script path relative to the compiled test binary's source path, which only works when running `go test` from source. It fails for any compiled binary or when the test is run from a different working directory.

---

### 🟡 Design / Architecture Concerns

**6. Business logic inside `_test.go` files**
Core benchmark logic (runner, artifact writing, figure generation, deployer selection) lives in `*_test.go` files in the `benchmark` package. This means the code is only compiled when running tests and cannot be imported or reused. For a tool that may grow, consider moving the runner into a proper package (or `main`).

**7. `TestAIBrixBenchmarkSuite` is both the test harness and a real integration test**
The main test function runs a full Kubernetes deployment. Running `go test ./benchmark/...` without guards (e.g., a `//go:build integration` tag or a `testing.Short()` skip) will attempt to contact a live cluster. This will fail CI and confuse developers running unit tests. The unit tests (resolver, fallback endpoint, time utilities) are clean but co-mingled with integration tests.

**8. Metrics export is a stub (`ErrNotImplemented`)**
The `PrometheusPushExporter.Export` returns `ErrNotImplemented` always, and `executeScenarioTestCase` treats this as a fatal error. This means configuring `BENCHMARK_PUSHGATEWAY_URL` will always fail the benchmark case even though metric export is explicitly documented as a future feature.

**9. Hardcoded Volcengine-internal image registry in testdata YAMLs**
All benchmark/deployment YAMLs reference `aibrix-container-registry-cn-beijing.cr.volces.com` and `aibrix-public-release-cn-beijing.cr.volces.com`. These are internal/private registries not accessible to community contributors. The README should note that contributors must substitute their own image registries, or the YAMLs should use publicly accessible images.

**10. `aibrix-hello-world.yaml` scenario references `testdata/deployments/aibrix/models/model.yaml` which doesn't exist in this PR**
Several scenario files reference `testdata/deployments/aibrix/models/model.yaml` and `testdata/deployments/aibrix/models/pd-model.yaml`, but only `qwen3-8b-pd-*.yaml` files are present. The default scenario will fail.

---

### 🔵 Code Quality

**11. `.trae/rules/speckit.*.md` in `.gitignore**`
The `.gitignore` mentions `.trae/rules/speckit.*.md` and `.specify/memory/` — these look like AI assistant tooling artifacts and suggest some AI-generated code was imported without filtering workspace-local config. The negative pattern `!.trae/rules/project_rules.md` also implies `.trae/` files exist in the repo tree.

**12. `Dockerfile.builder` installs packages twice**
In `Dockerfile.builder`: `libkrb5-dev` and `libzmq3-dev` are installed with a plain `apt-get install` and then again inside a `--mount=type=cache` layer. The first non-cached install is dead weight.

**13. `exportMetricsIfConfigured` prints to `fmt.Printf` rather than using `t.Log**`

```go
fmt.Printf("[benchmark] Metrics export disabled; no exporter configured\n")

```

This is called from within a test context where `t.Log` / `t.Logf` should be used so output is captured and associated with the correct test case.

Jeffwan (approved):

PR #2184, opened 2026-05-06
feat(batch): basic planner passthrough for integration test

nwangfw:
## Pull Request Description

This is the very basic planner implementation and used for integration test between console, planner, RM, and MDS. Has tested in the desktop k8s env. Will replace it with a queue-based planner later.

In order to make it work. The PR https://github.com/vllm-project/aibrix/pull/2183 needs to be merged first


Jeffwan (approved):
few feedback, let's address them in future PRs

PR #2196, opened 2026-05-12
optimize: gateway-plugin cpu usage optimize when stream is true #2196

rayne-Li:
Signed-off-by: rayne-Li <lizhy1@chinatelecom.cn>

## Pull Request Description
optimize cpu usage by reduce memory alloc and gc, use gjson instead avoid fully unmarshall resp when stream=True

when 1000QPS with total 5000 request

- before: cpu: 32C, after: 17c

- before cpu-pprof: 
image

- before mem-pprof
image

- after mem-pprof
image

- diff
image

### DIFF
--- a/apps/console/api/planner/api/jobs.go
+++ b/apps/console/api/planner/api/jobs.go
@@ -93,4 +93,3 @@ type ModelTemplateRef struct {
 	Name    string `json:"name"`
 	Version string `json:"version,omitempty"`
 }
-
--- a/pkg/plugins/gateway/gateway_rsp_body.go
+++ b/pkg/plugins/gateway/gateway_rsp_body.go
@@ -20,16 +20,13 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"io"
 	"math"
-	"net/http"
 	"strconv"
 	"strings"
 	"time"
 
 	"github.com/bytedance/sonic"
-	"github.com/openai/openai-go/v3"
-	"github.com/openai/openai-go/v3/packages/ssestream"
+	"github.com/tidwall/gjson"
 	v1 "k8s.io/api/core/v1"
 	"k8s.io/klog/v2"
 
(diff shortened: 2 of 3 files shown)


rayne-Li:
@varungup90 please take a look @

varungup90:
Nice Work!!

## ⚠️ Correctness Issues

### 1. Error handling is entirely removed (Significant Regression)

The old code detected SSE parse/decode errors and returned an `InternalServerError` with `HeaderErrorStreaming: true`:

```go
if err := streaming.Err(); err != nil {
    return generateErrorResponse(envoyTypePb.StatusCode_InternalServerError, ...)
}

```

The new implementation drops this completely. A corrupted response, malformed JSON, or upstream error in the stream will now silently pass through. The only remaining signal is a `klog.Warningf` when tokens are zero, which is not actionable for the caller.

### 2. Trace completion flag not set on error path

**Location:** [`[pkg/plugins/gateway/gateway_rsp_body.go:110](https://www.google.com/search?q=%23)`](https://www.google.com/search?q=%23)

The previous implementation ensured `complete = true` was set before returning the error response. Since the new code lacks an error path for streaming, `complete` may not be accurately set in degraded scenarios. The deferred `DoneRequestTrace` call depends on this flag being accurate.

### 3. SSE line ending assumption

The SSE spec permits `\r\n` line endings. While the previous library handled this automatically, the new code only splits on `\n`.

> [!NOTE]
> `bytes.TrimSpace` on each line will strip a trailing `\r`, but this is implicit and fragile. This should at least be documented with a comment, or handled explicitly.

### 4. Pre-filter false positives

```go
if bytes.Contains(bodyBytes, []byte(`"usage"`)) {

```

While this guards against unnecessary work, `"usage"` could appear in a model's generated content (e.g., a response explaining how to use an API). In these cases, the loop will run, find no valid tokens, and potentially emit a spurious warning.

---

## 🛠 Code Quality & Suggestions

* **Missing Tests:** Manual SSE parsing is subtle and error-prone. We should add unit tests covering:
* Normal usage chunks.
* `[DONE]` terminators.
* Partial or missing usage fields.
* Malformed JSON.
* `\r\n` line endings.


* **Non-idiomatic Comments:** The comment `// use gjson.GetBytes accelerate` is a bit unclear.
* **Suggestion:** `// gjson avoids full deserialization by only extracting the usage field.`


* **Assumptions on Usage Accumulation:** The change from overwriting token counts to accumulating them assumes the upstream only sends usage in the final chunk (standard vLLM behavior). This assumption should be documented in the code.

rayne-Li:
@varungup90 thanks for your reply

## for 1 and 2

- we add json malformed validation like `gjson.ValidBytes(jsonBytes)`
```
// Look for the SSE data prefix
if bytes.HasPrefix(line, []byte("data:")) {
	// Slice the "data:" prefix (zero allocation)
	jsonBytes := bytes.TrimSpace(line[5:])

	// Check for the end of the stream
	if bytes.Equal(jsonBytes, []byte("[DONE]")) {
		continue
	}
	if !gjson.ValidBytes(jsonBytes) {
		complete = true
		return generateErrorResponse(
			envoyTypePb.StatusCode_InternalServerError,
				Key: HeaderErrorStreaming, RawValue: []byte("true"),
			}}},
			"malformed JSON in SSE stream", "", ""), complete
	}

	// gjson avoids full deserialization by only extracting the usage field.
	usageResult := gjson.GetBytes(jsonBytes, "usage")
    // ......
}
```

## for 3 and 4

i add some comments above and test cases

varungup90 (approved):

PR #2197, opened 2026-05-13
feat(batch): implement async planner for batch orchestration

nwangfw:
## Pull Request Description
Replaces the synchronous `Passthrough` planner with an asynchronous queue-backed implementation to support the case when the provisioning takes time. The gRPC caller now returns immediately with a synthetic "validating" batch while a worker pool runs `Provision` + `CreateBatch` in the background. The `plannerapi.Planner` interface is unchanged; the swap is one line in `server.go`.


apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,428 @@
...
+		}{
+			ProvisionID: provResult.ProvisionID,
+		},
+		ModelTemplate: req.ModelTemplate,
+	}
+
+	klog.Infof("[planner.scheduler] submit job_id=%q provision_id=%q model_template=%v",
+		req.JobID, provResult.ProvisionID, req.ModelTemplate)
+
+	batch, err := q.bc.CreateBatch(q.baseCtx, req.BatchParams, aibrix)
```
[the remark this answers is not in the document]
nwangfw: The race is handled at the post-CreateBatch checkpoint. If the job was canncelled. The provisioned resource will be released and the submission will be aborted.

apps/console/api/server/server.go
```diff
@@ -116,7 +119,8 @@ func (s *Server) StartGRPC(addr string) error {
 	if err != nil {
 		return fmt.Errorf("resource manager init: %w", err)
 	}
-	planner := plannerimpl.NewPassthrough(batchClient, rm.Provisioner)
+	planner := plannerimpl.NewScheduler(batchClient, rm.Provisioner, plannerimpl.DefaultWorkerCount)
+	s.planner = planner
```
Jeffwan: 122 and 123 could be single line?

apps/console/api/planner/impl/planner_test.go
```diff
@@ -0,0 +1,1047 @@
...
+// =============================================================================
+// Fakes
+// =============================================================================
+
+// fakeProvisioner implements provisioner.Provisioner with caller-supplied
+// behavior per call site. Tests set ProvisionFn / ReleaseFn to inject
+// success/error/latency; the struct also records every call for later
+// assertion and tracks peak concurrent in-flight Provisions so worker-pool
+// parallelism can be measured directly.
+type fakeProvisioner struct {
```
Jeffwan, line 44: this should be common provisioner interface. Can they be reused or not? 

why not put them under `async_planner_test.go`..

apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,548 @@
...
+	mu         sync.RWMutex          // guards jobs and jobByBatch
+	jobs       map[string]*queuedJob // JobID -> state
+	jobByBatch map[string]string     // batch.ID -> JobID (for ListJobs tagging)
+
+	// provPollInterval is how often waitForProvisionReady polls the RM.
+	// Same-package tests override for fast assertions.
+	provPollInterval time.Duration
+}
+
+// jobState is the planner-side lifecycle. Before submission, statusFor maps
```
Jeffwan, line 59: it should be not part of this file? is there planner types could be used?

apps/console/api/server/server.go
```diff
@@ -207,6 +211,11 @@ func (s *Server) Shutdown(ctx context.Context) {
 			klog.Errorf("failed to shutdown http server: %v", err)
 		}
 	}
+	if c, ok := s.planner.(io.Closer); ok {
+		if err := c.Close(); err != nil {
```
Jeffwan: what  happened for those workers on provisioning?

apps/console/api/planner/impl/async_planner.go
```diff
@@ -0,0 +1,548 @@
...
+		}
+		return &plannerapi.Job{JobID: jobID, Batch: batch}, nil
+	}
+	// Already terminal (failed/canceled) — return current view, no double-cancel side effects.
+	return &plannerapi.Job{JobID: jobID, Batch: placeholderBatch(req, statusFor(state), enqueuedAt, terminalAt)}, nil
+}
+
+// ListJobs merges MDS batches with local not-yet-submitted jobs. Local jobs
+// are shown only on the first page so the MDS cursor remains valid.
+func (q *AsyncPlanner) ListJobs(ctx context.Context, req *plannerapi.ListJobsRequest) (*plannerapi.ListJobsResponse, error) {
```
Jeffwan: is paging still working? seem not?

apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,548 @@
...
+const DefaultWorkerCount = 8
+
+// defaultProvPollInterval matches the cadence used by Provisioner-level
+// integration tests when waiting for "running" status.
+const defaultProvPollInterval = 5 * time.Second
+
+// provReadyTimeout caps how long a single worker will wait for a
+// provision to reach Running. Beyond this, the job is marked Failed and
+// the resource is released.
+const provReadyTimeout = 2 * time.Minute
```
Jeffwan, line 115: I do not quite understand why it's 2 mins? any analytics?

apps/console/api/planner/impl/async_planner.go
```diff
@@ -0,0 +1,548 @@
...
+		JobID: jobID,
+		Batch: placeholderBatch(req, statusFor(state), enqueuedAt, terminalAt),
+	}, nil
+}
+
+// Cancel marks a pending/provisioning job canceled, or forwards cancel to
+// MDS for a submitted job. A cancel that lands mid-Provision or
+// mid-CreateBatch is honored at the worker's post-CreateBatch checkpoint
+// (which forwards CancelBatch and releases the resource).
+func (q *AsyncPlanner) Cancel(ctx context.Context, jobID string) (*plannerapi.Job, error) {
```
Jeffwan: does it work with checkpoint pretty well?

apps/console/api/planner/impl/async_planner.go
```diff
@@ -0,0 +1,548 @@
...
+		q.releaseAfter(jobID, provResult.ProvisionID, "wait failure")
+		q.markFailed(jobID, errors.Join(plannerapi.ErrInsufficientResources, err))
+		return
+	}
+	klog.Infof("[planner] provision ready job_id=%q provision_id=%q provider=%q",
+		jobID, provResult.ProvisionID, q.prov.Type())
+
+	aibrix := plannerclient.AIBrixExtraBody{
+		JobID: req.JobID,
+		PlannerDecision: &struct {
```
Jeffwan: why do you replicate a struct here?

apps/console/api/planner/impl/async_planner.go
```diff
@@ -0,0 +1,548 @@
...
+	for _, job := range unsubmitted {
+		out = append(out, &plannerapi.Job{
+			JobID: job.req.JobID,
+			Batch: placeholderBatch(job.req, statusFor(job.state), job.enqueuedAt, job.terminalAt),
+		})
+	}
+	return out
+}
+
+func (q *AsyncPlanner) deleteJob(jobID string) {
```
Jeffwan: how do you manage jobs in the queue？ Seem delete is not invoked if job is full or expired?

apps/console/api/planner/impl/async_planner.go
```diff
@@ -0,0 +1,428 @@
...
+				jobID, provResult.ProvisionID, relErr)
+		}
+		q.markFailed(jobID, err)
+		return
+	}
+
+	// Record batch.ID and mark the job submitted.
+	q.mu.Lock()
+	if job, ok := q.jobs[jobID]; ok {
+		job.state = jobStateSubmitted
```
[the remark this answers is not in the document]
nwangfw: Same as the previous comment.

apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,428 @@
...
+
+	submit chan string // buffered FIFO of pending JobIDs
+
+	baseCtx    context.Context
+	baseCancel context.CancelFunc
+	wg         sync.WaitGroup // tracks live workers; Close waits on this
+
+	mu         sync.RWMutex          // guards jobs and jobByBatch
+	jobs       map[string]*queuedJob // JobID -> state
+	jobByBatch map[string]string     // batch.ID -> JobID (for ListJobs tagging)
```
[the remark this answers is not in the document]
nwangfw: The concern is true. Thinking to address this in a seperate PR in the near future.

apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,428 @@
...
+	jobStateCanceled
+)
+
+type queuedJob struct {
+	req        *plannerapi.EnqueueRequest
+	state      jobState
+	batchID    string // populated when state == jobStateSubmitted
+	err        error  // populated when state == jobStateFailed
+	enqueuedAt time.Time
+}
```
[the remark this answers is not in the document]
nwangfw: Addressed

apps/console/api/planner/impl/async_planner.go
```diff
@@ -0,0 +1,428 @@
...
+	if len(req.BatchParams.Metadata) > 0 {
+		b.Metadata = map[string]string(req.BatchParams.Metadata)
+	}
+	switch st {
+	case openai.BatchStatusFailed:
+		b.FailedAt = time.Now().Unix()
+	case openai.BatchStatusCancelled:
+		b.CancelledAt = time.Now().Unix()
+	}
+	return b
```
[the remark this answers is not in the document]
nwangfw: Addressed by adding failedAt time.Time and canceledAt time.Time fields

apps/console/api/server/server.go
nwangfw: There are two cases. If the the planner has received the provisionID, it will release the resource accordingly. If the RM has allocated internally but hasn't returned the ProvisionID, there will be a caveat, bounded by the RPC duration. In practice, this window should be very small?

apps/console/api/planner/impl/async_planner.go
nwangfw: Actually this is something I want to discuss and get feedback on regarding how to merge job status for jobs in queue and jobs in MDS.

Paging works but not perfectly. Right now, the implementation idea is to display all local jobs (jobs not submited to MDS) first. The jobs in MDS will be displayed after that. The reason we can't paginate strictly across both: the MDS After cursor only encodes positions in MDS's batch ID space; local jobs have no stable IDs (they're in-memory, lost on restart), so we can't build a unified cursor across both sources.   

The fix of this issue can be deferred to a future PR.

apps/console/api/planner/impl/planner.go
nwangfw: It depends on the provisioning speed of resource providers.  The 2*min is more than enough for the current k8s provisioning process but we may need to adjust this value for other resource providers.

apps/console/api/planner/impl/async_planner.go
nwangfw: Yes, we re-read state at the end of process() after CreateBatch returns to check if a cancel landed; covered by TestCancelDuringProvisioning* and TestCancelDuringCreateBatch*. Let me know if you see any potential issues.

apps/console/api/planner/impl/async_planner.go
nwangfw: Jobs are dequeued and removed from the Go channel implicitly by the worker's <-q.submit receive — no explicit delete is needed for the queue itself. The previous deleteJob has been renamed to rollbackEnqueue for clarity and is only used to undo the bookkeeping insert in two edge cases where Enqueue itself fails.  Added some comments in the code as well.

apps/console/api/server/server.go
nwangfw: Modified.

apps/console/api/planner/impl/async_planner.go
nwangfw: Created a named type for reuse.

apps/console/api/planner/impl/planner.go
nwangfw: No existing planner type fits.  jobState describes the AsyncPlanner-internal pre-submission lifecycle (Pending → Provisioning, plus terminal Failed/Canceled before MDS takes over) and isn't referenced outside this file,

apps/console/api/planner/impl/planner_test.go
nwangfw: Right now, the test file is only used by planner. Maybe we can keep it here until later it is reused by another compoment.

apps/console/api/planner/impl/async_planner.go
```diff
@@ -0,0 +1,548 @@
...
+			ProvisionResourceDeadline int64  `json:"provision_resource_deadline,omitempty"`
+			ResourceDetails           []struct {
+				ResourceType    string `json:"resource_type"`
+				EndpointCluster string `json:"endpoint_cluster,omitempty"`
+				GPUType         string `json:"gpu_type,omitempty"`
+				WorkerNum       int    `json:"worker_num,omitempty"`
+			} `json:"resource_details,omitempty"`
+		}{
+			ProvisionID: provResult.ProvisionID,
+		},
```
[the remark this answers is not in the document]
nwangfw: Created PlannerDecision name type

apps/console/api/planner/client/submission.go
```diff
@@ -20,21 +20,27 @@ import (
 	plannerapi "github.com/vllm-project/aibrix/apps/console/api/planner/api"
 )
 
+// PlannerDecision is the planner's allocation decision projected into
+// AIBrixExtraBody. ResourceDetails stays as a nested anonymous struct
+// because its shape will evolve as the RM populates per-group allocation
+// details (currently always nil — RM doesn't return them yet).
+type PlannerDecision struct {
+	ProvisionID               string `json:"provision_id,omitempty"`
+	ProvisionResourceDeadline int64  `json:"provision_resource_deadline,omitempty"`
```
DwyaneShi: need put a comment here to explain if the deadline is a unix timestamp in seconds, milliseconds, etc.

apps/console/api/planner/client/submission.go
```diff
@@ -20,21 +20,27 @@ import (
...
+	ResourceDetails           []struct {
```
DwyaneShi, line 33: 1. let's have enough comments to explain these fields
2. need a comment to explain that the actual gpu num of each deployment is defined in the template, otherwise other contributors will get confused

apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,546 @@
...
+
+// Planner is an asynchronous implementation of plannerapi.Planner.
+// Enqueue records the job in memory, returns a placeholder batch in
+// "pending" status, and lets workers run Provision, wait for the
+// resource to reach Running, then CreateBatch.
+type Planner struct {
+	bc   plannerclient.BatchClient
+	prov provisioner.Provisioner
+
+	submit chan string // buffered FIFO of pending JobIDs
```
DwyaneShi: seems like planner's and queue's implementations are mixed together, let's split them for better flexibility

apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,546 @@
...
+// reason string ("wait failure", "CreateBatch failure", "cancel-race")
+// appears in the log line so each call site is self-identifying.
+func (q *Planner) releaseAfter(jobID, provisionID, reason string) {
+	if err := q.prov.Release(q.baseCtx, provisionID); err != nil {
+		klog.Warningf("[planner] release after %s job_id=%q provision_id=%q: %v",
+			reason, jobID, provisionID, err)
+	}
+}
+
+func (q *Planner) markFailed(jobID string, err error) {
```
DwyaneShi, line 298: Once a job is marked as failed, we need to release its provision (if valid) as well.

apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,546 @@
...
+func (q *Planner) process(jobID string) {
+	// Atomic check-and-flip Pending → Provisioning.
+	q.mu.Lock()
+	job, ok := q.jobs[jobID]
+	if !ok {
+		q.mu.Unlock()
+		return
+	}
+	if job.state != jobStatePending {
+		// Cancel raced ahead; drop without provisioning.
```
DwyaneShi: didn't get what does the comment mean, let's put a simple example to explain it

apps/console/api/planner/impl/planner.go
```diff
@@ -0,0 +1,546 @@
...
+	}
+	q.mu.Unlock()
+
+	switch state {
+	case jobStatePending, jobStateProvisioning:
+		klog.Infof("[planner] cancel pre-submit job_id=%q state=%d", jobID, state)
+		return &plannerapi.Job{JobID: jobID, Batch: placeholderBatch(req, openai.BatchStatusCancelled, enqueuedAt, terminalAt)}, nil
+	case jobStateSubmitted:
+		klog.Infof("[planner] cancel submitted job_id=%q batch_id=%q", jobID, batchID)
+		batch, err := q.bc.CancelBatch(ctx, batchID)
```
DwyaneShi, line 429: need to release provision

apps/console/api/planner/impl/planner.go
DwyaneShi: got this part after reviewing the cancel function. let's explicitly check if the state is cancelled and write error log if encountering other invalid states.

apps/console/api/planner/client/submission.go
nwangfw: Added

apps/console/api/planner/client/submission.go
nwangfw: Sure. Comments here are added for every field here.

apps/console/api/planner/impl/planner.go
nwangfw: Great point and have split the queue implementation into a separate file.

apps/console/api/planner/impl/planner.go
nwangfw: Yes, the current code will release the provision for if a job has been marked as failed under the process().

apps/console/api/planner/impl/planner.go
nwangfw: Added an example here regarding what does this statement mean. The state info is logged so that we can figure out the exact reason.

apps/console/api/planner/impl/planner.go
nwangfw: Great catch. Release provision added.

apps/console/api/planner/impl/queue.go
```diff
@@ -0,0 +1,62 @@
...
+// errQueueClosed reports that the planner is shutting down and no longer
+// accepts new work.
+var errQueueClosed = errors.New("job queue closed")
+
+// queueCapacity caps the buffered job queue. When full, Push blocks on the
+// caller's context.
+const queueCapacity = 256
+
+// jobQueue owns the buffered FIFO of pending JobIDs.
+type jobQueue struct {
```
DwyaneShi: let's have a queue interface first and this impl as a fifo queue implementation. do not use the name of `jobQueue`, a name describing its functionality is preferred since we will have multiple queue implementations in the future.

apps/console/api/planner/impl/planner.go
Jeffwan: job state can not be used by other planner implmenetation? I think this should be common? I didn't get this part, if you have a new planner implementation, then you will have other jobstate definition?

apps/console/api/planner/impl/queue.go
nwangfw: Make sense. How about this version?

apps/console/api/planner/impl/async_planner.go
Jeffwan: >  The jobs in MDS will be displayed after that.

I didn't get this part. there's logical job shown on UI. Since you manage the job state, you can do the merge whithin planner?

apps/console/api/planner/impl/planner.go
Jeffwan: how to measure it, what if we switch ot other cloud providers? I just like to know 2mins is from some of your test or a magic number you just defined?

apps/console/api/planner/impl/async_planner.go
Jeffwan: I mean this is not defined in earlier PR?

apps/console/api/planner/impl/async_planner.go
Jeffwan: could you point it out? I didn't find the logics

apps/console/api/planner/impl/async_planner.go
nwangfw: Planner does the job merge and currently displays sorted local jobs first followed by the sorted MDS job.  Ideally, we don't need to distinguish them and sort them uniformly. I can address this issue in a future PR.

apps/console/api/planner/impl/planner.go
nwangfw: 2 mins are from my local provisioning tests.  I feel that we need to test and config different numbers for different cloud providers.

apps/console/api/planner/impl/planner.go
nwangfw: We only have one planner now.  If we will have multiple planners, I agree that we should make the job state in common.

apps/console/api/planner/impl/async_planner.go
nwangfw: No

apps/console/api/planner/impl/async_planner.go
nwangfw: Is this part of code that you are looking for? 
https://github.com/nwangfw/aibrix/blob/201ef7fc838728db85160e63ca2947e85c8920b0/apps/console/api/planner/impl/planner.go#L155-L159

PR #2216, opened 2026-05-20
chart: add env for controller-manager

rayne-Li:
## Pull Request Description
according to #1349, now controller-manager will curl /metrics on vllm pods by interval `AIBRIX_POD_METRIC_REFRESH_INTERVAL_MS` in `pkg/cache/cache_metrics.go`, so we need to add env in chart to control that

- cmd/controllers/main.go
```
	if features.IsControllerEnabled(features.ModelAdapterController) {
		// cache is enabled for model adapter scheduling.
		cache.InitWithOptions(config, stopCh, cache.InitOptions{})
		// All options default to zero values (false, nil, nil)
	}
```

- and ModelAdapterController is enabled by default
```
// EnableAllControllers is used to enable all known controllers.
func EnableAllControllers() {
	// This should be updated to reflect all available controllers.
	EnabledControllers[PodAutoscalerController] = true
	EnabledControllers[ModelAdapterController] = true
	EnabledControllers[DistributedInferenceController] = true
	EnabledControllers[ModelRouteController] = true
	EnabledControllers[KVCacheController] = true
	EnabledControllers[StormServiceController] = true
}
```

### DIFF
--- a/dist/chart/templates/controller-manager/deployment.yaml
+++ b/dist/chart/templates/controller-manager/deployment.yaml
@@ -30,9 +30,10 @@ spec:
         env:
         - name: AIBRIX_GATEWAY_TIMEOUT_SECONDS
           value: "{{ .Values.controllerManager.gatewayTimeoutSeconds }}"
-{{- range $key, $val := .Values.controllerManager.container.envs }}
+{{- $envs := .Values.controllerManager.container.envs }}
+{{- range $key := keys $envs | sortAlpha }}
         - name: {{ $key }}
-          value: "{{ $val }}"
+          value: {{ index $envs $key | quote }}
 {{- end }}
         image: {{ .Values.controllerManager.container.image.repository }}:{{ .Values.controllerManager.container.image.tag }}
         imagePullPolicy: {{ .Values.controllerManager.container.image.imagePullPolicy }}
--- a/dist/chart/templates/controller-manager/deployment.yaml
+++ b/dist/chart/templates/controller-manager/deployment.yaml
@@ -30,6 +30,10 @@ spec:
         env:
         - name: AIBRIX_GATEWAY_TIMEOUT_SECONDS
           value: "{{ .Values.controllerManager.gatewayTimeoutSeconds }}"
+{{- range $key, $val := .Values.controllerManager.container.envs }}
+        - name: {{ $key }}
+          value: "{{ $val }}"
+{{- end }}
         image: {{ .Values.controllerManager.container.image.repository }}:{{ .Values.controllerManager.container.image.tag }}
         imagePullPolicy: {{ .Values.controllerManager.container.image.imagePullPolicy }}
         name: manager
--- a/dist/chart/templates/controller-manager/deployment.yaml
+++ b/dist/chart/templates/controller-manager/deployment.yaml
@@ -32,7 +32,7 @@ spec:
           value: "{{ .Values.controllerManager.gatewayTimeoutSeconds }}"
 {{- range $key, $val := .Values.controllerManager.container.envs }}
         - name: {{ $key }}
-          value: "{{ $val }}"
+          value: "{{ $val | quote }}"
 {{- end }}
         image: {{ .Values.controllerManager.container.image.repository }}:{{ .Values.controllerManager.container.image.tag }}
         imagePullPolicy: {{ .Values.controllerManager.container.image.imagePullPolicy }}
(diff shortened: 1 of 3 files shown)


rayne-Li:
@varungup90 could you please help me to find why chat install ci always failed? 

do we lost "webhook-configuration" creation in CI?

```
E0527 02:15:09.334412       1 rotator.go:822] "msg"="Webhook not found. Unable to update certificate." "error"="ValidatingWebhookConfiguration.admissionregistration.k8s.io \"aibrix-validating-webhook-configuration\" not found" "gvk"={"Group":"admissionregistration.k8s.io","Version":"v1","Kind":"ValidatingWebhookConfiguration"} "logger"="cert-rotation" "name"="aibrix-validating-webhook-configuration"
E0527 02:15:09.334534       1 rotator.go:822] "msg"="Webhook not found. Unable to update certificate." "error"="MutatingWebhookConfiguration.admissionregistration.k8s.io \"aibrix-mutating-webhook-configuration\" not found" "gvk"={"Group":"admissionregistration.k8s.io","Version":"v1","Kind":"MutatingWebhookConfiguration"} "logger"="cert-rotation" "name"="aibrix-mutating-webhook-configuration"
```

varungup90:
@rayne-Li helm chart CI is broken right now. Will review and merge it shortly.

varungup90:
Thanks for putting this together! I’ve reviewed the changes and have a few observations and suggestions to help make this implementation a bit more robust:

### ⚠️ Issues & Suggestions

**1. Redundant and Fragile Default Value**
The Go code already has a hardcoded fallback (`defaultPodMetricRefreshIntervalInMS = 50`). Setting this same default explicitly in the chart's `values.yaml` introduces a couple of minor issues:

* Users will always get this environment variable injected, even if they don't explicitly need to tune it.
* If the Go default ever changes in the future, the chart's `values.yaml` will silently override it, causing a drift between Helm deployments and raw Kustomize deployments.
* **Suggestion:** Consider omitting this from `values.yaml` defaults (leaving `envs: {}`) and documenting it as a comment instead:
```yaml
envs: {}
  # AIBRIX_POD_METRIC_REFRESH_INTERVAL_MS: "50"  # Uncomment to tune for high-latency or low-traffic environments

```



**2. Duplicate Environment Variable Risk (`AIBRIX_GATEWAY_TIMEOUT_SECONDS`)**
The template currently has two injection paths: a hardcoded path for `AIBRIX_GATEWAY_TIMEOUT_SECONDS` and a generic loop for the `envs` map. If a user accidentally adds `AIBRIX_GATEWAY_TIMEOUT_SECONDS` to the `envs` block, it will create a duplicate environment variable in Kubernetes. While this is a pre-existing design inconsistency, this PR widens the surface area for it.

* **Suggestion:** Add a brief note in `values.yaml` clarifying that `AIBRIX_GATEWAY_TIMEOUT_SECONDS` is managed exclusively via `controllerManager.gatewayTimeoutSeconds` and should not be added to the `envs` map.

**3. Silent Acceptance of Non-Matching Keys in Schema**
In `values.schema.json`, the regex pattern requires keys to be uppercase (`"^[A-Z_][A-Z0-9_]*$"`). However, without setting `"additionalProperties": false`, invalid keys (like lowercase ones) will be silently accepted rather than rejected by the schema. This matches the existing behavior of `gatewayPlugin`, so it is consistent, but it is worth noting in case we want stricter validation.

**4. Non-Deterministic Environment Variable Ordering**
Using `range` over a map (dict) in Helm produces a non-deterministic order when generating the manifests. While this won't cause functional bugs, it can produce noisy diffs in GitOps tools like ArgoCD or when running `helm diff`. This also matches the existing `gatewayPlugin` behavior, but is something to be aware of.

**5. Missing Context in PR Description**
The PR description mentions that the controller-manager will now curl `/metrics` on vLLM pods, but it doesn't link to the feature PR or commit that introduced this behavior.

* **Suggestion:** Adding a link to the relevant PR/commit would be incredibly helpful for reviewers trying to understand the motivation behind this change.

### 🔍 Minor Observation

* **YAML Hierarchy:** Just a quick sanity check—ensure that the `envs` key in `values.yaml` is properly indented as a sibling to `probes` under the `container` key, rather than being accidentally nested inside `probes`. The diff context looks correct, but it’s worth verifying in the full file!

rayne-Li:
@varungup90 
i have fixed 5 points as you mentioned

rayne-Li:
@varungup90 please take a look at this 😄

PR #2234, opened 2026-05-27
chart: optimize redis passwd logic in helper.tpl

rayne-Li:
## Pull Request Description

in https://github.com/vllm-project/aibrix/pull/2230, we use `get .Values "metadata"` level by level which create may temp variable make helpers.tpl complicated

as gemini bot suggest, now we use `dig` to make it simple

### DIFF
--- a/dist/chart/README.md
+++ b/dist/chart/README.md
@@ -42,6 +42,27 @@ resolve the issue and retry until you get:
 1 chart(s) linted, 0 chart(s) failed
 ```
 
+### helm unittest
+
+test cases in dist/chart/tests
+```
+helm plugin install https://github.com/helm-unittest/helm-unittest.git
+
+helm unittest dist/chart
+### Chart [ aibrix ] dist/chart
+
+ PASS  Test Redis Configuration Logic   dist/chart/tests/redis_config_test.yaml
+ PASS  Test Redis Dependency Passwd Logic       dist/chart/tests/redis_config_test.yaml
+ PASS  Test Redis Shared Passwd Conflict        dist/chart/tests/redis_config_test.yaml
+ PASS  Test Redis Shared Passwd dist/chart/tests/redis_config_test.yaml
+
+Charts:      1 passed, 1 total
+Test Suites: 4 passed, 4 total
+Tests:       4 passed, 4 total
+Snapshot:    0 passed, 0 total
+Time:        14.85875ms
+```
+
 ### Render yaml files
 
 Render all manifests using:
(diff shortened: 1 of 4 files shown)


varungup90:
Thanks for putting this refactor together! Using `dig` definitely cleans up the nested value retrieval nicely.

I reviewed the logic changes and verified that the `dig` nil-safety with top-level keys is completely safe in Helm 3. Additionally, the `default $sharedPass $gatewayPass` precedence in `redis.yaml` perfectly matches the original component-password vs. shared-password fallback logic. ✅

I do have a few suggestions to polish this up before we merge:

### 1. Refine Inline Comments

Some of the inline comments are a bit rough or unnecessary. I recommend updating them for better readability:

* **Change:** `{{- /* is not use builtInRedis，must set external host */ -}}`
**To:** `{{- /* When builtIn Redis is disabled, all components must declare an external Redis host */ -}}`
* **Change:** `{{- /* passwd cannot be empty when sharedEnablePassword */ -}}`
**To:** `{{- /* Shared Redis password must be non-empty when enablePassword is true */ -}}`
* **Remove:** `{{- /* get value by dig */ -}}` (The `dig` calls are self-explanatory, so this comment doesn't add much value).

### 2. Add Helm Unit Tests

Since this PR alters core configuration logic (and builds on the recent changes in #2230), it would be highly valuable to add tests before this lands. Even a basic `helm-unittest` or a `helm template` golden file comparison would go a long way in catching future regressions.

### 3. Minor Readability Tweaks (Optional)

For the `{{- if and $builtInRedisEnabled (or ...` block, it might be helpful to add a line-continuation comment clarifying the three-path `OR` logic (`metadataService` / `gatewayPlugin` / `gpuOptimizer`) for future readers.

rayne-Li:
@varungup90 thanks for your review, i've changed comments and add helm unittest

```
helm unittest dist/chart

### Chart [ aibrix ] dist/chart

 PASS  Test Redis Configuration Logic   dist/chart/tests/redis_config_test.yaml
 PASS  Test Redis Dependency Passwd Logic       dist/chart/tests/redis_config_test.yaml
 PASS  Test Redis Shared Passwd Conflict        dist/chart/tests/redis_config_test.yaml
 PASS  Test Redis Shared Passwd dist/chart/tests/redis_config_test.yaml

Charts:      1 passed, 1 total
Test Suites: 4 passed, 4 total
Tests:       4 passed, 4 total
Snapshot:    0 passed, 0 total
Time:        14.85875ms
```

varungup90 (approved):

PR #2238, opened 2026-05-29
[feat]: Support per-pod KV connector type selection via pod labels

zhutong196:
## Summary

Previously, the KV connector type used by the router was determined only by the global startup configuration, which meant all PD pods in a cluster had to use the same KV transfer backend.

This PR adds support for selecting the KV connector type dynamically from pod labels, allowing different PD pods within the same cluster to use different KV transfer implementations.

For example:

* Some pods can use `nixl`
* Other pods can use `shfs`

The router now checks the pod label:

```text
kv-connector-type
```

and uses it to determine the KV transfer behavior for the target pod.

If the label is absent or invalid, the router falls back to the existing global configuration behavior to preserve backward compatibility.

---

## Motivation

Previously, the connector type was effectively cluster-wide static configuration:

```text
startup env -> all PD pods use same connector
```

This caused several limitations:

* Impossible to run mixed connector deployments
* Difficult to perform gradual migration between connector backends
* Hard to compare or benchmark connector implementations in the same cluster
* Prevented heterogeneous PD deployments

This PR enables pod-level connector selection:

```text
pod label -> connector behavior
```

which makes the routing layer more flexible and extensible.

---

### New Pod Label

```text
kv-connector-type
```

Supported values currently include:

* `nixl`
* `shfs`

---


varungup90:
Thanks for the PR! This is a great feature for enabling mixed PD deployments. I've reviewed the code and have a few pieces of feedback, including a logging bug and some suggestions for robustness and testing.

### 🐛 Bug: Stale global value still logged after per-pod selection

In `updateRoutingContextWithKVTransferParams`, the logic correctly uses `kvConnectorType` (the resolved per-pod value), but the log statements on lines 376 and 403 still log `aibrixKVConnectorType` (the global value). When a pod label overrides the global config, the logs will show the wrong connector type, which makes debugging and observability unreliable.

```go
// Line 376 — logs global, not resolved per-pod value
klog.InfoS("updated routing context with disagg_prefill_resp (NIXL mode)",
    "kv_connector_type", aibrixKVConnectorType)  // <-- should be kvConnectorType

// Line 403 — same issue
klog.InfoS("updated routing context with kv_transfer_params (SHFS mode)",
    "kv_connector_type", aibrixKVConnectorType)  // <-- should be kvConnectorType

```

---

### 💡 Suggestions for Improvement

**1. Label key is not Kubernetes-namespaced**
The label key `kv-connector-type` is a bare, unqualified string. Kubernetes best practices recommend prefixing custom labels with a domain (e.g., `aibrix.ai/kv-connector-type`). Bare labels risk silent collisions with other controllers, admission webhooks, or user-applied labels.

**2. No case normalization on label value**
A label value of `NIXL` or `Shfs` falls through to the default (global config) without warning. A `strings.ToLower()` call, or at minimum an error log on unrecognized values, would improve robustness:

```go
func selectKvConnectorType(value string) string {
    switch strings.ToLower(value) {
    case KVConnectorTypeNIXL, KVConnectorTypeSHFS:
        return strings.ToLower(value)
    default:
        if value != "" {
            klog.Warningf("unrecognized kv-connector-type label %q, falling back to global config", value)
        }
        return aibrixKVConnectorType
    }
}

```

**3. Missing Tests**
`selectKvConnectorType` has no unit tests. Given the fallback logic is the core of this feature, it warrants at minimum a table-driven test covering: valid `nixl`, valid `shfs`, empty string, unrecognized string, and case variants.

---

### 🧹 Minor / Nits

* **Go doc comment:** The comment on `selectKvConnectorType` should start with the function name per Go conventions: `// selectKvConnectorType returns...` (not `// selectKvConnectorType Allow...`).
* **Avoid re-reading labels:** `preparePrefillPayload` and `updateRoutingContextWithKVTransferParams` each independently call `selectKvConnectorType`. If these are called in the same request path, consider resolving the connector type once higher up and threading it down, rather than re-reading the pod label twice.

zhutong196:
> Thanks for the PR! This is a great feature for enabling mixed PD deployments. I've reviewed the code and have a few pieces of feedback, including a logging bug and some suggestions for robustness and testing.
> 
> ### 🐛 Bug: Stale global value still logged after per-pod selection
> In `updateRoutingContextWithKVTransferParams`, the logic correctly uses `kvConnectorType` (the resolved per-pod value), but the log statements on lines 376 and 403 still log `aibrixKVConnectorType` (the global value). When a pod label overrides the global config, the logs will show the wrong connector type, which makes debugging and observability unreliable.
> 
> ```go
> // Line 376 — logs global, not resolved per-pod value
> klog.InfoS("updated routing context with disagg_prefill_resp (NIXL mode)",
>     "kv_connector_type", aibrixKVConnectorType)  // <-- should be kvConnectorType
> 
> // Line 403 — same issue
> klog.InfoS("updated routing context with kv_transfer_params (SHFS mode)",
>     "kv_connector_type", aibrixKVConnectorType)  // <-- should be kvConnectorType
> ```
> 
> ### 💡 Suggestions for Improvement
> **1. Label key is not Kubernetes-namespaced** The label key `kv-connector-type` is a bare, unqualified string. Kubernetes best practices recommend prefixing custom labels with a domain (e.g., `aibrix.ai/kv-connector-type`). Bare labels risk silent collisions with other controllers, admission webhooks, or user-applied labels.
> 
> **2. No case normalization on label value** A label value of `NIXL` or `Shfs` falls through to the default (global config) without warning. A `strings.ToLower()` call, or at minimum an error log on unrecognized values, would improve robustness:
> 
> ```go
> func selectKvConnectorType(value string) string {
>     switch strings.ToLower(value) {
>     case KVConnectorTypeNIXL, KVConnectorTypeSHFS:
>         return strings.ToLower(value)
>     default:
>         if value != "" {
>             klog.Warningf("unrecognized kv-connector-type label %q, falling back to global config", value)
>         }
>         return aibrixKVConnectorType
>     }
> }
> ```
> 
> **3. Missing Tests** `selectKvConnectorType` has no unit tests. Given the fallback logic is the core of this feature, it warrants at minimum a table-driven test covering: valid `nixl`, valid `shfs`, empty string, unrecognized string, and case variants.
> 
> ### 🧹 Minor / Nits
> * **Go doc comment:** The comment on `selectKvConnectorType` should start with the function name per Go conventions: `// selectKvConnectorType returns...` (not `// selectKvConnectorType Allow...`).
> * **Avoid re-reading labels:** `preparePrefillPayload` and `updateRoutingContextWithKVTransferParams` each independently call `selectKvConnectorType`. If these are called in the same request path, consider resolving the connector type once higher up and threading it down, rather than re-reading the pod label twice.


Thanks for the detailed review and suggestions!

Addressed in the latest update:

* Fixed the stale logging issue to use the resolved per-pod `kvConnectorType`
* Updated the label key to a Kubernetes-namespaced format
* Added case-insensitive handling and warning logs for unrecognized values
* Added unit tests covering fallback and case variants
* Updated the Go doc comment to follow Go conventions

For the repeated label lookup suggestion: I looked into resolving the connector type once higher up and passing it through the sync prefill path. I agree the approach could improve consistency, but in the current implementation it would require expanding callback signatures and touching multiple call sites in `handleSyncPrefill`.

Since connector resolution is currently a lightweight label lookup and the value is not expected to change during request processing, I kept the lookup local in this PR to keep the change scope focused. I think this could be a reasonable follow-up refactor if the routing flow becomes more complex in the future.

Please take another look — thanks again for the review!

varungup90 (approved):

PR #2239, opened 2026-05-29
refactor[planner] split provider-agnostic core from planner

nwangfw:
## Pull Request Description

The PRs aim to refactor planner core logics to support different providers to develop their implementation without any conflict.

### DIFF
--- a/apps/console/api/planner/impl/backend.go
+++ b/apps/console/api/planner/impl/backend.go
@@ -51,12 +51,6 @@ type provisionResponseLogger interface {
 	LogProvisionResponse(jobID string, prov *rmtypes.ProvisionResult, spec rmtypes.ResourceProvisionSpec)
 }
 
-// provisionOverride is an optional capability to bypass the real RM call
-// and supply a ProvisionResult directly.
-type provisionOverride interface {
-	TryProvisionOverride(req *plannerapi.EnqueueRequest) (*rmtypes.ProvisionResult, bool)
-}
-
 // backendFactory constructs a plannerBackend for a provisioner type.
 type backendFactory func(prov provisioner.Provisioner) plannerBackend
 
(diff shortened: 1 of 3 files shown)


apps/console/api/planner/impl/backend.go
```diff
@@ -0,0 +1,147 @@
...
+func (b *defaultPlannerBackend) Schedule(_ context.Context, req *plannerapi.EnqueueRequest) (rmtypes.ResourceProvisionSpec, string, int, error) {
+	spec := rmtypes.ResourceProvisionSpec{
+		Credential: rmtypes.ResourceCredential{Provider: b.provider},
+	}
+	if req == nil || req.ModelTemplate == nil {
+		return spec, "", 0, nil
+	}
+	gpuType, gpusPerReplica, err := decodeAcceleratorFromTemplate(req.ModelTemplate)
+	if err != nil {
+		return rmtypes.ResourceProvisionSpec{}, "", 0, err
```
DwyaneShi: let's return nil instead of rmtypes.ResourceProvisionSpec{}

apps/console/api/planner/impl/planner.go
```diff
@@ -174,28 +177,58 @@ func (q *Planner) process(jobID string) {
...
 		return
 	}
-	q.mu.Lock()
-	q.jobs[jobID].provisionID = provResult.ProvisionID
-	q.mu.Unlock()
+
+	// Optional backend override can bypass real RM provisioning.
+	var provResult *rmtypes.ProvisionResult
+	overrode := false
+	if override, ok := q.backend.(provisionOverride); ok {
```
DwyaneShi: Left an offline message regarding the purpose of provisionOverride

nwangfw: Had the offline discussion regarding to have mock backend rather than adding this override logic here.  Remove the overwrite logic entirely in this PR.

apps/console/api/planner/impl/backend.go
nwangfw: Used named return + bare return to avoid the redundant rmtypes.ResourceProvisionSpec{} on the error path. Let me know if that addresses this issue.

DwyaneShi (approved):
LGTM

PR #2330, opened 2026-06-12
Fix BlockHashes cache key identity

### LINKED ISSUE #2329: BlockHashes cache keys can conflate distinct KV cache entries
# AIBrix Bug Report

## 🐛 Describe the bug

`BlockHashes` cache keys can collapse distinct KV cache entries when different block-hash sequences share the same last block hash and block count.

In `python/aibrix_kvcache/aibrix_kvcache/cache_hashable.py`, `BlockHashes.__eq__` compares only the final block hash and the number of blocks, and `BlockHashes.__hash__` hashes only the final block hash plus the number of blocks. `BlockCacheKey` then hashes and compares only `query`, ignoring `prefix`. As a result, L1 cache entries built from different prefixes or different earlier block hashes can be treated as the same key.

This is security-relevant for KV cache reuse. In multi-tenant or shared-cache deployments, an incorrect cache hit can reuse KV tensors from a different prompt/model context, potentially causing response contamination or leakage of another request's prompt context. The built-in Python `hash(str)` also makes the hash value non-deterministic across interpreter processes, although the equality confusion is the main issue.

Affected code paths observed:

- `python/aibrix_kvcache/aibrix_kvcache/cache_hashable.py`
  - `BlockHashes.__eq__`
  - `BlockHashes.__hash__`
  - `BlockCacheKey.__hash__`
  - `BlockCacheKey.__eq__`
- `python/aibrix_kvcache/aibrix_kvcache/l1/l1_cache.py`
  - `L1Cache._put_mrs_impl`
  - `L1Cache.acquire`
  - `L1Cache.exists`
  - `L1Cache.delete`
- `python/aibrix_kvcache/aibrix_kvcache/l2/l2_cache.py`
  - block-hash mode in `_cache_block_keys`

## Steps to Reproduce

1. Use the `BlockHashes` API path with two different block-hash sequences that have the same final block hash and the same block count:

   ```python
   from aibrix_kvcache.cache_hashable import BlockHashes, KVCacheKey

   block_ntokens = 16

   a = BlockHashes(["hA", "hB", "hC"], block_ntokens)
   b = BlockHashes(["hX", "hY", "hC"], block_ntokens)

   assert a != b  # expected, but this currently fails
   ```

2. Build per-block cache keys as L1 does, where the first two hashes are prefix and the last hash is the queried block:

   ```python
   ka = KVCacheKey(a[:32], a[32:48])
   kb = KVCacheKey(b[:32], b[32:48])
   ```

3. Observe that the keys compare equal and collide in a Python dictionary:

   ```python
   print(a == b)
   print(hash(a) == hash(b))
   print(ka == kb)
   print(hash(ka) == hash(kb))

   cache = {ka: "cached KV for sequence hA,hB,hC"}
   print(cache.get(kb))
   ```

4. Actual behavior:

   ```text
   BlockHashes equal: True
   BlockHashes hashes equal: True
   KVCacheKey equal: True
   KVCacheKey hashes equal: True
   Lookup with different prefix/same final block: cached KV for sequence hA,hB,hC
   ```

5. The same issue is reachable through L1 cache operations because `L1Cache._put_mrs_impl` stores `KVCacheKey(block_prefix, block_query)`, while `L1Cache.acquire`, `exists`, and `delete` look up keys using the same equality semantics.

6. In L2 block-hash mode, `_cache_block_keys` builds backend keys from only the current `BlockHashes` slice:

   ```python
   "".join(query[i : i + self.block_ntokens]).encode("utf-8")
   ```

   This also omits prefix, position, model id, and tenant identity unless the caller has already encoded all of those into each block hash.

## Expected behavior

Distinct KV cache entries should not compare equal unless their security-relevant identity is the same.

For block-hash keys, the equality and hash material should include enough information to distinguish:

- full block-hash sequence or cumulative prefix commitment,
- block position,
- prefix/query boundary where relevant,
- model identity or namespace,
- tenant/request isolation material when used in multi-tenant deployments.

At minimum, the cache layer should not silently assume that the caller-provided last block hash already commits to all prior blocks, model identity, and block position without enforcing or validating that invariant.

## Environment

- AIBrix version: repository commit `0b026afe50a3f967e2d60e688945ae96`
- Deployment environment: source checkout / static audit
- Component: `python/aibrix_kvcache`
- Cache mode: `BlockHashes` public API path
- Default-related note: `AIBRIX_KV_CACHE_OL_TOKEN_VALIDATION_ENABLED` defaults to disabled, so compact layout is used and block-hash mode does not have token validation as a fallback.
- Existing issue search: no matching issue found for `BlockHashes cache_hashable`, `KVCache BlockHashes`, or `cache_hashable BlockCacheKey` via `gh issue list`.

3em0:
## Summary
- compare and hash BlockHashes using the full block hash sequence and block size
- include prefix context in BlockCacheKey identity
- build L2 BlockHashes backend keys from the full accumulated block sequence digest
- add regression coverage for the BlockHashes collision case

Fixes #2329

## Testing
- python -m py_compile python/aibrix_kvcache/aibrix_kvcache/cache_hashable.py python/aibrix_kvcache/aibrix_kvcache/l2/l2_cache.py python/aibrix_kvcache/tests/test_l1cache.py python/aibrix_kvcache/tests/test_l2cache.py
- lightweight local reproduction confirming distinct BlockHashes/KVCacheKey values no longer collide

## Not run
- python -m pytest python/aibrix_kvcache/tests/test_l1cache.py::test_block_hashes_use_full_sequence_for_identity python/aibrix_kvcache/tests/test_l1cache.py::test_block_cache_key_includes_prefix python/aibrix_kvcache/tests/test_l1cache.py::test_block_hashes_hash_is_stable_across_processes python/aibrix_kvcache/tests/test_l2cache.py::test_block_hash_l2_keys_include_prefix (local environment is missing test/runtime dependencies such as redis/torch)

### DIFF
--- a/python/aibrix_kvcache/aibrix_kvcache/cache_hashable.py
+++ b/python/aibrix_kvcache/aibrix_kvcache/cache_hashable.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import hashlib
 from abc import ABC, abstractmethod
 from typing import List, Sequence, TypeAlias
 
@@ -19,7 +20,6 @@ import numpy as np
 from farmhash import FarmHash32
 
 from .common import CachedPyObjectBase
-from .utils import hash_combine_128
 
 
 class KVCacheHashable(ABC):
@@ -230,6 +230,22 @@ class TokenListView(CachedPyObjectBase, KVCacheIds):
 BlockHashList: TypeAlias = List[str]
 
 
+def _digest_block_hashes(data: BlockHashList, block_ntokens: int) -> bytes:
+    hasher = hashlib.blake2b(digest_size=16)
+    hasher.update(block_ntokens.to_bytes(8, "little", signed=False))
+    hasher.update(len(data).to_bytes(8, "little", signed=False))
+    for block_hash in data:
+        encoded = block_hash.encode("utf-8")
+        hasher.update(len(encoded).to_bytes(8, "little", signed=False))
+        hasher.update(encoded)
+    return hasher.digest()
+
+
+def _hash_block_hashes(data: BlockHashList, block_ntokens: int) -> int:
+    digest = _digest_block_hashes(data, block_ntokens)
+    return int.from_bytes(digest[:8], "little", signed=False)
+
+
 def is_block_hash_list_type(data) -> bool:
     if not isinstance(data, List):
         return False
@@ -306,20 +322,19 @@ class BlockHashes(CachedPyObjectBase, KVCacheIds):
     def __eq__(self, other):
         if not isinstance(other, BlockHashes):
             return False
-        # we assume a block hash includes the information of both position and
-        # all tokens belonging to this block, therefore, we only need to check
-        # the last block hash and num of blocks to see if two BlockHashes are
-        # the same
-        return self._data[-1] == other._data[-1] and len(self._data) == len(
-            other._data
+        return (
+            self.block_ntokens == other.block_ntokens
+            and self._data == other._data
         )
 
     def __ne__(self, value):
         return not self.__eq__(value)
 
     def __hash__(self) -> int:
-        # use the last block hash and num of blocks for calculating the hash
-        return hash_combine_128(hash(self._data[-1]), len(self._data))
+        return _hash_block_hashes(self._data, self.block_ntokens)
+
+    def to_bytes(self) -> bytes:
+        return _digest_block_hashes(self._data, self.block_ntokens)
 
 
 class TokenCacheKey(KVCacheHashable, CachedPyObjectBase):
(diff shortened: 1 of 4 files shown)


DwyaneShi:
@3em0 Thanks for the issue report and the PR. We agree that incorrect KV cache reuse would be security-relevant, so we reviewed the concern carefully.

However, the reported collision scenario does not apply to the current `BlockHashes` design. The key assumption here is that each block hash is constructed as a chained hash, similar to vLLM: the current block hash already incorporates the prefix / previous block hash context. Therefore, the final block hash plus the block count is intended to uniquely identify the full block sequence. Two sequences with the same last block hash and block count should not represent different KV entries unless the chained-hash invariant itself is broken.

That said, the PR highlights that this invariant is not obvious from the code. We should improve the comments / documentation around `BlockHashes` to make it clear that block hashes are chained and prefix-dependent.

For now, we are going to close this PR because the proposed change is not needed for correctness under the current hashing model.

PR #2381, opened 2026-06-19
[Misc] Refactor simplify client.MatchingLabelsSelector as client.MatchingLabels

zedongh:
## Pull Request Description
Here are some selector construct not necessary, just using client.MatchingLabels.

## Related Issues
Noop, just take the best practise

### DIFF
--- a/pkg/controller/podset/podset_controller.go
+++ b/pkg/controller/podset/podset_controller.go
@@ -28,9 +28,7 @@ import (
 	apiequality "k8s.io/apimachinery/pkg/api/equality"
 	apierrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/runtime"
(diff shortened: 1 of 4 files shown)


googs1025 (commented):
LGTM. This is an equivalent simplification for exact-match label selectors

googs1025 (approved):

varungup90:
A few notes and observations from reviewing these changes:

* **Removed pre-existing risks:** In `podset_rollsyncer.go:452` and `sync.go:174`, the original code used `labels.NewRequirement(..., _)`, which silently ignored errors. This PR correctly removes those patterns, meaning the new code avoids that pre-existing risk entirely.
* **Accurate semantics:** `client.MatchingLabels` produces an exact equality selector (matching the behavior of `selection.Equals`). This is the correct approach for all uses here, as resource names (which are valid label values) are being matched.
* **Test coverage:** The PR doesn't update tests, but since this is a mechanical refactor with identical behavior, that's completely acceptable. Existing tests should adequately cover correctness.

Overall, this is a clean and safe simplification!

Jeffwan (approved):

Issue #2287, opened 2026-06-08
Prefix cache index not invalidated when a vLLM engine clears its prefix cache (reset/OOM/sleep)


Jeffwan:
### Describe the bug

When an engine clears its prefix cache — `/reset_prefix_cache`, OOM-driven reset, or `/sleep level>=1` — it wipes the cache and enqueues `AllBlocksCleared`. We don't act on it: `handleAllBlocksCleared` is a no-op (`pkg/kvevent/handler.go`), and the only caller of `RemovePrefix` is pod unsubscribe, which doesn't fire here (the pod stays `Running`, `/health` still returns 200). So stale entries linger, and because the prefix router prefers the pod with the most matching blocks, it actively routes prefix traffic to the pod whose cache was just wiped — a cold miss.

The event alone isn't a reliable fix: vLLM publishes queued KV events only from inside a scheduler step (`update_from_output`), so an idle/sleeping engine may never flush `AllBlocksCleared`, and ZMQ delivery is lossy.

### Steps to Reproduce

1. Two replicas with the kv-event prefix router; warm a shared prefix on pod A.
2. `POST /reset_prefix_cache` on pod A (or `/sleep?level=1`).
3. Send the same prefix again → still routed to A, now a cold miss.

### Expected behavior

- `handleAllBlocksCleared` calls `RemovePrefix(model, loraID, podKey)` (already on the `SyncIndexer` interface).
- A reconcile against the scraped `vllm:engine_sleep_state` metric backstops dropped events (purge entries for pods reporting non-awake).
- When AIBrix drives sleep itself (separate sleep-mode RFC), it should also purge synchronously on `/sleep` — tracked there.

### Environment

AIBrix: main · vLLM: main (kv-events enabled).

Jeffwan:
wait until #2290 is implemented

pjdurden:
Opened #2385 for part 1: handleAllBlocksCleared now calls RemovePrefix(model, loraID, podKey) to purge the pod from the prefix index, mirroring handleBlockRemoved. Tests cover the purge call, temporary-error swallow, and error propagation.

Left the metric-driven reconcile backstop (purge against scraped vllm:engine_sleep_state) for a follow-up since it is a larger change and, as you noted, the event alone is not fully reliable. Happy to take that next if the approach in #2385 looks right.

ankit373:
[...]

Issue #2290, opened 2026-06-09
[RFC]: Control-plane orchestration of vLLM sleep mode and kvcached for warm standby and multi-model GPU pooling


Jeffwan:
### Summary

AIBrix scales replicas as 0-or-N and only *observes* engine memory state; it never drives it. Two engine-side mechanisms now make GPU memory reclaimable at runtime — vLLM **sleep mode** (park a whole idle model's weights+KV) and **kvcached** (elastically shrink/grow a model's KV so several models share one GPU) — but both are mechanism-only, with no control plane. This RFC proposes AIBrix own the *policy*: a control-plane-driven "warm/asleep" tier between one replica and zero, plus KV-budget arbitration across co-resident models. The result: multi-minute cold starts become second-scale wakes, and one GPU can densely host many long-tail models.

### Motivation

Scaling is binary today. To reclaim a GPU we either keep a replica running (burning HBM while idle) or delete the pod and pay a multi-minute cold start (weight load + CUDA graph capture) when traffic returns. Neither fits bursty or long-tail traffic.

The engine now exposes the levers, but not the policy:

| | vLLM sleep mode | kvcached (github.com/ovg-project/kvcached) |
|---|---|---|
| Scope | whole model (weights + KV) | KV cache only (weights allocated outside it) |
| Granularity | all-or-nothing, single process | page-level elastic, cross-process co-residency |
| Target | a fully **idle** model | several **active** models sharing a GPU |
| Mechanism | CUDA VMM; `/sleep` `/wake_up`; level 1 weights→CPU, level 2 discard | CUDA VMM; per-process `cudaMemGetInfo` headroom + `/dev/shm` budget via `kvctl` |
| Control | manual API only; `VLLM_SERVER_DEV_MODE`-gated; no autoscaling | decentralized by design; no scheduler/admission/cross-node; `controller/` is a demo |

They are complementary: kvcached handles elastic KV among active co-resident models; sleep parks idle models' weights. Both are VMM actuators that enforce/observe — neither *decides*. The missing piece is a control plane that decides **which** model sleeps/wakes and **when**, **how much** KV budget each co-resident model gets, **admission** when a GPU is oversubscribed, and **routing/readiness** as models move between awake / warm / asleep. AIBrix already scrapes `vllm:engine_sleep_state` but never acts on it, and has no notion of co-resident KV budget — so it is the natural owner.

### Proposed Change

[...]

PR #2385, opened 2026-06-21
fix(kvevent): purge pod prefixes on AllBlocksCleared (#2287)

pjdurden:
**Description**

`handleAllBlocksCleared` was a no-op. When a vLLM engine wipes its entire prefix cache (`/reset_prefix_cache`, an OOM-driven reset, or `/sleep` with `level>=1`) it enqueues `AllBlocksCleared`, but AIBrix ignored it: the pod stays `Running` and `/health` still returns 200, so the pod-unsubscribe path that normally calls `RemovePrefix` never fires. The pod's entries linger in the cross-pod prefix index, and because the prefix router prefers the pod with the most matching blocks, it keeps routing prefix traffic to the pod whose cache was just cleared, turning each match into a cold miss.

This wires `handleAllBlocksCleared` to purge the pod from the index via `RemovePrefix(ctx, modelName, loraID, podKey)` (already on the `SyncIndexer` interface), mirroring `handleBlockRemoved`: temporary errors fetching the indexer are swallowed, real failures propagate.

Scope: this is part 1 of #2287 (act on the event). As the issue notes, events alone are not a complete fix, vLLM only flushes queued KV events from inside a scheduler step, so an idle/sleeping engine may never emit this, and ZMQ delivery is lossy. The metric-driven reconcile backstop (purge against scraped `vllm:engine_sleep_state`) is left as a follow-up and called out in a code comment.

**Related issue**

Part 1 of #2287.

**Tests**

- Rewrote `TestHandleAllBlocksClearedEvent` to assert `RemovePrefix` is called exactly once, scoped to the handler's model/lora/pod.
- Added `TestHandleAllBlocksClearedTemporaryError` (temporary indexer error is swallowed, no purge) and `TestHandleAllBlocksClearedProcessingError` (RemovePrefix failure propagates).
- The mock sync indexer now records `RemovePrefix` calls.

Verified locally (the package is behind the `zmq` build tag):

```
go test -tags zmq ./pkg/kvevent/
go vet  -tags zmq ./pkg/kvevent/
go build ./...
gofmt -l pkg/kvevent/
```

### DIFF
--- a/pkg/kvevent/handler.go
+++ b/pkg/kvevent/handler.go
@@ -132,10 +132,32 @@ func (h *eventHandler) handleBlockRemoved(ctx context.Context, event *kvcache.Bl
 }
 
 func (h *eventHandler) handleAllBlocksCleared(ctx context.Context, event *kvcache.AllBlocksClearedEvent) error {
-	// Not implemented: AllBlocksCleared events are pod-local optimizations that
-	// clear the entire cache for a specific model. These events don't need cross-pod
-	// synchronization as they represent local memory management decisions. Each pod
-	// manages its own cache lifecycle independently based on its memory constraints.
-	klog.V(4).Infof("Received AllBlocksCleared event for pod %s (not implemented)", h.podKey)
+	// The engine wiped its entire prefix cache (/reset_prefix_cache, OOM-driven
+	// reset, or /sleep level>=1). The pod stays Running and /health still returns
+	// 200, so the pod-unsubscribe path that normally calls RemovePrefix never
+	// fires. If we don't purge the pod's entries here, the prefix router keeps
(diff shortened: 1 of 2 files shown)


pjdurden:
Saw the note on #2287 to wait on #2290, so I have marked this as draft.

This PR is only the part-1 event handling: acting on AllBlocksCleared, which vLLM emits today via the existing kv-events path. It does not touch the sleep-mode control plane or the metric reconcile backstop, so it should be independent of the #2290 RFC. Happy to keep it as draft and sequence it after #2290 if you prefer, or land it early as a standalone correctness fix for the no-op handler. Your call on the sequencing.

CarolWinddd (approved):
Code reviewed and no issue found. Implementation logic, error handling, test cases align with existing code conventions.

PR #2395, opened 2026-06-25
feat(pd): roleset eligibility extraction, load-imbalance fixes, and tracker registration ordering

varungup90:
## Summary

This PR makes three related fixes to the PD disaggregation router and adds a dedicated `pd_roleset.go` package for roleset eligibility logic.

### 1. Roleset eligibility extraction (`pd_roleset.go`)

Extracted roleset grouping, replica cardinality, and eligibility checks into a new `pd_roleset.go` file. A roleset is now **eligible** only when it has ≥1 prefill replica **and** ≥1 decode replica. Incomplete or partially-ready rolesets are excluded from routing candidates before any scoring or imbalance checks run. Public functions (`HasEligibleRoleset`, `DescribePDRolesetEligibility`) expose fleet-wide diagnostics for external callers (e.g., gateway pod-readiness logic).

Replica count is computed from distinct `role-replica-index` label values, so multi-node tensor-parallel groups (multiple pods sharing one logical slot) are counted correctly.

### 2. Load-imbalance alignment fixes

Two bugs in the imbalance fast paths:

- **Prefill imbalance**: Previously logged before narrowing, and the guard `isImbalanced` could be true with a nil `targetPod`. Added a nil check before narrowing `prefillPods` and aligning `decodePods`.
- **Decode imbalance**: The previous alignment (`if len(prefillPods) > 1`) skipped alignment when only one prefill pod was present, leaving a stale single-pod slice pointing at the wrong roleset. Now always aligns `prefillPods` to the selected decode pod's roleset, falling back gracefully when the roleset filter produces no results.

### 3. Tracker registration ordering

`AddPrefillRequest` and `AddPendingDecode` were called in `Route()` *before* pod selection completed, meaning concurrent requests saw stale counts during scoring. Registration is now deferred to the end of `filterPrefillDecodePods`, immediately after the winning prefill/decode pods are known, so load imbalance and scoring see up-to-date counts from the moment of selection.

The duplicate `AddPendingDecode` call in `Route()` is removed; cleanup (`RemovePendingDecode`) is still deferred in `Route()` and `RemovePrefillRequest` is now called on sync failure (was previously missing).

Additionally, the per-request `validateAndGetLLMEngine` call at the top of `Route()` is removed — engine validation across all ready pods on every request was expensive and redundant since `ctx.Engine` is already set from pod labels by the gateway before `Route()` is invoked.

## Test changes

- Fixed two async-prefill tests (`TestPDRouter_RouteDoesNotEmitAsyncPrefillSuccessBeforeHTTPCompletes`, `TestPDRouter_RouteRecordsAsyncPrefillFailureWithoutSuccess`) that omitted `ctx.Engine = SGLangEngine`, causing the sync code path to run instead of the async one — the first test timed out waiting for a metric that would never be emitted, the second would incorrectly propagate a 500 as a route error.
- Added `TestFilterPrefillDecodePods_NoBucketMatchNoCombined`, `TestFilterPrefillDecodePods_CombinedPickImbalance`, `TestFilterPrefillDecodePods_CombinedFallbackBucketing`, `TestFilterPrefillDecodePods_SelectCorrectBucketPods` for prompt-length bucketing edge cases.
- Added `TestScorePrefillPods_NilPolicyFallback` for nil prefill policy behaviour.
- Added `pd_roleset_test.go` covering roleset grouping, replica cardinality, eligibility, and `DescribePDRolesetEligibility`.

## Related Issues

Part of ongoing PD disaggregation stabilisation.

### DIFF
--- a/pkg/plugins/gateway/algorithms/pd_disaggregation.go
+++ b/pkg/plugins/gateway/algorithms/pd_disaggregation.go
@@ -516,7 +516,7 @@ func (r *pdRouter) loadImbalanceSelectPrefillPod(readyPods []*v1.Pod, podRequest
 // maxFreeGPUUsage from the same pass (maxima include all pods, including those without metrics).
 // maxRequestCount and maxThroughput are floored at 1 and maxFreeGPUUsage is floored at 1 so
 // that scoreDecodePods normalization denominators are always positive. The caller narrows
-// prefillPods to the selected pod's roleset. KV cache headroom uses KVCacheUsagePerc only
+// prefillPods to the selected pod's roleset. KV cache headroom uses KVCacheUsagePerc
 // (missing metric is treated as 0% usage = 100% free).
 func (r *pdRouter) loadImbalanceSelectDecodePod(ctx *types.RoutingContext, filteredDecodePods []*v1.Pod) (*v1.Pod, float64, float64, float64, map[string]float64, map[string]float64, map[string]float64) {
 	podRequestCounts := make(map[string]float64)
@@ -565,10 +565,11 @@ func (r *pdRouter) loadImbalanceSelectDecodePod(ctx *types.RoutingContext, filte
 		}
 
 		kvUsage, err := r.cache.GetMetricValueByPodModel(pod.Name, pod.Namespace, ctx.Model, metrics.KVCacheUsagePerc)
-		if err != nil {
-			kvUsage = &metrics.SimpleMetricValue{Value: 0}
+		kvUsageVal := float64(0)
+		if err == nil {
+			kvUsageVal = kvUsage.GetSimpleValue()
 		}
-		podFreeGpuUsage[pod.Name] = math.Round(100 - kvUsage.GetSimpleValue()*100)
+		podFreeGpuUsage[pod.Name] = math.Round(100 - kvUsageVal*100)
 		if podFreeGpuUsage[pod.Name] <= 0 {
 			podFreeGpuUsage[pod.Name] = 0.1
 		}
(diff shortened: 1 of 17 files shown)


pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -333,50 +369,88 @@ func (r *pdRouter) filterPrefillDecodePods(routingCtx *types.RoutingContext, rea
...
 	decodeRun := r.scoreDecodePods(routingCtx, decodePods, maxRequestCount, maxThroughput, maxFreeGPUUsage, podRequestCounts, podThroughputs, podFreeGpuUsage, decodePol)
-	return r.finalPDScore(routingCtx, prefixHashes, prefillScores, maxPrefillScore, decodeRun)
+	selectedPrefill, selectedDecode, err := r.finalPDScore(routingCtx, prefixHashes, prefillScores, maxPrefillScore, decodeRun)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	// Register pending prefill/decode immediately after selection so concurrent routing
+	// calls read up-to-date counts when evaluating load imbalance and scoring.
+	r.prefillRequestTracker.AddPrefillRequest(routingCtx.RequestID, selectedPrefill.Name)
```
googs1025:
```go
  r.prefillRequestTracker.AddPrefillRequest(routingCtx.RequestID, selectedPrefill.Name)
```

`DefaultExecutor.Execute()` still calls `e.tracker.AddPrefillRequest(routingCtx.RequestID, prefillPod.Name)` for the same request before dispatching the HTTP prefill.
  `PrefillRequestTracker.AddPrefillRequest()` is not idempotent: the same request increments the pod count twice, while RemovePrefillRequest() can only decrement once because the request-to-pod mapping is deleted on the first removal.

varungup90:
cc @googs1025 @Jeffwan help review when you get a chance

googs1025:
> cc @googs1025 @Jeffwan help review when you get a chance

will help to review tomorrow 😄

googs1025 (approved):
LGTM

PR #2402, opened 2026-06-26
Support in-place role image rollout

### LINKED ISSUE #1827: Feature Request: Support In-Place Update for RoleSet to avoid unnecessary Pod recreation
Currently, the `RoleSet` controller only supports **recreating Pods** during updates (e.g., when the container image changes). This means every update — even a simple image bump — results in:

- Pod deletion and rescheduling  
- Potential IP/Node changes  

For many use cases (e.g., model hot-reloading, inference service version upgrades), **only the container image needs to change**, and the rest of the Pod spec (volumes, env, resources, etc.) remains identical. In such cases, Kubernetes supports **in-place update via strategic merge patch** on `spec.containers[*].image`, which triggers container restart **without recreating the Pod**.

We propose adding native support for **in-place update** in `RoleSet`, similar to [OpenKruise’s CloneSet](https://openkruise.io/docs/next/core-concepts/inplace-update)

---

### Design

1. **Add `type` field to `RoleUpdateStrategy`**:
   ```go
   type RoleUpdateStrategyType string

   const (
       RecreateRoleUpdateStrategyType      RoleUpdateStrategyType = "Recreate"
       InPlaceOnlyRoleUpdateStrategyType   RoleUpdateStrategyType = "InPlaceOnly"
       InPlaceIfPossibleRoleUpdateStrategyType RoleUpdateStrategyType = "InPlaceIfPossible"
   )

   type RoleUpdateStrategy struct {
       Type            RoleUpdateStrategyType     `json:"type,omitempty"`
       MaxUnavailable  *intstr.IntOrString        `json:"maxUnavailable,omitempty"`
       MaxSurge        *intstr.IntOrString        `json:"maxSurge,omitempty"`
   }
   ```

2. **In-place conditions**:
   - Only allowed when **container images differ**, and all other fields (`command`, `args`, `env`, `resources`, `volumeMounts`, etc.) are unchanged.
   - If strategy is `InPlaceIfPossible` and changes are not in-place compatible → fall back to recreate.
   - If strategy is `InPlaceOnly` and changes are incompatible → reject update.

3. **Respect `maxUnavailable` during in-place updates**:
   - Treat **NotReady Pods (during container restart)** as “unavailable”.
   - Limit concurrent in-place updates to ensure `#NotReady <= maxUnavailable`.

4. **Mark updated Pods via label**:

---

---

### Example Usage

```yaml
  ...
  - name: worker
    replicas: 5
    template:
      spec:
        containers:
        - name: app
          image: my-model:v2  # ← only this changes
    updateStrategy:
      type: InPlaceIfPossible
      maxUnavailable: 1
```

googs1025:
## Summary
- Add RoleSet role-level `InPlaceIfPossible` image rollout support for live Pods, preserving Pod identity for image-only updates.
- Wire StormService in-place update through RoleSet roles and emit RoleSet `InPlaceFallback` events when a rollout falls back to recreate.
- Remove the user-facing `InPlaceOnly` strategy to avoid blocking rollouts on unsupported updates.
- Add unit and controller integration coverage for RoleSet and StormService in-place rollout and fallback behavior.

Fixes #1827


config/samples/orchestration_v1alpha1_roleset.yaml
```diff
@@ -6,4 +6,15 @@ metadata:
     app.kubernetes.io/managed-by: kustomize
   name: roleset-sample
 spec:
-  # TODO(user): Add fields here
+  roles:
+  - name: prefill
+    replicas: 1
+    updateStrategy:
+      type: InPlaceOnly
+      maxUnavailable: 1
```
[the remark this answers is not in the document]
googs1025: removed

Jeffwan:
@googs1025 I merged another PR from you and notice there's conflict now. please help rebase main branch

pkg/controller/roleset/inplace_update.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+	roleRevisionNameKey := fmt.Sprintf("%s.%s", constants.RoleRevisionNameAnnotationPrefix, role.Name)
+	if roleRevisionName, ok := roleSet.Annotations[roleRevisionNameKey]; ok {
+		pod.Labels[constants.RoleRevisionNameLabelKey] = roleRevisionName
+	}
+}
+
+// podReadyWithRuntimeImages verifies that kubelet has restarted the containers
+// and is reporting image IDs that match the patched pod spec.
+func podReadyWithRuntimeImages(pod *corev1.Pod) bool {
```
Jeffwan: so far, we do not consider the timeout case,right? technically, it can be stucked.

Jeffwan: this is acceptable, let's leave more comments on the limitaiton for future improvement

pkg/controller/roleset/inplace_update.go
```diff
@@ -0,0 +1,505 @@
...
+	if defaulted := defaultDockerImageName(specImage); defaulted != specImage {
+		candidates = append(candidates, defaulted)
+		if !imageHasTagOrDigest(defaulted) {
+			candidates = append(candidates, defaulted+":latest")
+		}
+	}
+	return candidates
+}
+
+func defaultDockerImageName(image string) string {
```
Jeffwan: em. is this generic enough?

let's document the supported formats and then we can have a review

Jeffwan:
@googs1025 check the comments and let me know your throughts

pkg/controller/roleset/inplace_update.go
googs1025: Addressed. Added comments documenting the no-timeout/no-rollback limitation, and added pending-reason annotation, events, and structured logs so stuck in-place updates are diagnosable.

pkg/controller/roleset/inplace_update.go
googs1025: Updated in d727e96. I removed the runtime image string hard match from completion and now use the Image id change check instead. `status.Image` is kept only in pending logs for debugging, so vLLM/SGLang/Harbor/runtime-normalized image strings no longer block completion.

googs1025:
## RoleSet in-place update flow

```mermaid
flowchart TD
    A[RoleSet spec/template changes] --> B[Role syncer calculates desired template hash]
    B --> C{Role updateStrategy is InPlaceIfPossible?}
    C -- No --> R[Use existing recreate rollout]
    C -- Yes --> D{PodSet role or grouped pods?}
    D -- Yes --> R
    D -- No --> E{Only regular container images changed?}
    E -- No --> F[Fallback to recreate and emit InPlaceFallback event]
    E -- Yes --> G[Select outdated pods within rollout budget]
    G --> H[Patch pod spec.containers image]
    H --> I[Annotate target hash and previous container ImageID state]
    I --> J[Set optional in-place readiness gate to false]
    J --> K[Emit InPlaceUpdateStarted event]
    K --> L[Next reconcile checks pod status]
    L --> M{Pod ready and changed containers ready?}
    M -- No --> P[Record pending reason annotation and structured log]
    M -- Yes --> N{Recorded ImageID changed?}
    N -- No --> P
    N -- Yes --> O[Promote target hash to labels, update revision labels, clear in-place annotations, readiness gate true, emit InPlaceUpdateCompleted]
    P --> L
```

### Normal path

1. A RoleSet role using `InPlaceIfPossible` gets a new template hash.
2. For normal pod roles, the syncer renders the desired pod and checks whether the live pod differs only by regular container image fields. Init container image changes, command/env/resources/metadata changes, PodSet roles, or other pod spec changes do not use in-place update.
3. If eligible, the controller patches `pod.spec.containers[*].image`, records the desired hash in `stormservice.orchestration.aibrix.ai/in-place-update-target-hash`, and stores the previous container `ImageID` in `stormservice.orchestration.aibrix.ai/in-place-update-state`.
4. If the pod uses the optional in-place readiness gate, the controller sets it to false before the runtime restart finishes.
5. On later reconciles, completion is based on pod readiness plus changed container `ImageID`. `status.Image` is kept for diagnostics only and does not block completion, so runtime-specific image string normalization does not affect this path.
6. Once complete, the controller promotes the target hash into the pod labels, updates role revision labels, clears the in-place annotations, sets the readiness gate true if present, and emits `InPlaceUpdateCompleted`.

### Fallback and failure behavior

- If the change is not image-only or the role is not eligible for pod-level in-place update, the controller falls back to the existing recreate rollout and emits `InPlaceFallback`.
- If the image update is patched but the pod does not become ready, the container status is missing, a container is not ready, or the recorded `ImageID` does not change, the rollout stays in progress. The controller writes `stormservice.orchestration.aibrix.ai/in-place-update-pending-reason` with one of `PodNotReady`, `ContainerStatusMissing`, `ContainerNotReady`, or `ImageIDUnchanged`.
- The controller also logs `in-place update pending` with roleset, role, pod, container, desired image, runtime image, old ImageID, current ImageID, and reason. This is the main debug signal for stuck updates.
- There is currently no timeout or automatic rollback for a stuck in-place update. Operators should inspect the pending reason/logs and either fix the image/runtime issue, clear the stuck pod by recreation, or update the workload again after resolving the underlying problem.

pkg/controller/roleset/inplace_update.go
Jeffwan: ok

Jeffwan:
@googs1025 please update the docs as well so user could know this change

PR #2426, opened 2026-07-07
Fix RayClusterReplicaSet scale-down ordering

### LINKED ISSUE #2425: rayfleet scale down support ways to determine the scaling-down priority.

Currently, when scaling down a RayClusterReplicaSet, it directly removes the first N rayclusters in the list. It doesn't support scaling down specific replicas, which might result in deleting ready replicas while keeping unready ones, making the whole service unavailable. We need to support ways to determine the scaling-down priority of rayclusters, like sorting by deletion cost.

### Use Case

When reducing the number of replicas in rayfleet, the cluster list should be sorted in reverse order according to the deletion cost.

### Proposed Solution

_No response_

googs1025:
## Summary
- Sort RayClusterReplicaSet scale-down candidates before deletion.
- Prefer deleting NotReady RayClusters, then lower `controller.kubernetes.io/pod-deletion-cost`, then older creation timestamp, then namespace/name for deterministic tie-breaking.
- Add unit coverage and RayClusterReplicaSet lifecycle integration coverage.

Fixes #2425

## Test Plan
- `go test -count=1 ./pkg/controller/rayclusterreplicaset -run 'TestScaleDown(PrioritizesNotReadyThenLowerDeletionCost|UsesNameAsFinalTieBreaker)'`\n- `go test -count=1 ./pkg/controller/rayclusterreplicaset -run '^$'`\n- `go test -count=1 ./test/integration/controller -run '^$'`\n- `git diff --check`\n\nNote: focused integration execution was attempted locally, but envtest could not start because `bin/k8s/1.30.0-darwin-arm64/etcd` is not present in this workspace.

### DIFF
--- a/pkg/controller/rayclusterreplicaset/rayclusterreplicaset_controller.go
+++ b/pkg/controller/rayclusterreplicaset/rayclusterreplicaset_controller.go
@@ -181,6 +181,7 @@ func (r *RayClusterReplicaSetReconciler) scaleDown(ctx context.Context, replicas
 	const maxConcurrency = 10
 	semaphore := make(chan struct{}, maxConcurrency)
 
+	sortRayClustersForScaleDown(clusters)
 	for i := 0; i < diff; i++ {
 		cluster := clusters[i]
 		semaphore <- struct{}{}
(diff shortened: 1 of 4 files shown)


Jeffwan (approved):
overall lgtm

PR #2428, opened 2026-07-08
feat(gateway-plugin): support bearer token auth and chart config

NelZyhh:
## Background
This PR adds Bearer Token authentication support for the `gateway-plugin` and updates the Helm chart so the auth token can be configured through values.

## Changes
- Add Bearer Token auth support to `gateway-plugin`
- Load the auth token from the `AIBRIX_AUTH_BEARER_TOKEN` environment variable
- Add `gatewayPlugin.auth.bearerToken` to the Helm chart values
- Inject `gatewayPlugin.auth.bearerToken` directly into the container as `AIBRIX_AUTH_BEARER_TOKEN`
- Add chart validation to prevent duplicate configuration with `gatewayPlugin.container.envs.AIBRIX_AUTH_BEARER_TOKEN`
- Add chart test coverage for env injection and validation failure

## Usage
Configure the token in Helm values:

```yaml
gatewayPlugin:
  auth:
    bearerToken: "your-token"
```

This renders the following environment variable in the `gateway-plugin` container:

```yaml
- name: AIBRIX_AUTH_BEARER_TOKEN
  value: "your-token"
```

## Validation
- `helm template` renders successfully
- The bearer token is injected into the `gateway-plugin` container environment
- Duplicate configuration fails as expected

## Notes
- The current implementation injects the token directly as an environment variable, so it will appear in the rendered manifest
- Do not configure both of the following at the same time:
  - `gatewayPlugin.auth.bearerToken`
  - `gatewayPlugin.container.envs.AIBRIX_AUTH_BEARER_TOKEN`


varungup90:
## Code Review & Risk Assessment Summary

### Minor Observations

* **Variable Scope & Readability:** The `presentedAPIKey` variable is populated inside the header-parsing loop and utilized after it. While this works perfectly without any functional issues, it silently accepts the raw header value even if other cases (like `HeaderExternalFilter`) write to `reqHeaders["Authorization"]`. Keeping `authorizationKey` as a distinct case is safe, but separating this more clearly could improve code readability.
* **Copyright Check:** The copyright year in the new files is correctly updated to **2026**, which is fully consistent and accurate.

### Risk Assessment

> **Risk Level:** Low 🟢

* **Isolation:** The feature is completely opt-in. If the environment variable is unset, no behavior changes. It is neatly isolated to a new file and does not impact existing authentication or rate-limiting logic.

### Recommended Follow-Ups (Before Merge)

1. **Test Coverage:** Add a dedicated test case specifically handling a **wrong-token / unauthorized** scenario.
2. **Production Documentation:** Clarify and document secret-based injection methods as an alternative to using plaintext values in production environments.

NelZyhh:
@varungup90 
Thanks for the review and suggestions.

I updated the request header handling to make the auth flow clearer: the header parsing loop now only collects headers, and the `Authorization` value is looked up separately before auth validation using a case-insensitive match. This avoids relying on a loop-scoped auth variable and keeps the auth input explicit.

```go
if s.apiKeyAuth != nil && s.apiKeyAuth.token != "" {
    authHeaderValue := ""
    for key, value := range reqHeaders {
        if strings.EqualFold(key, authorizationKey) {
            authHeaderValue = value
            break
        }
    }
    if !s.apiKeyAuth.isAuthorized(authHeaderValue) {
        ...
    }
}
```

I also added a dedicated unauthorized test case for the wrong-token scenario, in addition to the existing valid-token, missing-token, and non-bearer cases.

For the Helm chart, I switched the bearer token injection to a Secret-backed env var instead of rendering the token directly in the manifest.

```yaml
- name: AIBRIX_AUTH_BEARER_TOKEN
  valueFrom:
    secretKeyRef:
      name: {{ include "aibrix.fullname" . }}-gateway-plugin-auth
      key: bearer-token
```

varungup90 (approved):

PR #2450, opened 2026-07-14
[API] Add explicit StormService deployment mode field

### LINKED ISSUE #2449: Make StormService deployment mode explicit
## Problem

StormService currently has two deployment modes, but the mode is implicit:

- `spec.replicas > 1` means replica mode.
- `spec.replicas == 1` means pooled mode.

This makes the API hard to understand and easy to misuse because `spec.replicas` is doing two jobs at once:

1. It is the desired number of RoleSets.
2. It is also used as the deployment mode selector.

The ambiguity is especially visible when `spec.replicas == 1`. A user may intend either:

- pooled mode: one RoleSet with independently scalable roles, or
- replica mode with a single service replica.

The autoscaling path already exposes this ambiguity. `PodAutoscaler` uses the annotation `autoscaling.aibrix.ai/storm-service-mode` to decide whether role-level scaling should update `StormService.spec.replicas` or `StormService.spec.template.spec.roles[].replicas`. The code comment also notes that it is hard to know whether `replicas=1` means pool mode or replica mode.

## Why this matters

For users, the current API is too implicit:

- There is no obvious field showing whether a StormService is in pooled mode or replica mode.
- `replicas: 1` is overloaded and cannot distinguish pooled mode from single-replica replica mode.
- Autoscaling requires a separate annotation to recover the missing mode signal.
- Documentation explains the convention, but the resource itself does not encode the intent.

This can lead to surprising scaling behavior, especially for users who start with one replica and later add autoscaling or role-level scaling.

## Current examples

Replica mode:

```yaml
spec:
  replicas: 3
  updateStrategy:
    type: RollingUpdate
  template:
    spec:
      roles:
        - name: prefill
          replicas: 1
        - name: decode
          replicas: 1
```

Pooled mode:

```yaml
spec:
  replicas: 1
  updateStrategy:
    type: InPlaceUpdate
  template:
    spec:
      roles:
        - name: prefill
          replicas: 4
        - name: decode
          replicas: 8
```

Both are valid, but the mode is inferred rather than declared.

## Recommended implementation

Add an explicit deployment mode field to `StormServiceSpec`, for example:

```go
type StormServiceMode string

const (
    StormServiceReplicaMode StormServiceMode = "Replica"
    StormServicePooledMode  StormServiceMode = "Pooled"
)

type StormServiceSpec struct {
    Mode StormServiceMode `json:"mode,omitempty"`
    Replicas *int32 `json:"replicas,omitempty"`
    // ...
}
```

Suggested semantics:

- `mode: Replica`
  - `spec.replicas` means number of RoleSets.
  - `replicas: 1` is valid and means one replica-mode RoleSet.
  - Autoscaling the whole StormService updates `spec.replicas`.

- `mode: Pooled`
  - `spec.replicas` should be `1` or omitted/defaulted to `1`.
  - Each role is scaled through `spec.template.spec.roles[].replicas`.
  - Role-level autoscaling updates the selected role replicas.

## Backward compatibility

To avoid breaking existing users:

1. Keep `mode` optional initially.
2. If `mode` is omitted, preserve the current behavior:
   - `replicas > 1` => replica mode
   - `replicas == 1` or omitted => pooled mode
3. Add defaulting and status reporting so users can see the resolved mode.
4. Update documentation and samples to set `spec.mode` explicitly.
5. Eventually consider warning on omitted mode once the field is broadly adopted.

## Webhook validation

The webhook can make invalid combinations explicit:

- `mode: Pooled` with `spec.replicas > 1` should be rejected or normalized, depending on compatibility preference.
- `mode: Replica` should allow `spec.replicas >= 1`.
- Role-level autoscaling should require `mode: Pooled`, unless the intended behavior is clearly documented for replica mode.

## Autoscaler changes

After `spec.mode` exists, `PodAutoscaler` should read `StormService.spec.mode` instead of relying on:

```yaml
metadata:
  annotations:
    autoscaling.aibrix.ai/storm-service-mode: "pool"
```

The annotation can be kept temporarily as a compatibility fallback, but `spec.mode` should become the source of truth.

## Status and observability

Consider adding the resolved mode to status:

```yaml
status:
  mode: Pooled
```

This would make `kubectl get stormservice -o yaml` self-explanatory and help users/debuggers understand how the controller interpreted the resource.

## Tests

Suggested test coverage:

- API defaulting/resolution when `mode` is omitted.
- Validation for `mode: Pooled` with `replicas > 1`.
- `mode: Replica` with `replicas: 1` remains valid.
- Autoscaler updates `spec.replicas` for replica mode.
- Autoscaler updates `spec.template.spec.roles[].replicas` for pooled mode.
- Existing manifests without `mode` continue to behave as before.

V-3604:
## Pull Request Description
StormService has no field that records its deployment mode. The controller drives behavior from `spec.updateStrategy.type` (`RollingUpdate` for the replica path, `InPlaceUpdate` for the pooled path in `sync.go`), and the autoscaler recovers the mode from the `autoscaling.aibrix.ai/storm-service-mode` annotation on the `PodAutoscaler` (`workload_scale.go`). The design doc defines the canonical rule (`replicas > 1` is replica mode, `replicas == 1` is pooled mode), but nothing on the resource encodes it, so `replicas: 1` is ambiguous and role-level autoscaling needs a separate annotation to recover the intent.

This adds an explicit `Mode` field (`Replica` or `Pooled`) to `StormServiceSpec` and keeps it optional for backward compatibility:

- `StormServiceSpec.ResolvedMode()` returns `spec.mode` when set, otherwise infers it from `spec.replicas` using the documented rule.
- The validating webhook rejects `mode: Pooled` with `spec.replicas > 1` on create and update, since pooled mode runs a single RoleSet and scales roles through `spec.template.spec.roles[].replicas`. Only an explicitly declared `spec.mode` is checked.

`spec.mode` is not defaulted. Recording an inferred mode on an object that never declared one would pin it: a pre-existing `replicas: 1` StormService would get `Pooled` written during an unrelated update and then be blocked from scaling up. Objects that leave `spec.mode` empty keep the inferred behavior on both create and update.

The field is advisory in this change: it does not alter the controller update-path selection, which still keys off `spec.updateStrategy.type`. Moving the autoscaler to read `spec.mode`, with the annotation kept as a temporary fallback, is a focused follow-up so this stays reviewable. The design doc and the `samples/autoscaling` manifests are updated to declare `spec.mode` explicitly.

Generated CRDs and applyconfiguration were regenerated with the pinned `controller-gen` v0.16.1 and `applyconfiguration-gen` v0.31.8.

```
$ go test ./api/orchestration/v1alpha1/... ./pkg/webhook/... -run TestStormService
ok github.com/vllm-project/aibrix/api/orchestration/v1alpha1 0.140s
ok github.com/vllm-project/aibrix/pkg/webhook 0.185s
```

## Related Issues
Resolves: #2449

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
    <li>[x] Documentation updated to reflect changes (if applicable)</li>
</ul>

</details>

### DIFF
--- a/api/orchestration/v1alpha1/stormservice_types.go
+++ b/api/orchestration/v1alpha1/stormservice_types.go
@@ -30,8 +30,8 @@ type StormServiceSpec struct {
 
 	// Mode is the deployment mode of the StormService. When left empty it is resolved
 	// from spec.replicas for backward compatibility: "Replica" when replicas > 1,
-	// otherwise "Pooled" (see ResolvedMode). The mutating webhook defaults this field
-	// so the intent is recorded on the object.
+	// otherwise "Pooled" (see ResolvedMode). The field is not defaulted, so an object
+	// that does not set it keeps the inferred behavior.
 	// +optional
 	// +kubebuilder:validation:Enum={Replica,Pooled}
 	Mode StormServiceMode `json:"mode,omitempty"`
@@ -112,10 +112,6 @@ type RoleSetTemplateSpec struct {
 type StormServiceStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
-
-	// Mode is the resolved deployment mode the controller is operating the StormService in.
-	// +optional
-	Mode StormServiceMode `json:"mode,omitempty"`
 
 	// Total number of non-terminated roleSets targeted by this stormService (their labels match the selector).
 	// +optional
--- a/api/orchestration/v1alpha1/stormservice_types.go
+++ b/api/orchestration/v1alpha1/stormservice_types.go
@@ -28,6 +28,14 @@ type StormServiceSpec struct {
 	// +optional
 	Replicas *int32 `json:"replicas,omitempty"`
 
+	// Mode is the deployment mode of the StormService. When left empty it is resolved
+	// from spec.replicas for backward compatibility: "Replica" when replicas > 1,
+	// otherwise "Pooled" (see ResolvedMode). The mutating webhook defaults this field
+	// so the intent is recorded on the object.
+	// +optional
+	// +kubebuilder:validation:Enum={Replica,Pooled}
+	Mode StormServiceMode `json:"mode,omitempty"`
+
 	// Label selector for roleSets. Existing ReplicaSets whose roleSets are
 	// selected by this will be the ones affected by this stormService.
 	// It must match the roleSet template's labels.
@@ -66,6 +74,32 @@ const (
 	DefaultStormServiceUniqueLabelKey string = "stormservice-template-hash"
 )
 
+// StormServiceMode is the deployment mode of a StormService.
+// +enum
+type StormServiceMode string
+
+const (
+	// StormServiceReplicaMode treats each RoleSet as an independent replica of the
+	// service, so spec.replicas is the desired number of RoleSets.
+	StormServiceReplicaMode StormServiceMode = "Replica"
+	// StormServicePooledMode runs a single RoleSet whose roles scale independently
+	// through spec.template.spec.roles[].replicas.
+	StormServicePooledMode StormServiceMode = "Pooled"
+)
+
+// ResolvedMode returns the effective deployment mode. An explicit spec.mode takes
+// precedence; otherwise the mode is inferred from spec.replicas for backward
+// compatibility (replicas > 1 is Replica mode, replicas <= 1 or unset is Pooled mode).
+func (s *StormServiceSpec) ResolvedMode() StormServiceMode {
+	if s.Mode != "" {
+		return s.Mode
+	}
+	if s.Replicas != nil && *s.Replicas > 1 {
+		return StormServiceReplicaMode
+	}
+	return StormServicePooledMode
+}
+
 type RoleSetTemplateSpec struct {
 	// +optional
 	metav1.ObjectMeta `json:"metadata,omitempty" protobuf:"bytes,1,opt,name=metadata"`
@@ -78,6 +112,10 @@ type RoleSetTemplateSpec struct {
 type StormServiceStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
+
+	// Mode is the resolved deployment mode the controller is operating the StormService in.
+	// +optional
+	Mode StormServiceMode `json:"mode,omitempty"`
 
 	// Total number of non-terminated roleSets targeted by this stormService (their labels match the selector).
 	// +optional
--- a/pkg/webhook/stormservice_webhook.go
+++ b/pkg/webhook/stormservice_webhook.go
@@ -53,9 +53,6 @@ func (r *StormServiceCustomDefaulter) Default(_ context.Context, obj runtime.Obj
 		return fmt.Errorf("expected a StormService object but got %T", obj)
 	}
 
-	// Record the resolved deployment mode so the intent is explicit on the object.
-	stormService.Spec.Mode = stormService.Spec.ResolvedMode()
-
 	// Only proceed if the sidecar injection annotation is present
 	if _, exists := stormService.GetAnnotations()[SidecarInjectionAnnotation]; !exists {
 		return nil
@@ -212,23 +209,10 @@ func (r *StormServiceCustomDefaulter) ValidateCreate(ctx context.Context, obj ru
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
-func (r *StormServiceCustomDefaulter) ValidateUpdate(_ context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
+func (r *StormServiceCustomDefaulter) ValidateUpdate(_ context.Context, _, newObj runtime.Object) (admission.Warnings, error) {
 	stormService, ok := newObj.(*orchestrationv1alpha1.StormService)
 	if !ok {
 		return nil, fmt.Errorf("expected a StormService object but got %T", newObj)
-	}
-	oldStormService, ok := oldObj.(*orchestrationv1alpha1.StormService)
-	if !ok {
-		return nil, fmt.Errorf("expected a StormService object but got %T", oldObj)
-	}
-	// The deployment mode selects the RoleSet topology the controller builds, so
-	// switching it on a live StormService would force a disruptive teardown and
-	// rebuild. Treat the resolved mode as immutable and require recreation instead.
-	// Comparing ResolvedMode rather than spec.Mode keeps objects created before the
-	// mode field existed (stored spec.mode is empty) from being rejected on updates.
-	if oldStormService.Spec.ResolvedMode() != stormService.Spec.ResolvedMode() {
-		return nil, fmt.Errorf("StormService deployment mode is immutable; cannot change from %s to %s",
-			oldStormService.Spec.ResolvedMode(), stormService.Spec.ResolvedMode())
 	}
 	if err := validateStormServiceMode(stormService); err != nil {
 		return nil, err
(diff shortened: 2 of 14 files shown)


pkg/webhook/stormservice_webhook.go
```diff
@@ -206,10 +213,28 @@ func (r *StormServiceCustomDefaulter) ValidateCreate(ctx context.Context, obj ru
...
-	// TODO(user): fill in your validation logic upon object update.
+	stormService, ok := newObj.(*orchestrationv1alpha1.StormService)
+	if !ok {
+		return nil, fmt.Errorf("expected a StormService object but got %T", newObj)
+	}
+	if err := validateStormServiceMode(stormService); err != nil {
+		return nil, err
+	}
 	return nil, nil
 }
```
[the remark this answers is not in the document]
V-3604: Done. ValidateUpdate now rejects a change to the resolved deployment mode and points the user to recreate the object.

Two implementation notes worth surfacing:

- The comparison is on ResolvedMode() for both old and new, not spec.mode. An object created before this field existed has an empty stored spec.mode; comparing raw spec.mode would falsely flag its first update as a mode change. ResolvedMode() normalizes both sides so those objects update cleanly.
- This does enforce immutability against the backward-compat inference path: a legacy object with no explicit spec.mode that scales spec.replicas from 1 to 3 now resolves Pooled -> Replica and is rejected. That is the intended safety behavior (switching topology on a live service is destructive), and the migration is to set spec.mode explicitly or recreate. Flagging it in case a maintainer wants the implicit flip to stay permitted.

pkg/webhook/stormservice_webhook_test.go
```diff
@@ -0,0 +1,92 @@
...
+			}
+			_, err := validator.ValidateCreate(context.Background(), ss)
+			if tc.expectError {
+				require.Error(t, err)
+			} else {
+				require.NoError(t, err)
+			}
+		})
+	}
+}
```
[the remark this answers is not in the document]
V-3604: Added TestStormServiceValidateUpdate_ModeImmutability covering unchanged mode, an explicit Pooled -> Replica change, and a replica-driven implicit change. All three pass.

api/orchestration/v1alpha1/stormservice_types.go
```diff
@@ -79,6 +113,10 @@ type StormServiceStatus struct {
 	// +optional
 	ObservedGeneration int64 `json:"observedGeneration,omitempty"`
 
+	// Mode is the resolved deployment mode the controller is operating the StormService in.
+	// +optional
+	Mode StormServiceMode `json:"mode,omitempty"`
```
googs1025: I think we should drop status.mode from this change.

  Once spec.mode exists, status.mode mostly duplicates spec.mode/ResolvedMode and creates an extra consistency contract. Since the controller still selects the execution path from
  spec.updateStrategy.type in this PR, reporting status.mode as "the mode the controller is operating in" can be misleading. Keeping only spec.mode makes the API source of truth
  clearer and avoids status/spec drift.

  For observability, kubectl get -o yaml will already show spec.mode for newly defaulted objects. Legacy objects can still be interpreted through ResolvedMode() internally until they
  are migrated.

pkg/webhook/stormservice_webhook.go
```diff
@@ -206,10 +213,41 @@ func (r *StormServiceCustomDefaulter) ValidateCreate(ctx context.Context, obj ru
...
+	oldStormService, ok := oldObj.(*orchestrationv1alpha1.StormService)
+	if !ok {
+		return nil, fmt.Errorf("expected a StormService object but got %T", oldObj)
+	}
+	// The deployment mode selects the RoleSet topology the controller builds, so
+	// switching it on a live StormService would force a disruptive teardown and
+	// rebuild. Treat the resolved mode as immutable and require recreation instead.
+	// Comparing ResolvedMode rather than spec.Mode keeps objects created before the
+	// mode field existed (stored spec.mode is empty) from being rejected on updates.
+	if oldStormService.Spec.ResolvedMode() != stormService.Spec.ResolvedMode() {
```
googs1025: This breaks compatibility for existing StormService objects.

  For a pre-existing object without spec.mode, scaling from replicas: 1 to replicas: 3 used to be a valid operation. With this PR, ValidateUpdate rejects it as a resolved mode change.
  There is also a second compatibility trap: because the mutating webhook runs on update, an old replicas: 1 object can get spec.mode defaulted to Pooled during an unrelated update, after which a later scale-up to replicas > 1 is rejected by the Pooled/replicas validation.

api/orchestration/v1alpha1/stormservice_types.go
V-3604: Dropped status.mode. Agreed it duplicates spec.mode without the controller acting on it, and ResolvedMode() covers legacy objects internally.

pkg/webhook/stormservice_webhook.go
V-3604: Reverted the immutability check. Also dropped the spec.mode defaulting, since the second trap comes from that on its own: once Default() stamps Pooled on a replicas: 1 object, the Pooled/replicas rule blocks the later scale-up whether or not the immutability check is there. spec.mode is now user-declared only, so objects that don't set it behave exactly as before on create and update.

pkg/webhook/stormservice_webhook.go
```diff
@@ -205,11 +209,30 @@ func (r *StormServiceCustomDefaulter) ValidateCreate(ctx context.Context, obj ru
...
-func (r *StormServiceCustomDefaulter) ValidateUpdate(_ context.Context, oldObj, newObj runtime.Object) (admission.Warnings, error) {
-	// TODO(user): fill in your validation logic upon object update.
+func (r *StormServiceCustomDefaulter) ValidateUpdate(_ context.Context, _, newObj runtime.Object) (admission.Warnings, error) {
+	stormService, ok := newObj.(*orchestrationv1alpha1.StormService)
+	if !ok {
+		return nil, fmt.Errorf("expected a StormService object but got %T", newObj)
+	}
+	if err := validateStormServiceMode(stormService); err != nil {
+		return nil, err
+	}
```
googs1025, line 219: I found one API-boundary question that may need attention here. @jeffwan could you also take a look and confirm whether this is something we need to cover?

  `validateStormServiceMode()` covers normal StormService create/update, but does it also cover updates through the `/scale` subresource?

  StormService exposes `/scale` with `specpath=.spec.replicas`, so clients like `kubectl scale`, HPA-style controllers, or external autoscalers can update replicas via `stormservices/
  scale`. If that admission request carries an `autoscaling/v1.Scale` object instead of a full `StormService`, this helper may not run unless the webhook is explicitly registered for and handles the scale subresource.

  This was not an invariant before `spec.mode` existed, because `replicas > 1` simply inferred Replica mode. With this PR, explicit `mode: Pooled` introduces a new API invariant: `spec.replicas` must stay <= 1. Could we either verify that the current webhook catches `/scale` updates, or add coverage for that path so explicit `mode: Pooled` cannot be scaled above 1 through the scale API?

pkg/webhook/stormservice_webhook.go
```diff
@@ -205,11 +209,30 @@ func (r *StormServiceCustomDefaulter) ValidateCreate(ctx context.Context, obj ru
 }
 
 // ValidateUpdate implements webhook.Validator so a webhook will be registered for the type
```
googs1025, line 211: please also add case in `test/integration/webhook`

googs1025 (commented):
only left two comments

pkg/webhook/stormservice_webhook.go
Jeffwan: Reasonable concern. I kind of forget the webhook behavior but I doubt we checked the scale resource earlier. this would be some improvement we can do later.

Jeffwan:
@googs1025 do you have changes requested? Assign to you to review and approve.

googs1025:
@V-3604 Could you please take another look at the latest comments on this PR and confirm whether anything else needs to be updated?

googs1025:
@V-3604 are you still working on this? 😄

pkg/webhook/stormservice_webhook.go
V-3604: Added in `test/integration/webhook/stormservice_webhook_test.go`: explicit `Pooled` with `replicas: 2` rejected, explicit `Pooled` with one replica accepted, undeclared mode with `replicas: 3` accepted, and a scale-up update on an explicit `Pooled` object rejected.

pkg/webhook/stormservice_webhook.go
V-3604: Agreed, leaving it out of this PR. Nothing here changes the scale path, so `stormservices/scale` behaves as it did before `spec.mode`. I can open a separate issue for CEL enforcement on `StormServiceSpec` if needed.

V-3604:
@googs1025 I'm sorry abut the delay. Thanks for keeping the branch rebased.

googs1025 (commented):
This PR introduces the explicit mode field and preserves backward compatibility. Controller/autoscaler adoption and scale-subresource enforcement are follow-ups.

googs1025 (approved):

Issue #2440, opened 2026-07-13
[Bug Report] sync.Pool use-after-free causes cross-request stats corruption and request count miscounting


allswell00:
# [Bug Report] sync.Pool use-after-free causes cross-request stats corruption and request count miscounting

## TL;DR

The `RoutingContext` object is returned to `sync.Pool` via `Delete()` in deferred functions inside `HandleResponseHeaders`/`HandleResponseBody`, while `Process()` continues to hold and use the same pointer. When another goroutine picks up the same object from the pool, the original request's subsequent `DoneRequestCount` call operates on the new request's `statsUpdated` field, causing the new request's `donePodStats` to be skipped by CAS and `pendingRequests` to be double-counted.

## Background: How I discovered this

I started by fixing an issue where `runningRequests` counts would become inaccurate — sometimes going negative — because Pod status updates (non-restart) would reset the counters, and then a late `-1` from a historical request would push them below zero. I implemented a `sync.Map`-based approach to precisely track the relationship between Pods and request IDs (see the detailed fix guide).

**After implementing this precise tracking, I discovered a much deeper problem**: the tracking data showed that one request's `-1` operation was being executed twice, while another request's `-1` was being completely skipped. This led me to the root cause documented below.

### The Mechanism

1. **Request A** receives a non-200 response → `HandleResponseHeaders` deferred function executes:
   - `DoneRequestCount()` → CAS `statsUpdated` 1→2 succeeds ✅
   - `routerCtx.Delete()` → **object returned to `sync.Pool`** ⚠️
   - But `Process()`'s main loop is still running and **still holds `st.routerCtx` pointing to the same object**

2. **Request B** (different goroutine) calls `requestPool.Get()` → gets the **exact same object** that Request A just returned

3. **Request A**'s `Process()` loop continues → `srv.Recv()` returns an error (e.g., context canceled) → calls `DoneRequestCount(st.routerCtx, ...)` → **operates on Request B's `statsUpdated`!** → CAS `statsUpdated` 1→2 succeeds (but this is Request B's state!)

4. **Request B** completes normally → `HandleResponseBody` defer calls `DoneRequestTrace()` → CAS `statsUpdated` 1→2 **fails** (already set to 2 by Request A!) → `donePodStats` is skipped → **Request B's count is never decremented**

### Concrete Evidence (from tracing logs)

| Time | Event | ctx_ptr | request_id | statsUpdated |
|------|-------|---------|------------|-------------|
[...]

PR #2463, opened 2026-07-17
fix(cache): decrement pendingRequests exactly once per request (#2440)

Anai-Guo:
## What

`pendingRequests` (the per-model in-flight counter in `metaModels`) is
incremented once in `AddRequestCount`, guarded by `CanAddTrace()` — a CAS on
the `traceAdded` flag, so it fires exactly once per request:

```go
if ctx == nil || ctx.CanAddTrace() {
    ...
    atomic.AddInt32(&meta.pendingRequests, 1)
}
```

The matching decrements in `DoneRequestCount` and `DoneRequestTrace` were **not**
guarded:

```go
meta, ok := c.metaModels.Load(modelName)
if ok {
    atomic.AddInt32(&meta.pendingRequests, -1)
}
```

A single request routes through several completion paths — response-headers
defer, response-body defer, and the receive-error exits in `Process()` — and
each one calls into `DoneRequestCount`/`DoneRequestTrace`. Because the decrement
had no once-only guard, one request could decrement `pendingRequests` multiple
times, drifting the counter below its true value.

## Fix

Add a `CanDoneTrace()` CAS (`traceAdded`: `statusAdded → statusDone`) and guard
both decrements with it, so the decrement fires exactly once — mirroring the
increment.

Pairing on `traceAdded` (via `CanAddTrace`/`CanDoneTrace`) rather than on
`statsUpdated` (`CanDoneStats`) is deliberate: the increment is gated by
`CanAddTrace`, so only a `traceAdded`-based CAS keeps increment and decrement
balanced for a given request. `traceAdded` is already reset to `statusInitial`
when a `RoutingContext` is recycled from the pool, so the next request starts a
fresh `AddTrace → DoneTrace` cycle.

The `ctx == nil` short-circuit mirrors the increment side's `ctx == nil ||
ctx.CanAddTrace()`, preserving the existing behaviour for contexts that are
cancelled before routing.

## Scope

This addresses **defect B1** described in #2440 (the unguarded `pendingRequests`
decrement). The other defects in that report (the `sync.Pool` use-after-free and
the double `pool.Put`) are broader and left for separate discussion.

## Test

`go build ./...` and existing cache unit tests. No behaviour change for the
normal single-completion path; only redundant decrements from additional
completion paths are now suppressed.

### DIFF
--- a/pkg/cache/cache_impl.go
+++ b/pkg/cache/cache_impl.go
@@ -237,7 +237,7 @@ func (c *Store) DoneRequestCount(ctx *types.RoutingContext, requestID string, mo
 	}
 
 	meta, ok := c.metaModels.Load(modelName)
-	if ok {
+	if ok && (ctx == nil || ctx.CanDoneTrace()) {
 		atomic.AddInt32(&meta.pendingRequests, -1)
 	}
 
@@ -267,7 +267,7 @@ func (c *Store) DoneRequestTrace(ctx *types.RoutingContext, requestID string, mo
 	}
 
 	meta, ok := c.metaModels.Load(modelName)
-	if ok {
+	if ok && (ctx == nil || ctx.CanDoneTrace()) {
 		atomic.AddInt32(&meta.pendingRequests, -1)
 	}
 
--- a/pkg/cache/cache_test.go
+++ b/pkg/cache/cache_test.go
@@ -694,6 +694,17 @@ var _ = Describe("Cache", func() {
 
 		// RunningReq should be 0
 		Expect(runningReqAfter.GetSimpleValue()).To(Equal(0.0))
+
+		// Directly assert on the model-level pendingRequests counter. It is
+		// incremented exactly once (AddRequestCount is guarded by CanAddTrace),
+		// so the two DoneRequestCount calls above must decrement it exactly
+		// once in total. This is the counter the fix protects: without the
+		// CanDoneTrace() guard the second DoneRequestCount over-decrements it
+		// to -1, whereas RealtimeNumRequestsRunning (asserted above) is derived
+		// from pod stats and stays 0 even with the bug present.
+		meta, ok := cache.metaModels.Load(modelName)
+		Expect(ok).To(BeTrue())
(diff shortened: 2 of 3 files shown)


Anai-Guo:
Good catch — `RoutingContext` is pooled via `sync.Pool`, so `traceAdded` needed to be reset in `reset()` alongside `statsUpdated`; otherwise a recycled context stuck at `statusDone` would skip the decrement on its next request. Fixed in 41c2741.

varungup90:
Please sign the PR and check integration tests.

Test coverage — gap
No new/updated test asserts on meta.pendingRequests for the actual buggy scenario (non-nil ctx, Done* called from multiple completion paths). The closest existing test, "should multiple calls to AddRequestCount and DoneRequestCount not affect request counting" (pkg/cache/cache_test.go:660), already calls DoneRequestCount twice with the same non-nil ctx — exactly the regression scenario — but only asserts on the RealtimeNumRequestsRunning metric (which comes from CanDoneStats/pod stats), not on pendingRequests directly. That test would pass even without this fix. Suggest adding an assertion like Expect(meta.pendingRequests).To(Equal(int32(0))) there, or a small dedicated test exercising DoneRequestCount + DoneRequestTrace both called for the same non-nil-ctx request to confirm the counter only decrements once.

Anai-Guo:
Thanks @varungup90 — addressed both points:

- **DCO**: all three commits are now signed off (the middle commit was missing the trailer). DCO check is green.
- **Test coverage**: added a direct assertion on `meta.pendingRequests` to the existing "multiple calls to AddRequestCount and DoneRequestCount" test (`pkg/cache/cache_test.go`). As you noted, that test only asserted on `RealtimeNumRequestsRunning`, which is derived from pod stats (`CanDoneStats`) and stays `0` even with the bug. `pendingRequests` is incremented exactly once (`AddRequestCount` is guarded by `CanAddTrace`), so the two `DoneRequestCount(ctx, ...)` calls with the same non-nil ctx must net to a single decrement. Without the `CanDoneTrace()` guard the counter goes to `-1`, so this assertion now fails on the pre-fix code and passes with the fix.

varungup90 (approved):

PR #2467, opened 2026-07-20
docs: update AI Gateway integration samples

googs1025:
## Summary
Update the AI Gateway integration samples for the current Envoy AI Gateway / GIE flow.

This includes v1beta1 `AIGatewayRoute`, GIE v1.5.0, Envoy Gateway v1.8.2, missing InferencePool RBAC, and refreshed llm-d-router disaggregation docs.

## Testing
- `git diff --check -- samples/ai-gateway-integration`
- `kubectl apply --dry-run=client -f samples/ai-gateway-integration -R`
- Verified llama2, mistral, and qwen2 routes on arm64 minikube.

### DIFF
--- a/samples/ai-gateway-integration/README.md
+++ b/samples/ai-gateway-integration/README.md
@@ -8,14 +8,15 @@ This guide walks you through deploying a multi-model AI inference gateway using 
 samples/ai-gateway-integration
 ├── gateway.yaml                  # GatewayClass + Gateway
 ├── aigatewayroute.yaml           # Multi-model routing rules (llama2-7b, mistral-7b)
+├── envoy-gateway-inferencepool-rbac.yaml # Envoy Gateway access to InferencePool
 ├── llama-7b-inferencepool.yaml   # InferencePool + EPP for Llama2-7B
 ├── mistral-7b-inferencepool.yaml # InferencePool + EPP for Mistral-7B
-└── llama-7b.yaml                 # Mock model llama-7b deployments
+├── llama-7b.yaml                 # Mock model llama-7b deployments
 └── mistral-7b.yaml               # Mock model mistral-7b deployments
 ```
 
 ### Prerequisites
-- Kubernetes cluster (v1.24+)
+- Kubernetes cluster (v1.32+ recommended for Envoy AI Gateway v1.0.0)
 - kubectl configured
 - helm v3.8+
 - Internet access to pull images from docker.io and GitHub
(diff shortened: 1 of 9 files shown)


varungup90:
### 🔍 Pull Request Review Notes

* **Silent removal of liveness/readiness probes**
`llama-7b-inferencepool.yaml` drops the gRPC liveness/readiness probes on port `9003` (lines ~95–106 in the old file) with no explanation in the PR body. While `mistral-7b-inferencepool.yaml` never had these probes—meaning this brings the two files in line with each other—it is worth confirming this was intentional (e.g., because GIE v1.5.0's EPP doesn't expose the same gRPC health service, causing false restarts) rather than an incidental deletion.
* **Floating image tag on the disaggregation EPP**
`llm-d-inference-scheduler-epp.yaml` now points at `ghcr.io/llm-d/llm-d-router-endpoint-picker-dev:main` with `imagePullPolicy: IfNotPresent`. A `-dev` image on a mutable `:main` tag combined with `IfNotPresent` means a node that already cached an older main build will silently keep running it. While this isn't a regression (the previous tag was `:latest`), reproducibility for the sample is weak. It would be worth pinning to a specific digest or release tag if `llm-d-router` has one available.
* **RBAC file duplication**
`envoy-gateway-inferencepool-rbac.yaml` is duplicated near-verbatim between `samples/ai-gateway-integration/` and `samples/ai-gateway-integration/disaggregation/`, with a minor style inconsistency between them (inline `apiGroups: ["..."]` vs. multi-line list form). This is acceptable for copy-paste-runnable samples, but flagging in case a shared location was originally intended.
* **`inferencemodelrewrites` RBAC resource**
This resource was added to the EPP `ClusterRole` in three files based on the new `InferenceModelRewrite` CRD from GIE v1.5.0. It is worth double-checking that `kubectl apply --dry-run=client` (mentioned in the PR's testing notes) was actually validated against a live v1.5.0 API server rather than just verifying YAML syntax, as the exact plural resource name should be verified against the upstream CRD manifest.
* **Testing claims & behavioral validation**
The PR notes state that routes were verified on arm64 minikube for llama2/mistral/qwen2, which is excellent. However, note that `git diff --check` and `kubectl apply --dry-run=client` only catch syntax issues rather than RBAC or behavioral correctness (e.g., whether the new `label-selector-filter`/`prefix-based-pd-decider` plugin config in the EPP `ConfigMap` actually produces the same prefill/decode routing behavior as before). Since manual verification is mentioned, this is likely fine, but worth noting that automated checks alone wouldn't catch a functional regression here.

varungup90 (approved):
Overall LGTM

PR #2469, opened 2026-07-20
[Feat] Add Dynamo provider support to brixbench

whalepark:
This PR adds Dynamo provider support to `brixbench` and enables AIBrix/Dynamo comparison scenarios.

Changes include:
- Dynamo provider validation, release resolution, deployment, readiness, cleanup, and artifact capture.
- Dynamo VKE smoke fixtures and a combined AIBrix/Dynamo comparison scenario.
- Bounded Dynamo log artifact capture.
- Figure generation fallback to system `python3`/`python`, with warning-only skips for missing plotting dependencies.


brixbench/internal/deployers/dynamo_release.go
```diff
@@ -0,0 +1,184 @@
...
+
+	if pathExists(repoPath) {
+		if err := syncDynamoReleaseCheckout(ctx, s.runner, repoPath, s.repoURL, version); err != nil {
+			return nil, err
+		}
+		if !dirExists(chartPath) {
+			return nil, fmt.Errorf("Dynamo release checkout %s exists but chart path %s was not found", repoPath, chartPath)
+		}
+		return release, nil
+	}
```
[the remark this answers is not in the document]
whalepark: Addressed. PrepareRelease now validates an existing checkout before reuse, syncs only when it looks like a git checkout, and removes/re-clones the cached checkout if sync fails and the context was not canceled. Added a regression test for the failed-sync re-clone path.

brixbench/internal/deployers/dynamo_artifacts.go
```diff
@@ -0,0 +1,156 @@
...
+		content, err := os.ReadFile(d.engineManifest)
+		if err != nil {
+			content = []byte(fmt.Sprintf("capture failed: failed to read Dynamo engine manifest %s: %v\n", d.engineManifest, err))
+		}
+		if err := writeDynamoArtifactFile(artifactDir, "engine-manifest.yaml", string(content)); err != nil {
+			return err
+		}
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
whalepark: Addressed. CaptureArtifacts now continues writing the remaining artifacts when an individual artifact write fails, collects write errors, and returns them with errors.Join at the end.

brixbench/internal/deployers/dynamo.go
```diff
@@ -18,50 +18,1185 @@ package deployers
...
 	return nil
 }
 
 func (d *DynamoDeployer) DeployControlPlane(ctx context.Context) error {
-	return fmt.Errorf("Dynamo deployer is not implemented")
+	release, err := d.releaseSource.PrepareRelease(ctx, d.projectRoot, d.version)
+	if err != nil {
+		return fmt.Errorf("failed to prepare Dynamo release %s: %w", d.version, err)
+	}
+	d.release = release
```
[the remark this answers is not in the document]
whalepark: Addressed. DeployControlPlane now checks for a nil release after PrepareRelease and returns an explicit error before accessing release.ChartPath.

brixbench/internal/deployers/dynamo.go
```diff
@@ -18,50 +18,1185 @@ package deployers
...
+		}
+		args = append(args, "-f", d.engineManifest, "--ignore-not-found", "--wait=false")
+		addCriticalErr(d.runDynamoCleanupCommand(ctx, "delete-dynamo-graph-deployment", "kubectl", args...))
+		waitArgs := []string{"wait", "--for=delete"}
+		if strings.TrimSpace(namespace) != "" {
+			waitArgs = append(waitArgs, "-n", namespace)
+		}
+		waitArgs = append(waitArgs, "-f", d.engineManifest, "--timeout=2m")
+		addCriticalErrUnlessNotFound(d.runDynamoCleanupCommand(ctx, "wait-delete-dynamo-graph-deployment", "kubectl", waitArgs...))
+	}
```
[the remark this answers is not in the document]
whalepark: Addressed. Teardown now checks that the local engine manifest exists before running kubectl -f patch/delete/wait commands, while still continuing component, Helm, PVC, and namespace cleanup. Added a regression test for stale cleanup with a missing manifest.

brixbench/internal/deployers/dynamo.go
```diff
@@ -18,50 +18,1185 @@ package deployers
...
+
+func (d *DynamoDeployer) runDynamoHelmCommandWithRetry(ctx context.Context, stage string, attempts int, args ...string) error {
+	if attempts <= 1 {
+		return d.runDynamoHelmCommand(ctx, stage, args...)
+	}
+
+	var lastErr error
+	for attempt := 1; attempt <= attempts; attempt++ {
+		lastErr = d.runDynamoHelmCommand(ctx, stage, args...)
+		if lastErr == nil {
```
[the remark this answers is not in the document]
whalepark: Addressed. runDynamoHelmCommandWithRetry now checks ctx.Err() before executing each attempt, including the single-attempt path.

brixbench/internal/deployers/dynamo.go
```diff
@@ -18,50 +18,1185 @@ package deployers
...
+
+func (d *DynamoDeployer) waitForDynamoComponentPods(ctx context.Context, component string, expectedReplicas int, timeout time.Duration, interval time.Duration) error {
+	if expectedReplicas < 1 {
+		expectedReplicas = 1
+	}
+	deadline := time.Now().Add(timeout)
+	var lastErr error
+	for {
+		output, err := d.captureDynamoCommand(ctx, "get-dynamo-"+component+"-pods", "kubectl", "get", "pod", "-n", d.effectiveNS, "-l", dynamoFrontendComponentLabel+"="+component, "-o", "json")
+		if err == nil {
```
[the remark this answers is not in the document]
whalepark: Addressed. waitForDynamoComponentPods now checks ctx.Err() at the beginning of each polling iteration before running kubectl.

brixbench/benchmark/testdata/deployments/dynamo/qwen3-32b-round-robin-4p8d-vke.yaml
```diff
@@ -0,0 +1,339 @@
...
+              export TP=4
+              python3 -m dynamo.vllm \
+                --model /models/$MODEL \
+                --served-model-name $SERVED_MODEL_NAME \
+                --max-model-len $MAX_MODEL_LEN \
+                --gpu-memory-utilization $GPU_MEM_UTIL \
+                --tensor-parallel-size $TP \
+                --disaggregation-mode decode \
+                --kv-transfer-config '{"kv_connector":"NixlConnector","kv_role":"kv_both"}' \
+                --enforce-eager
```
JinKim48: Both VKE Dynamo fixtures pass `--enforce-eager`. For smoke under tight GPU memory that can be convenient, but it changes the execution path and biases latency/throughput vs a normal eager+CUDA graph run.
Since this option was not used in the existing AIBrix test scenario, it should also be removed from the Dynamo test as well.

brixbench/benchmark/testdata/deployments/dynamo/qwen3-32b-round-robin-4p8d-vke.yaml
```diff
@@ -0,0 +1,339 @@
...
+              python3 -m dynamo.vllm \
+                --model /models/$MODEL \
+                --served-model-name $SERVED_MODEL_NAME \
+                --max-model-len $MAX_MODEL_LEN \
+                --gpu-memory-utilization $GPU_MEM_UTIL \
+                --tensor-parallel-size $TP \
+                --disaggregation-mode prefill \
+                --kv-transfer-config '{"kv_connector":"NixlConnector","kv_role":"kv_both"}' \
+                --kv-events-config '{"publisher":"zmq","topic":"kv-events","endpoint":"tcp://*:20080","enable_kv_cache_events":true}' \
+                --enforce-eager
```
JinKim48: It is the same as the comment made on line 199.

brixbench/internal/deployers/dynamo.go
```diff
@@ -18,50 +18,1198 @@ package deployers
...
+		if strings.TrimSpace(namespace) != "" {
+			waitArgs = append(waitArgs, "-n", namespace)
+		}
+		waitArgs = append(waitArgs, "-f", d.engineManifest, "--timeout=2m")
+		addCriticalErrUnlessNotFound(d.runDynamoCleanupCommand(ctx, "wait-delete-dynamo-graph-deployment", "kubectl", waitArgs...))
+	}
+	if strings.TrimSpace(namespace) != "" {
+		addCriticalErr(d.deleteDynamoComponentDeployments(ctx, namespace, componentDeployments))
+		addCriticalErr(d.runDynamoHelmCleanupCommand(ctx, "uninstall-dynamo-platform", "uninstall", dynamoPlatformHelmReleaseName, "-n", namespace, "--ignore-not-found", "--wait", "--timeout", "5m"))
+		_ = d.runDynamoCleanupCommand(ctx, "delete-dynamo-pvcs", "kubectl", "delete", "pvc", "--all", "-n", namespace, "--ignore-not-found")
```
JinKim48, line 293: The VKE Dynamo engine manifests create a cluster-scoped PV (`models-pv-brixbench-dynamo`, Retain) plus a namespaced PVC. Teardown deletes PVCs and the namespace, but does not clear the PV `/spec/claimRef`.

After a failed or partial run this can leave the PV Bound/Released in a state where the next apply of the same fixture fails or hangs on PVC binding. Reference cleanup scripts patch the claimRef away for exactly this reason.

This looks Dynamo-fixture-specific today (AIBrix smoke mounts hostPath directly and does not use this PV/PVC pattern). Suggestion: best-effort

  kubectl patch pv models-pv-brixbench-dynamo --type=json \
    -p '[{"op":"remove","path":"/spec/claimRef"}]'

in Dynamo teardown when that PV exists, or document the manual cleanup step next to the VKE scenario.

brixbench/internal/deployers/dynamo.go
```diff
@@ -18,50 +18,1198 @@ package deployers
...
+		`export\s+MODEL=["']?([A-Za-z0-9._-]+)["']?`,
+	} {
+		if value := firstRegexGroup(content, pattern); value != "" && !strings.HasPrefix(value, "$") {
+			return value
+		}
+	}
+	return ""
+}
+
+func (d *DynamoDeployer) waitForDynamoModelReady(ctx context.Context) error {
```
JinKim48, line 559: Missing inference readiness probe + silent skip risk

1) At HEAD, `waitForDynamoModelReady` only waits for `/v1/models` registration. A short inference smoke (as in the reference deploy flow) catches cases where pods are Ready and the model is listed but the stack still cannot serve. If you restore a probe, please use `/v1/chat/completions` so it matches the VKE smoke / comparison benchmarks (`endpoint: /v1/chat/completions`, `openai-chat`), not `/v1/completions`.

2) If `inferDynamoModelNameFromManifest` returns empty, the function returns nil with no warning, so even the `/v1/models` check is skipped. A single warning would make “Ready” less easy to over-read as serving-ready.

brixbench/benchmark/testdata/deployments/dynamo/qwen3-32b-round-robin-4p8d-vke.yaml
whalepark: Removed `--enforce-eager`, so the comparison path uses the normal CUDA graph path while smoke coverage stays intact.

brixbench/benchmark/testdata/deployments/dynamo/qwen3-32b-round-robin-4p8d-vke.yaml
whalepark: Removed --enforce-eager, so the comparison path uses the normal CUDA graph path while smoke coverage stays intact.

brixbench/internal/deployers/dynamo.go
whalepark: Addressed. Dynamo teardown now deletes PVCs and then best-effort clears the retained models PV `claimRef`.

The cleanup is non-fatal so teardown still succeeds if the PV is already absent or clean.

brixbench/internal/deployers/dynamo.go
whalepark: Addressed. Dynamo readiness now checks `/v1/models` first, then probes `/v1/chat/completions` with a minimal request.

This catches cases where the model is registered but inference is not ready yet.

Addressed. If Dynamo model-name inference fails, the readiness probe skip now emits a warning with the manifest path.

This makes the skipped readiness check visible in benchmark logs.

PR #2476, opened 2026-07-22
[Bug]preserve byte-level field order in SGLang PD prefill/decode bodies

### LINKED ISSUE #2475: [Bug] SGLang PD nondeterministically re-serializes tool schemas, causing prefill/decode prompt-token mismatch
### 🐛 Describe the bug

AIBrix SGLang PD routing re-serializes the complete OpenAI request through
`map[string]any` twice:

1. `prefill.PreparePayload` unmarshals `routingCtx.ReqBody`.
2. `SGLangHandler.AugmentPrefillRequest` marshals the map to create the decode body.
3. `PreparePayload` changes the prefill generation fields and marshals the same map again.

Affected source:

- https://github.com/vllm-project/aibrix/blob/e397f9866e19866abd943b93244f3dd03b5873cb/pkg/plugins/gateway/algorithms/pd/prefill/payload.go#L36-L62
- https://github.com/vllm-project/aibrix/blob/e397f9866e19866abd943b93244f3dd03b5873cb/pkg/plugins/gateway/algorithms/pd/engine/sglang.go#L50-L65

AIBrix uses `sonic.Marshal`, whose default configuration does not sort map keys.
Nested objects such as OpenAI tool schemas can therefore be emitted in different
orders between the prefill and decode bodies, or between repeated identical
requests.

SGLang serializes the tool definition as part of the chat template. A changed
field order produces different `prompt_token_ids`, even though the JSON remains
semantically equivalent. In our deployment this caused identical requests to hit
only partial page-aligned prefixes (for example 192, 256, or 320 cached tokens)
instead of the full reusable prefix.

The combined-pod path is not affected because it does not issue a separate
prefill request.

A possible fix is to mutate only the required top-level bootstrap and prefill
control fields on the original JSON bytes, for example with `sjson`, rather than
unmarshalling and re-marshalling the complete request.

### Steps to Reproduce

1. Deploy AIBrix `v0.7.0` or `main@e397f9866e19866abd943b93244f3dd03b5873cb`
   with separate SGLang prefill and decode pods.
2. Send requests through the gateway with `routing-strategy: pd`.
3. Use an identical `/v1/chat/completions` body containing a nested tool schema,
   for example:

```json
{
  "model": "<served-model>",
  "messages": [
    {
      "role": "user",
      "content": "Call the lookup tool"
    }
  ],
  "tools": [
    {
      "type": "function",
      "function": {
        "name": "lookup",
        "description": "Lookup a value",
        "parameters": {
          "type": "object",
          "properties": {
            "query": {
              "type": "string",
              "description": "Lookup query"
            }
          },
          "required": ["query"]
        }
      }
    }
  ],
  "max_tokens": 256,
  "stream": false
}
```

4. Repeat the exact byte-identical request at least 10 times.
5. Compare the tool-schema ordering and `prompt_token_ids` seen by the prefill
   and decode workers, and inspect SGLang cached-token metrics.

Actual result:

- Tool-schema field order may differ between the generated request bodies.
- `prompt_token_ids` hashes are not stable.
- Repeated requests reuse only a partial page-aligned prefix.

### Expected behavior

AIBrix should preserve all prompt-bearing fields when injecting PD bootstrap
metadata.

For each request:

- Prefill and decode must receive byte-identical `messages`, `tools`, and other
  untouched fields.
- The same client request must produce stable `prompt_token_ids`.
- `bootstrap_host`, `bootstrap_port`, and `bootstrap_room` must still match
  between prefill and decode.
- Decode must retain the client's generation parameters.
- Prefill may continue to override only its existing control fields
  (`max_tokens`, `max_completion_tokens`, `stream`, `stream_options`, and
  `min_tokens`).
- Repeated requests should reuse the full page-aligned cacheable prefix after
  warmup.

### Environment

- AIBrix version: `v0.7.0` and `main@e397f9866e19866abd943b93244f3dd03b5873cb`
- Deployment environment: Kubernetes, SGLang disaggregated prefill/decode
- SGLang image/version: `<fill exact image tag>`
- Model: `<fill exact model and served-model name>`
- Transfer backend: `<NIXL or Mooncake>`
- Client: OpenAI-compatible `/v1/chat/completions`
- GPU / Kubernetes version: `<fill deployment details>`

qidaye:
## Pull Request Description

  AIBrix SGLang PD routing re-serialized the complete OpenAI request through`map[string]any` twice (once for decode, once for prefill) via `sonic.Marshal`, whose default configuration does not sort map keys. Nested objects such as
  tool schemas could be emitted in different orders between the prefill and decode bodies, or between repeated identical requests. SGLang serializes tool definitions as part of the chat template, so a changed field order
  produces different `prompt_token_ids` and degrades KV cache hit rates.

  ### Root Cause

  1. `prefill.PreparePayload` unmarshals `routingCtx.ReqBody` into  `map[string]any`.
  2. `SGLangHandler.AugmentPrefillRequest` marshals the map back to create the decode body.
  3. `PreparePayload` changes prefill generation fields and marshals the same map again.

  Because `sonic.Marshal` does not sort map keys by default, nested objects (e.g. OpenAI tool schemas) can be emitted in different field orders across these two serializations, changing `prompt_token_ids` even though the JSON remains semantically equivalent.

  ### Fix

  Introduce a `RawPrefillPayloadPreparer` optional interface on `EngineHandler`.
  `SGLangHandler` implements it using `gjson`/`sjson` to mutate only top-level fields (`bootstrap_host`, `bootstrap_port`, `bootstrap_room`, `max_tokens`, `max_completion_tokens`, `stream`, `stream_options`, `min_tokens`) on the
  original JSON bytes, preserving the byte ordering of all nested fields (`messages`, `tools`, …).

  Key design decisions:

  - **`ValidateSGLangRequest`** runs at `Route()` entry, before  `podSelector.Select`, so duplicate controlled fields are rejected without  polluting prefix index or selection counters. Covers both separate-PD and combined-pod paths.
  - **`InvalidRequestError`** typed error is checked in `gateway_req_body.go` to map client-side errors to HTTP 400 instead of the default 503.
  - **`PreparePrefillPayload`** also calls `ValidateSGLangRequest` to protect direct callers of the `RawPrefillPayloadPreparer` interface.
  - `sjson` promoted from indirect to direct dependency in `go.mod`.
  - Only SGLang `pd` routing is affected; vLLM, TRT-LLM, and combined paths  are unchanged.

## Related Issues
Resolves: #2475 

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>


varungup90:
### ⚡ Perf Nit: Redundant JSON validation on the SGLang prefill path

`PreparePrefillPayload` → `ValidateSGLangRequest` → `prepareSGLangRequestBodies` currently validates the same request body **three times** per request:

1. `Route()` calls `ValidateSGLangRequest` (`pd_disaggregation.go`)
2. `PreparePrefillPayload` calls `ValidateSGLangRequest` again (`sglang.go:217`)
3. `prepareSGLangRequestBodies` performs its own `gjson.ValidBytes` + `IsObject` check (`sglang.go:176`)

Because each check performs a full O(n) scan over the payload, the overhead adds up on larger requests:

| Body Size | Current (3x validate) | Single Validate | Overhead |
| :--- | :--- | :--- | :--- |
| **~1 KB** | 13.2 µs | 9.9 µs | **+33%** |
| **~12 KB** | 80.2 µs | 55.5 µs | **+31%** |
| **~106 KB** | 593.7 µs | 378.4 µs | **+36%** |
| **~396 KB** | 2.13 ms | 1.36 ms | **+36%** |

> **Note:** In absolute terms, this stays under ~1 ms even for ~400 KB tool-heavy requests against typical inference latencies, so this is **non-blocking**. 

Since `PreparePrefillPayload` is only called after `Route()` has already validated `ctx.ReqBody`, and `prepareSGLangRequestBodies` is internal to `PreparePrefillPayload`, two of the three passes are purely redundant.

#### Suggestion
* Keep the **`Route()`-level check** (necessary to avoid polluting prefix-index / selection counters prior to pod selection).
* Drop the duplicate `ValidateSGLangRequest` call inside `PreparePrefillPayload` and the `ValidBytes`/`IsObject` re-check in `prepareSGLangRequestBodies` (or add a brief comment if kept purely for defense-in-depth against external callers).

Non-blocking, but a nice free win on the hot path for large-context requests!

qidaye:
@varungup90 Thank you for the detailed review and the thorough performance analysis!
                                       
  Following your suggestions, I've made the following changes:
                                                                                                                                                                                             
 - Dropped the duplicate `ValidateSGLangRequest` call in `PreparePrefillPayload` — since `Route()` already validates the request body before pod selection, this was purely redundant.              
 - Kept the `ValidBytes/IsObject` guard in `prepareSGLangRequestBodies` with a defense-in-depth comment — as this is an internal utility function, retaining a lightweight JSON validity check    
  prevents sjson from silently producing invalid output if a future caller bypasses ValidateSGLangRequest. 
 - Kept the `Route()`-level check unchanged — necessary to reject malformed requests before any pod selection or prefix-index mutation.

  The relevant tests have been updated to reflect the new validation contract. Please let me know if anything else needs to be adjusted. Thanks again!

varungup90:
Can you sign it?

qidaye:
> Can you sign it?

Sorry, my mistake. Fixed.

varungup90 (approved):

PR #1894, opened 2026-01-16
[Docs][API] Add AWS Neuron/Trainium2 support for disaggregated inference

yahavb:
## Pull Request Description

This PR adds support for AWS Trainium2 (Neuron) instances in the aibrix disaggregated inference (P/D) routing implementation. The changes enable aibrix to route requests through the prefill→decode flow using NIXL for KV cache transfer over EFA (Elastic Fabric Adapter).

### Changes

1. **Documentation**: Added comprehensive setup guide for Neuron disaggregated inference (`docs/neuron-disaggregated-inference-setup.md`) covering:
   - EKS nodegroup configuration with EFA networking
   - Separate prefill/decode Deployment architecture for independent scaling
   - Aibrix cluster setup and configuration
   - Gateway plugin code changes for pd routing
   - Testing procedures

2. **P/D Routing Logic** (in `pkg/plugins/gateway/algorithms/pd_disaggregation.go`):
   - Pod filtering by `role-name` label (prefill/decode)
   - Prefill request modification (`max_tokens=1`)
   - KV transfer params extraction and forwarding to decode

3. **HTTPRoute Validation** (in `pkg/plugins/gateway/gateway.go`):
   - Skip HTTPRoute status validation for pd routing algorithm

### Request Flow

```
Client → Gateway → Prefill Pod (max_tokens=1) → NIXL KV Transfer → Decode Pod → Response
```

### Pod Configuration

Users must label pods with:
- `role-name: prefill` or `role-name: decode`
- `model.aibrix.ai/name: <model-name>`

And annotate with:
- `model.aibrix.ai/port: "8000"`

### Testing

Tested with:
- AWS Trainium2 instances (trn2.48xlarge)
- vLLM with Neuron support and NIXL connector
- Llama 3.1 8B model
- EFA-enabled EKS cluster

## Related Issues

Resolves: #N/A (New feature - Neuron/Trainium2 support for disaggregated inference)

---

### Submission Checklist

- [x] PR title includes appropriate prefix(es): `[Docs][API]`
- [x] Changes are clearly explained in the PR description
- [x] New and existing tests pass successfully
- [x] Code adheres to project style and best practices
- [x] Documentation updated to reflect changes
- [x] Thorough testing completed, no regressions introduced

PR #2490, opened 2026-07-23
docs(aws): Neuron (Trainium) DRA device allocation for disaggregated inference

yahavb:
## What

Extends the **Manual** AWS install guide (`docs/source/getting_started/installation/aws.rst`) with a section on running **disaggregated inference (prefill/decode) on AWS Neuron (Trn2 / Trn3)** using Dynamic Resource Allocation.

## Details

- **Why DRA:** on Neuron the KV cache moves prefill→decode over NIXL/`LIBFABRIC` on EFA, so each pod needs NeuronCores **and** EFA devices from the **same PCIe/NUMA group**. The section shows a `ResourceClaimTemplate` that pairs `neuron.aws.com` + `efa.networking.k8s.aws` under one `devicegroup8_id` constraint.
- Prereqs: EFA-enabled nodegroup + the Neuron and EFA DRA drivers (linked to the authoritative docs).
- `fi_pingpong` fabric-validation tip and the pointer to enable `ROUTING_ALGORITHM=pd`.

## Also

- Trims the long `eksctl create cluster` console log dump from the Manual steps — keeps just the command.

### DIFF
--- a/docs/source/getting_started/installation/aws.rst
+++ b/docs/source/getting_started/installation/aws.rst
@@ -80,6 +80,10 @@ valid. The recommended way to request them together is Kubernetes
 Prerequisites
 ~~~~~~~~~~~~~
 
+- Kubernetes **1.34 or later**, where Dynamic Resource Allocation is GA and the
+  API group is ``resource.k8s.io/v1`` (a fixed device request is wrapped in
+  ``exactly:``). On 1.30–1.33 DRA is beta (``resource.k8s.io/v1beta1``) with a
+  different request schema.
 - An EKS nodegroup of ``trn2.48xlarge`` (or ``trn3-dev1.48xlarge``) with EFA
   enabled (``efaEnabled: true`` and ``privateNetworking: true`` in the eksctl
   nodegroup config). EFA capacity is typically obtained through an on-demand
--- a/docs/source/getting_started/installation/aws.rst
+++ b/docs/source/getting_started/installation/aws.rst
@@ -30,82 +30,11 @@ Prerequisites
 Steps
 ~~~~~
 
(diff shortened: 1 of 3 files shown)


yahavb:
Thanks for the review! This is accurate for the **beta** DRA API (Kubernetes 1.30-1.33), but these manifests target **GA DRA (Kubernetes 1.34+)**:

- DRA graduated to GA in 1.34, promoting the API group to `resource.k8s.io/v1`.
- In the GA restructure, a fixed device request is wrapped in `exactly:` (an `ExactDeviceRequest` holding `deviceClassName`/`allocationMode`/`count`/`selectors`), with `firstAvailable:` as the alternative — so `resource.k8s.io/v1` **with** `exactly:` is the correct GA schema.

Verified on a live 1.35 cluster: `kubectl get resourceclaimtemplate ... -o yaml` round-trips these objects in exactly this form, and they allocate Neuron + EFA and serve disaggregated inference as-is. Using `v1beta1` + flat fields would fail validation on 1.34+.

I pushed a follow-up commit adding a prerequisite note that these require Kubernetes 1.34 or later (GA DRA), to avoid this confusion for readers on older clusters.

yahavb:
For context: this PR extends the AWS Neuron / Trainium disaggregated-inference work from #1894, documenting the DRA-based Neuron + EFA device allocation that setup relies on.

It also pairs with an end-to-end EKS example that simplifies the disaggregated-inference-on-AIBrix case (cluster, EFA nodegroup, DRA claims, prefill/decode deploys, and routing): https://github.com/aws-neuron/aws-neuron-eks-samples/tree/disagg-serve-vllm/disagg_serve_vllm

More docs to come.

varungup90 (approved):

PR #2496, opened 2026-07-25
[Bug] Reserve space for service suffix so generated resource names stay within 63-char limit

### LINKED ISSUE #2493: [Bug] Console deployment fails for long names when the generated Service name exceeds 63 characters
A long deployment name can make the Console deployment API return 500 even though the user-facing name itself is accepted.

The problem is that [`generateResourceName`](https://github.com/vllm-project/aibrix/blob/a40be9ba404730e4be2d2a0820942bca79275dc6/apps/console/api/deployment/provider/kubernetes.go#L916-L927) may return a 63-character name, and the Service builder then [appends `-svc`](https://github.com/vllm-project/aibrix/blob/a40be9ba404730e4be2d2a0820942bca79275dc6/apps/console/api/deployment/provider/kubernetes.go#L247-L252). The final Service name is therefore too long for Kubernetes.

I reproduced this with the dev `mock-vllm` template and a 100-character deployment name:

```text
Service "...-svc" is invalid: metadata.name: Invalid value: "...-svc":
must be no more than 63 characters
```

The Deployment is rolled back, but the request still fails with an internal error.

Expected behavior: all generated resource names should stay within the Kubernetes limit. We should reserve room for resource-specific suffixes while keeping the random suffix, or reject the name during validation with a useful error.

The existing [long-name test](https://github.com/vllm-project/aibrix/blob/a40be9ba404730e4be2d2a0820942bca79275dc6/apps/console/api/deployment/provider/kubernetes_test.go#L245-L255) only checks the base name. Please add boundary coverage for the final Service name as part of the fix.

Reproduced on `main` at `a40be9ba404730e4be2d2a0820942bca79275dc6` with Docker Desktop Kubernetes v1.30.5.

Related: #2202, #2198

addielaruee:
### Problem
The Console Kubernetes provider's `generateResourceName` allowed the base resource name to occupy the **full 63-character DNS label limit**. Callers then append a fixed suffix to derive related resources — most importantly the Service, named `resourceName + "-svc"`. When the base already filled 63 characters, the Service name reached 67 and the API server rejected it:

```
create service "aibrix-…-svc": Service "…-svc" is invalid: metadata.name:
Invalid value: "…-svc": must be no more than 63 characters
```

Because the Deployment is created first, the request failed with HTTP 500 *after* the Deployment already existed (the rollback removed the Deployment, but the otherwise-valid user-facing request still failed).

### Fix
`generateResourceName` now reserves room for the longest suffix appended to the base name (the Service's `-svc`) in addition to the prefix, the joining dash, and the unique suffix — so every derived name stays within the limit. To make the invariant explicit and drift-proof (the root cause was the generator not knowing about the suffix callers append), the magic values are extracted into named constants and the `"-svc"` literal is replaced with `serviceNameSuffix` at every use site:

- `maxResourceNameLength` (63), `resourceNamePrefix` (`aibrix-`), `resourceNameUniqueSuffixLength` (8), `serviceNameSuffix` (`-svc`)
- The base name is truncated to `maxResourceNameLength - len(prefix) - len("-") - suffixLen - len(serviceNameSuffix)` before the unique suffix is appended, and the random suffix is always preserved.

Deployment and HPA names use the base name directly (no appended suffix), so they were already within the limit and remain so.

### Testing
`go test ./apps/console/api/deployment/provider/...`, `go vet`, `gofmt`, and `go build ./apps/console/...` all pass. Added two tests (verified they fail without the fix):

- `TestGenerateResourceNameDerivedNamesStayWithinKubernetesLimit` — table test over names **below, at, and above** the boundary (incl. the 100-char case from the report). Asserts the **final** `resourceName + "-svc"` — not just the base — is a valid label via `validation.IsDNS1035Label`, and that the unique suffix survives truncation.
- `TestCreateWithLongNameProducesValidKubernetesResourceNames` — end-to-end `Create` with a 100-character name that asserts the actual created Deployment, Service, and HPA object names are valid DNS labels. (The fake client does not enforce API-server name validation, so names are validated explicitly, as the issue suggests.)

### Acceptance criteria
- [x] Generated Deployment, Service, and HPA names satisfy their Kubernetes naming limits.
- [x] Long deployment names no longer cause Service creation to fail with HTTP 500.
- [x] The random suffix remains present after truncation.
- [x] Tests cover names below, at, and above the boundary.
- [x] Tests assert the validity of the final `resourceName + "-svc"`, not only the base name.

## Related Issues
Resolves: #2493

### DIFF
--- a/apps/console/api/deployment/provider/kubernetes.go
+++ b/apps/console/api/deployment/provider/kubernetes.go
@@ -59,6 +59,19 @@ const (
 	defaultReplicaCount            int32 = 1
 	defaultReadyTimeoutSeconds     int32 = 600
 	kubernetesCleanupTimeout             = 30 * time.Second
+
+	// maxResourceNameLength is the Kubernetes DNS label limit (RFC 1035/1123) that
+	// applies to the Deployment, Service, and HPA names created by this provider.
+	maxResourceNameLength = 63
+	// resourceNamePrefix is prepended to every generated resource name.
+	resourceNamePrefix = "aibrix-"
+	// resourceNameUniqueSuffixLength is the length of the random suffix appended to
+	// generated names to keep them collision-free across deployments.
+	resourceNameUniqueSuffixLength = 8
+	// serviceNameSuffix is appended to the generated base name to derive the Service
+	// name. generateResourceName reserves room for it so the derived Service name also
+	// stays within maxResourceNameLength.
+	serviceNameSuffix = "-svc"
 )
 
 // KubernetesClientProvider resolves a Kubernetes client and its workload
@@ -244,7 +257,7 @@ func (d *kubernetesDeploymentProvider) Create(ctx context.Context, template *pb.
 	}
 	deploymentCreated := true
 
-	serviceName := resourceName + "-svc"
+	serviceName := resourceName + serviceNameSuffix
 	service := buildService(serviceName, namespace, labels, d.workload)
 	if _, err := clientset.CoreV1().Services(namespace).Create(ctx, service, metav1.CreateOptions{}); err != nil {
 		createErr := status.Errorf(codes.Internal, "create service %q: %v", serviceName, err)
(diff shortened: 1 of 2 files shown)


Jeffwan (approved):
overall looks good to me. thanks

PR #2504, opened 2026-07-28
[Batch] fix: use dedicated file upload timeout

scarlet25151:
## Pull Request Description

Large multipart uploads can exceed the metadata client's 60-second timeout even when the metadata service stores the file successfully. Use a configurable upload-specific timeout while preserving the shorter timeout for regular file operations and inbound request cancellation.

## Timeout Behavior

- Upload requests use `METADATA_FILE_UPLOAD_TIMEOUT`, which defaults to `30m` and accepts any positive Go duration.
- The 30-minute default is a conservative finite ceiling intended to prevent the proxy timeout from becoming the bottleneck for large uploads. Deployments can reduce it as needed.
- The inbound request context is propagated to the metadata request, so client disconnection cancels the upload early.
- List, metadata, and download operations retain the existing 60-second timeout.
- A server-level `ReadTimeout` would independently limit inbound uploads. This dependency is documented next to the HTTP server configuration.

## Related Issues

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>

### DIFF
--- a/apps/console/api/handler/file.go
+++ b/apps/console/api/handler/file.go
@@ -46,12 +46,18 @@ const (
 type FileHandler struct {
 	metadataServiceURL string
 	httpClient         *http.Client
+	uploadHTTPClient   *http.Client
 	store              store.Store
 	injector           error_injection.Injector
 }
 
 // NewFileHandler creates a new file proxy handler.
-func NewFileHandler(metadataServiceURL string, injector error_injection.Injector, stores ...store.Store) *FileHandler {
+func NewFileHandler(
+	metadataServiceURL string,
+	uploadTimeout time.Duration,
+	injector error_injection.Injector,
+	stores ...store.Store,
+) *FileHandler {
 	var st store.Store
 	if len(stores) > 0 {
 		st = stores[0]
@@ -59,6 +65,7 @@ func NewFileHandler(metadataServiceURL string, injector error_injection.Injector
 	return &FileHandler{
 		metadataServiceURL: strings.TrimRight(metadataServiceURL, "/"),
 		httpClient:         &http.Client{Timeout: fileHTTPClientTimeout},
+		uploadHTTPClient:   &http.Client{Timeout: uploadTimeout},
 		store:              st,
 		injector:           injector,
 	}
@@ -91,7 +98,7 @@ func (h *FileHandler) handleUpload(w http.ResponseWriter, r *http.Request, _ map
 		}
 	}
 	targetURL := h.metadataServiceURL + "/v1/files"
-	h.proxyRequest(w, r, "POST", targetURL, "upload", start)
+	h.proxyRequest(w, r, "POST", targetURL, "upload", start, h.uploadHTTPClient)
 }
 
 // handleList proxies file listing to GET {metadataServiceURL}/v1/files.
@@ -101,7 +108,7 @@ func (h *FileHandler) handleList(w http.ResponseWriter, r *http.Request, _ map[s
 	if r.URL.RawQuery != "" {
 		targetURL += "?" + r.URL.RawQuery
 	}
-	h.proxyRequest(w, r, "GET", targetURL, "list", start)
+	h.proxyRequest(w, r, "GET", targetURL, "list", start, h.httpClient)
 }
 
 // handleGetMetadata proxies file metadata retrieval to GET {metadataServiceURL}/v1/files/{file_id}.
@@ -112,7 +119,7 @@ func (h *FileHandler) handleGetMetadata(w http.ResponseWriter, r *http.Request, 
 	if r.URL.RawQuery != "" {
 		targetURL += "?" + r.URL.RawQuery
 	}
-	h.proxyRequest(w, r, "GET", targetURL, "get_metadata", start)
+	h.proxyRequest(w, r, "GET", targetURL, "get_metadata", start, h.httpClient)
 }
 
 // handleDownloadContent proxies file content download to GET {metadataServiceURL}/v1/files/{file_id}/content.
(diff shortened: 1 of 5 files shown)


apps/console/api/handler/file.go
```diff
@@ -194,7 +201,20 @@ func (h *FileHandler) authorizeFileDownload(ctx context.Context, fileID string)
 }
 
 // proxyRequest forwards an HTTP request to the target URL and copies the response back.
-func (h *FileHandler) proxyRequest(w http.ResponseWriter, r *http.Request, method, targetURL, op string, start time.Time) {
+func (h *FileHandler) proxyRequest(
+	w http.ResponseWriter,
+	r *http.Request,
+	method, targetURL, op string,
+	start time.Time,
+	client *http.Client,
```
DwyaneShi: why do we introduce the `client` parameter explictly rather than getting it from handler?

ErenAta16 (commented):
Splitting the upload client out is the right call and the change reads cleanly. Two things I checked while going through it.

**The inbound side is what actually bounds a long upload, and it's already unbounded here.** `proxyRequest` builds the outbound request with `http.NewRequestWithContext(r.Context(), ...)`, so the upload dies with the inbound request regardless of which client is used. That's the "preserving inbound request cancellation" the description mentions, and it works because `server.go:279` constructs `&http.Server{Addr, Handler}` with no `ReadTimeout` or `WriteTimeout`:

```go
s.httpServer = &http.Server{
    Addr:    httpAddr,
    Handler: httpHandler,
}
```

So today nothing on the server side cuts a 30-minute upload, and this change does what it says. Worth knowing that it depends on those fields staying unset — if someone later adds a `ReadTimeout` for slowloris protection, it will silently cap uploads again and the new `METADATA_FILE_UPLOAD_TIMEOUT` will look like it stopped working. A comment next to the `http.Server` literal saying uploads rely on the absence of a read timeout would save that debugging session.

**`fileHTTPClientTimeout` stays at 60s for everything else**, which is the part I'd have got wrong if I'd only skimmed — download is also proxied through the same helper, and a large file coming *back* through `handleDownload` is on the 60-second client. Uploads were the reported failure, so scoping the fix to them is defensible; just noting the symmetric case exists and isn't covered, in case the same report comes back for downloads.

Smaller: `defaultMetadataFileUploadTimeout = 30 * time.Minute` is a big jump from 60s. Since the client-side context still governs, an abandoned connection won't hold a goroutine for 30 minutes, so the risk is low — but it's worth stating in the PR description why 30 minutes rather than, say, 5, since that number becomes the ceiling for how long a stuck metadata service can hold a proxy request when the client *doesn't* disconnect.

The env var going through `envDurationOrDefault` with the error propagated out of `Load()` matches how the other timeouts in that file are handled, and the test covering the parse failure path is the right thing to pin.

scarlet25151:
hi @ErenAta16 Thanks for the thorough review. The scope is intentionally upload-only, so downloads and other file operations remain at the existing 60-second timeout and can be addressed separately if needed. I agree that a future server-level ReadTimeout could silently cap uploads again, so I’ll document that dependency next to the http.Server construction. The 30-minute default is a conservative finite ceiling intended to keep the proxy timeout from becoming the bottleneck for large uploads; it remains configurable, while request-context cancellation releases abandoned uploads early. I’ll clarify this rationale in the PR description as well.

apps/console/api/handler/file.go
scarlet25151: I updated proxyRequest to resolve the client from FileHandler through clientForOperation instead of accepting it as an explicit parameter.

ErenAta16 (commented):
Both things you said you would do have landed, and I checked them rather than taking the summary.

The `ReadTimeout` note is in the right place, immediately above the `http.Server` literal in `StartHTTP` where someone adding the field will actually be looking:

```go
// A server-level ReadTimeout would independently cap inbound file uploads.
// Keep it aligned with MetadataFileUploadTimeout if one is introduced.
s.httpServer = &http.Server{
```

That is the failure I was worried about. A server-level read deadline would cut the request body regardless of what the outbound client allows, and the symptom would look like the metadata service dropping large uploads rather than like a config change. Written next to the construction, it is hard to miss; written in the PR description it would have been gone in a month.

`clientForOperation` reads correctly. Resolving the client from the handler by operation, rather than threading it through `proxyRequest`'s parameter list, keeps the upload/non-upload decision in one place, so a future caller cannot pass the wrong one.

I did chase the `client == nil` fallback to `http.DefaultClient`, since `DefaultClient` carries no timeout and that would quietly undo the whole point of the change. It is not reachable: `NewFileHandler` sets both `httpClient` and `uploadHTTPClient` unconditionally, and `server.go:238` is the only construction path. So the guard is defensive rather than live, and there is nothing to change. Noting it only so you know it was checked.

On scope, upload-only is the right call and I am not asking for more. My point was about knowing which operations kept the 60-second bound rather than about widening the change, and the description now says so.

`envDurationOrDefault` rejecting a non-positive duration is worth keeping. `METADATA_FILE_UPLOAD_TIMEOUT=0` would otherwise parse cleanly into "no timeout at all", which is the one value a reader would assume means "disabled" and which behaves as the opposite of a conservative ceiling.

Nothing blocking from me.

DwyaneShi (approved):
LGTM

ErenAta16:
Following up now this is merged: you did the thing you said you would, and it landed where it needed to.

```go
// apps/console/api/server/server.go:292
// A server-level ReadTimeout would independently cap inbound file uploads.
// Keep it aligned with MetadataFileUploadTimeout if one is introduced.
s.httpServer = &http.Server{
```

That is the whole value of the point I raised. The failure mode was never the timeout you set, it was that someone adding a `ReadTimeout` later would have no way to know it silently caps uploads, and would have had to rediscover it from a truncated upload in production. A comment at the construction site is where that person actually looks.

Agreed on upload-only being the right scope. Downloads have a different failure profile and bundling them would have made this harder to reason about for no gain.

Nothing further from me on this one.

PR #2506, opened 2026-07-29
improve tracing logic and span name, add span attr

rayne-Li:
## Pull Request Description

This PR improves gateway observability by restructuring request traces and enriching the main request span with the routing and inference metadata already emitted in structured logs. It also updates the ESX Helm chart defaults for the current gateway release and makes the GPU optimizer optional.

- Rename gateway processing spans to consistent hierarchical names:
  - `process.handle_request_headers`
  - `process.handle_request_body`
  - `process.select_target_pod`
  - `process.handle_response_headers`
  - `process.response_error_processing_with_headers`
- Store the main request span in `RoutingContext` so downstream request/response handlers can add attributes.
- Record candidate/ready pod counts and the routing strategy during target selection.
- Record request ID, path, model, streaming mode, selected pod, outstanding requests, and routing time when inference starts.
- Convert request-end structured log fields into OpenTelemetry attributes and attach them to the main span.
- Add `llm.inference` tracing for inference processing.
- Split streaming latency into `llm.time_to_first_response_chunk` spans.

- non-stream trace
image

- stream trace
image

image

## Related Issues
Resolves:


rayne-Li:
image
@whalepark miss License at those files leading to CI failed

rayne-Li:
@varungup90 please take a look

pkg/plugins/gateway/gateway_req_headers.go
```diff
@@ -74,9 +74,6 @@ func (s *Server) HandleRequestHeaders(ctx context.Context, requestID string, req
 			reqHeaders[n.Key] = string(n.RawValue)
 		case HeaderSessionKey:
 			reqHeaders[n.Key] = string(n.RawValue)
-		case HeaderTraceParent:
```
googs1025, line 77: why remove this part

pkg/plugins/gateway/gateway.go
googs1025, line 533: If hit this condition, will the trace be lost?

varungup90:
Thanks for this PR! The new tracing spans and structured logging attributes are a great addition. I reviewed the changes and have a few findings, including a broken test and some edge cases with span attribute recording.

### 🚨 Main Findings

**1. Test no longer validates what it claims (`gateway_test.go`)**
The removal of `traceparent`-based request-ID derivation (see #2 below) breaks the premise of `TestProcess_CleansRequestBufferOnExitBeforeResponseEnd`.

* **The issue:** The test does `requestBuffers.Delete(requestID)` and asserts `requestBuffers.Load(requestID)` is absent, using a `requestID` constant embedded in a fake `traceparent` header. Since `HandleRequestHeaders` no longer reads `traceparent` to set `st.requestID`, and the mock's `Context()` returns a bare `context.Background()` (no span), the server now generates a random UUID internally. This is completely disconnected from the `requestID` the test checks against.
* **Impact:** The assertion passes trivially because that key was never used. The test silently stopped testing the buffer-cleanup behavior it's named for.
* **Action:** This is worth fixing or removing.

**2. Loss of trace correlation with client-supplied `traceparent` (`gateway_req_headers.go`)**
Previously, the code parsed the inbound HTTP `traceparent` header and used its trace ID as the request ID (`GetTraceID`), letting the gateway's request ID line up with a client-originated trace.

* **The issue:** This is now removed entirely (`HeaderTraceParent` and `GetTraceID` are dead code, only referenced in tests). The new request ID instead comes from `trace.SpanFromContext(srv.Context())` (i.e., whatever span `otelgrpc`'s stats handler attaches to the Envoy→gateway gRPC stream).
* **Impact:** Unless Envoy is explicitly configured to propagate the client's trace context into that gRPC call's metadata, this severs correlation between the gateway's traces/request IDs and the client's original trace.
* **Action:** Is this intentional? If so, we should remove the now-dead `GetTraceID`/`HeaderTraceParent` code and its associated tests.

**3. `selectTargetPod` skips attribute recording on common paths (`gateway.go`:490-536)**
The new `SetAttributes(candidate_pods, ready_pods, routing_strategy)` call sits *after* the single-ready-pod fast path (`if len(readyPods) == 1 && ... { return ... }`) and after all early error returns (no pods, no ready pods, label-selector failure, engine validation failure, routing.Select failure).

* **Impact:** Single-pod deployments (a very common case in dev/small clusters) and any routing failures will never get these attributes on the span, which is arguably where they are most useful for debugging.
* **Action:** Consider moving the `SetAttributes` call up, right after `readyPods` and `pods.Len()` are known, so it's recorded regardless of the outcome.

**4. `routing_time_taken` is stored as a string (`gateway_req_body.go`:170)**
Currently, `attribute.String("routing_time_taken", routingDelay.String())` stores a duration as free-text (e.g., `"12.3ms"`). Most tracing backends can't aggregate or filter numeric ranges on string attributes.

* **Action:** Consider using `attribute.Int64("routing_time_taken_ms", routingDelay.Milliseconds())` instead, which is consistent with how durations are usually recorded for OTel.

---

### 💡 Minor / Style

* **Type mismatch:** `attribute.Float64("outstanding_requests_at_start", request_count)` records a request count as a float. If `getRunningRequestsByPod` truly returns a count, an `Int` or `Int64` attribute would be more natural for downstream querying.

---

### ✅ What Looks Good

* **`fieldsToAttributes` (`util.go`):** This is a reasonable, well-tested helper with good edge-case coverage for odd-length/unsupported types. No issues there.
* **Span Lifecycle Management:** In `Process()` / `processOnce`, the lifecycle looks correct. `rootSpan` is intentionally not `.End()`'d (owned by the gRPC stats handler), while `inferenceSpan`, `firstRespSpan`, and `toLastRespSpan` are all covered by the completion path and the deferred cleanup. `RoutingContext.reset()` correctly clears `Span = nil` to avoid stale references when contexts are pooled.
* **Span Gating:** Gating `llm.inference` / `llm.time_to_first_response_chunk` span creation on `resp.GetImmediateResponse() == nil` is a great improvement over the old code, which unconditionally created the TTFT span even for rejected requests.

pkg/plugins/gateway/gateway_req_headers.go
rayne-Li: trace.SpanFromContext(srv.Context()) will extract this header and get trace-id from envoy stream. so we don't need to handle 'traceparent' in header

pkg/plugins/gateway/gateway.go
rayne-Li: i will also add span.attr in this condition

rayne-Li:
@varungup90 @googs1025  thanks for your review

now we use openTelemetry package automatically extra span and requestID by
```
func (s *Server) Process(srv extProcPb.ExternalProcessor_ProcessServer) error {
	rootSpan := trace.SpanFromContext(srv.Context())
	requestID := uuid.New().String()
	if rootSpan.SpanContext().HasTraceID() {
		requestID = rootSpan.SpanContext().TraceID().String()
	}
```
which will make rootSpan start more accurate than in handle_request_header, 
retrieving the trace ID directly via rootSpan.SpanContext().TraceID().String(). This is the standard OpenTelemetry API and relies on the framework’s built-in 

> 1. Test no longer validates what it claims (gateway_test.go)
traceparent still exist(created by envoy) and test will succeed

> 2. Loss of trace correlation with client-supplied traceparent (gateway_req_headers.go)
same as above, 

> 3. selectTargetPod skips attribute recording on common paths
fixed in last commit

> 4.routing_time_taken is stored as a string
changed in last commit

varungup90:
@rayne-Li fix the CI else LGTM

ErenAta16 (commented):
The span restructuring reads clearly, but one removal in `gateway_req_headers.go` changes behaviour beyond tracing and I couldn't find what replaces it.

```go
-		case HeaderTraceParent:
-			reqHeaders[n.Key] = string(n.RawValue)
-			requestID = GetTraceID(string(n.RawValue), requestID)
```

Two separate things go with it.

**`requestID` stops being derived from the incoming `traceparent`.** That value is what appears in every `klog` line, in `request_end`, and in the `request_id` span attribute this PR adds. Today a caller sending `traceparent: 00-<trace-id>-...` gets logs keyed by that trace id, so a request can be followed across the gateway and whatever sits in front of it. After this the gateway uses its own `uuid.New()` and the correlation is gone. `Process` now takes `trace.SpanFromContext(srv.Context())` instead — but I couldn't find an `otelgrpc` handler or interceptor anywhere under `pkg/` or `cmd/`, and no `grpc.NewServer` wiring that installs one, so I don't think the gRPC context carries the caller's span context. If that's right, `SpanFromContext` yields a non-recording span and the header was the only path in. If there *is* wiring I missed, saying where in the PR body would settle it.

**`traceparent` also stops being stored in `reqHeaders`.** Worth checking whether anything downstream reads it out of that map.

`GetTraceID` becomes unreferenced outside its own test after this — `grep` finds it only at the removed call site, its definition in `util.go:965`, and `TestGetTraceID` in `util_test.go:1321`. If dropping the correlation is deliberate, the function and its test should go too; if it isn't, this is the signal that something was lost.

`gateway_test.go:1481` builds a request whose `traceparent` embeds `requestID` and asserts on it, so that test is presumably where this shows up — worth confirming it still passes for the right reason rather than incidentally.

Smaller, on `fieldsToAttributes`: the loop is `i < len(fields)-1; i += 2`, so an odd-length `fields` silently drops the trailing key. That matches klog's own behaviour for unpaired args, so it's defensible, but a `len(fields)%2 != 0` guard with a log line would surface a caller mistake instead of hiding it. Also `%!s(MISSING)`-style keys from `fmt.Sprint` on a non-string key would become attribute names as-is; every current caller passes string keys, so it's theoretical.

The rename from `HandleRequestBody` to `process.handle_request_body` and the four-span split (`rootSpan` / `inferenceSpan` / `firstRespSpan` / `toLastRespSpan`) both look right to me, and the comments on the struct fields make the boundaries readable.

rayne-Li:
@ErenAta16  thanks for your review

1. traceID still derived from traceparent, and grpcHandler inject logic is in plugins/main.go L191~203, so klogs and span.attr will store right traceID
2. TestGetTraceID is free to delete

> 		case HeaderTraceParent:
>			reqHeaders[n.Key] = string(n.RawValue)
>			requestID = GetTraceID(string(n.RawValue), requestID)

3. gateway-plugins is only works for routing and getting target-pod-ip and inject it in header, next stream is called by envoy-gateway, traceparent is passed by envoy-gateway after routing(not by gateway-plugins), so remove codes above leading a miss-transfer of traceparent only appears at PD-disaggregation, let's add it back
4. The current behavior is intentional and covered by tests: an incomplete trailing field is ignored. Also, fmt.Sprint does not produce %!s(MISSING) for non-string keys—for example, fmt.Sprint(99) returns "99", which is already covered by the test. Since all current callers provide paired fields with string keys, I’d prefer to keep this unchanged.

ErenAta16 (commented):
You are right on both of the points where you pushed back, and I should have found the first one myself.

**On the gRPC span context.** I wrote that I could not find an `otelgrpc` handler under `pkg/` or `cmd/`. It is in `cmd/plugins/main.go`, exactly where you said:

```go
otelEnabled := utils.OTELEnabled()
if otelEnabled {
    ...
    opts = append(opts, grpc.StatsHandler(otelgrpc.NewServerHandler()))   // 202
}
...
s := grpc.NewServer(opts...)                                             // 208
```

The server handler extracts the incoming trace context from gRPC metadata, so `trace.SpanFromContext(srv.Context())` does carry the caller's trace id and the `request_id` attribute is correct. My search was on the wrong paths and I stated the conclusion far more confidently than the checking behind it justified. Sorry for sending you to defend something that was never broken.

**On `fmt.Sprint`, you are right and my example was wrong.** `%!s(MISSING)` comes from `Sprintf` when a format string has more verbs than arguments; `Sprint` has no format string and cannot produce it. `fmt.Sprint(99)` is `"99"`, as you say. What is left of the point is only that a non-string key becomes its default string form as an attribute name, which with every caller passing string keys is not worth a branch. Agreed on leaving it alone, and agreed on the trailing-field behaviour being intentional and tested.

**One thing to watch when you apply point 3.** Restoring the `HeaderTraceParent` case brings `GetTraceID` back into use, so point 2 stops being true at the same moment. If both land in one push, deleting `TestGetTraceID` would drop coverage on a function that is live again. Worth keeping the test if the call site comes back.

**And a narrow residual, only if the block does not come back.** The handler wiring sits inside `if otelEnabled`. With OTEL off, no stats handler is installed, `SpanFromContext` returns a non-recording span, and `requestID` falls back to `uuid.New()`. Before this PR, `GetTraceID(traceparent, requestID)` ran regardless of whether tracing was enabled, so klog lines carried the caller's trace id even in a deployment with no collector. That is the only piece of the original concern I think survives, and your point 3 removes it entirely, so I raise it just so the two are known to be the same fix rather than separate ones.

Nothing else from me; the four-span split and the naming still read well.

rayne-Li:
really appreciate your thorough review.  @ErenAta16 

for point 3 and 4,  you are right, to keep logic with `prefer use traceID from rootSpan over traceparent in Header` we make those change an add some unitests, then when span is noopSpan, requestID will be traceID from `traceparent` header

```
		case HeaderTraceParent: // Preserve the trace context for requests initiated by the gateway plugin. like PD
			reqHeaders[n.Key] = string(n.RawValue)
			if !rootSpan.SpanContext().HasTraceID() { // prefers rootSpan traceID over traceparent
				requestID = GetTraceID(string(n.RawValue), requestID)
			}
		}
```

ErenAta16 (commented):
That is exactly the shape I was describing, and reading it back it is better than what I suggested.

```go
case HeaderTraceParent: // Preserve the trace context for requests initiated by the gateway plugin. like PD
    reqHeaders[n.Key] = string(n.RawValue)
    if !rootSpan.SpanContext().HasTraceID() { // prefers rootSpan traceID over traceparent
        requestID = GetTraceID(string(n.RawValue), requestID)
    }
```

`HasTraceID()` is the right predicate rather than checking whether OTEL is enabled anywhere, because it answers the only question that matters at that point: does this context actually carry a trace id. With the stats handler installed it does, and the span wins; with OTEL off the span is non-recording, `HasTraceID()` is false, and the header supplies it. So the `klog` correlation I was worried about survives a deployment with no collector, and there is no second source of truth to keep in sync.

Preferring the span over the header when both exist is also the correct precedence and not the obvious one. The span context is what the rest of the trace is keyed by, so taking the header instead would produce logs keyed differently from the spans for the same request.

Restoring the case also keeps `traceparent` in `reqHeaders` for the PD path, which was the other half of what the removal took with it, and it means `GetTraceID` stays referenced so its test still earns its place.

Nothing further from me. Thanks for working through four separate points rather than just the two I got right.

varungup90 (approved):

pkg/plugins/gateway/gateway.go
```diff
@@ -364,20 +378,24 @@ func (s *Server) handleProcessingRequest(st *processState, req *extProcPb.Proces
 
 	switch req.Request.(type) {
 	case *extProcPb.ProcessingRequest_RequestHeaders:
-		st.ctx, st.span = tracer.Start(st.ctx, "ExtProc_HTTP_Request")
-
-		resp, st.user, st.rpm, st.routerCtx = s.HandleRequestHeaders(st.ctx, st.requestID, req)
+		resp, st.user, st.rpm, st.routerCtx = s.HandleRequestHeaders(st.ctx, st.requestID, st.rootSpan, req)
 		if st.routerCtx != nil {
 			st.model = st.routerCtx.Model
-			st.requestID = st.routerCtx.RequestID
```
Jeffwan, line 372: why do you delete this line? is this a problem?  I will merge this PR first but please check whether that's needed. If so, please file a follow up PR

Jeffwan (approved):

Jeffwan:
seems no obvious changes requested, let's merge into main.

pkg/plugins/gateway/gateway.go
rayne-Li: we should add it back and i create a new PR to handle this, https://github.com/vllm-project/aibrix/pull/2527

PR #2525, opened 2026-08-04
fix(test): use dynamic port in pd tests to avoid cross-package collisions

Jonglee999:
## Summary

Eliminate a flaky `address already in use` failure by switching `setupTestServer` in `pkg/plugins/gateway/algorithms` from a hardcoded `127.0.0.1:8000` to an OS-assigned free port, so it no longer collides with `pkg/controller/modeladapter`'s mock servers under `go test ./...` parallelism.

## Problem

`make test` (`go test ./...`) runs package test binaries in parallel. Two packages both bind `127.0.0.1:8000`:

- `pkg/plugins/gateway/algorithms/pd_disaggregation_test.go` — `setupTestServer`
- `pkg/controller/modeladapter/lora_client_test.go` — `mockServer` (ports are production constants `DefaultInferenceEnginePort`/`DefaultRuntimeAPIPort`, consumed by `BuildURLs`, so they cannot change)

When the binaries run concurrently, whichever binds second fails with `listen tcp 127.0.0.1:8000: bind: address already in use`.

Reproduced deterministically: the two packages together fail 5/5 rounds; `-p 1` (sequential) never fails — confirming package-level parallelism is the trigger, not any environment factor.

## Solution

Switch `setupTestServer` to listen on `127.0.0.1:0` (OS-assigned free port) and return the port. Each caller stamps the port onto its prefill pod via the `constants.ModelLabelPort` label, which `GetModelPortForPod` already reads (defaulting to 8000 when absent). This **reuses the dynamic-port pattern already present elsewhere in the same file** (e.g. `TestPDRouter_RouteDoesNotEmitAsyncPrefillSuccessBeforeHTTPCompletes`).

Why fix the pd side, not modeladapter: the pd port comes from a pod **label** (configurable in tests), whereas the modeladapter port is a **production constant** hardcoded in `BuildURLs`. So the pd side is the natural place to make dynamic.

This is cross-platform — `127.0.0.1:0` works on Linux, macOS, and Windows (an earlier attempt using a `127.0.0.11` loopback alias was correctly flagged as macOS-incompatible, since macOS only configures `127.0.0.1` on `lo0` by default).

Only test code changes; no production code, no assertions on the modeladapter side.

## Testing

| Check | Before | After |
|-------|--------|-------|
| Two packages concurrent ×5 | 5/5 FAIL (1–2 port collisions/round) | 5/5 PASS, 0 collisions |
| `go test ./pkg/plugins/gateway/algorithms/` | flaky | ok |
| `make lint-all` | pass | pass |
| `make test` | intermittent FAIL | pass |
| `make test-race-condition` | pass | pass |

## Impact

- Backward compatible (test-only)
- Cross-platform (Linux/macOS/Windows)
- CI becomes more reliable for these two packages

### DIFF
--- a/pkg/plugins/gateway/algorithms/pd_disaggregation_test.go
+++ b/pkg/plugins/gateway/algorithms/pd_disaggregation_test.go
@@ -109,8 +109,19 @@ func TestPDRouter_Route(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			ts := setupTestServer(t, tt.serverCode, tt.serverResp, tt.llmEngine)
+			ts, prefillPort := setupTestServer(t, tt.serverCode, tt.serverResp, tt.llmEngine)
 			defer ts.Close()
+
+			// Stamp the dynamic port onto each prefill pod so GetModelPortForPod
+			// resolves it instead of falling back to the default 8000.
+			for _, pod := range tt.readyPods {
+				if pod.Labels[PDRoleIdentifier] == "prefill" {
+					if pod.Labels == nil {
+						pod.Labels = map[string]string{}
+					}
+					pod.Labels[constants.ModelLabelPort] = prefillPort
+				}
+			}
 
 			ctx := types.NewRoutingContext(context.Background(), "test", "model", "message", "test-request", "user")
 			ctx.Engine = tt.llmEngine
@@ -1112,7 +1123,7 @@ func TestDoPrefillRequest(t *testing.T) {
 
 	for _, tt := range tests {
 		t.Run(tt.name, func(t *testing.T) {
-			ts := setupTestServer(t, tt.serverCode, tt.serverResp, tt.llmEngine)
+			ts, prefillPort := setupTestServer(t, tt.serverCode, tt.serverResp, tt.llmEngine)
 			defer ts.Close()
 
 			prefillPods := []*v1.Pod{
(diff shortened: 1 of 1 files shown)


Jonglee999:
@gemini-code-assist Good catch — you're right that the loopback-alias approach (`127.0.0.11`) is macOS-incompatible, since macOS only configures `127.0.0.1` on `lo0` by default. I've dropped that approach entirely.

This update switches to your second suggestion: the conflicting test (`setupTestServer` in `pd_disaggregation_test.go`) now uses an OS-assigned free port (`127.0.0.1:0`), which is cross-platform.

A few notes on the chosen approach:
- `setupTestServer` now returns the port, and each caller stamps it onto the prefill pod's `constants.ModelLabelPort` label. This reuses the dynamic-port pattern already present elsewhere in the same file, and works because production code resolves the pod port via `GetModelPortForPod` (which reads that label, defaulting to 8000).
- I did **not** make `DefaultInferenceEnginePort`/`DefaultRuntimeAPIPort` configurable (your first suggestion). Those are production constants consumed by `BuildURLs`, and the modeladapter tests must listen on the real ports to exercise the real path. Since the pd port comes from a pod *label* (configurable in tests) rather than a hardcoded constant, changing the pd side alone fully resolves the collision without touching production code.

Verified: two packages concurrent ×5 → 5/5 PASS (0 collisions); `make lint-all && make test && make test-race-condition` all pass. Thanks for the review.

varungup90 (approved):

googs1025 (approved):
LGTM

PR #2530, opened 2026-08-05
[API] Batch: add adaptive AIMD ramp-up controls

scarlet25151:
## Pull Request Description
This PR makes the two batch-client AIMD ramp-up controls configurable and adapts their effective behavior without changing the surrounding smart-client design.

- expose `adaptive_healthy_window` and `adaptive_additive_increase` through per-job client config and process-level environment fallbacks
- use a bounded 25% proportional increase while initially probing capacity
- switch permanently to the configured additive step after the first overload- or latency-driven decrease
- preserve the existing max limit, overload detection, multiplicative decrease, backoff, routing, and dispatch behavior

## Testing
- `poetry run pytest ...` (194 passed)
- `poetry run ruff check ...`
- `poetry run ruff format --check ...`

## Related Issues
Resolves: N/A

### DIFF
--- a/docs/source/features/batch-model-deployment-templates.rst
+++ b/docs/source/features/batch-model-deployment-templates.rst
@@ -312,6 +312,8 @@ The MDS API accepts this AIBrix extension block:
          "max_concurrency": 256,
          "adaptive_concurrency": true,
          "adaptive_max_factor": 16,
+         "adaptive_healthy_window": 2,
+         "adaptive_additive_increase": 4,
          "retry_policy": {
            "max_retries": 5,
            "base_delay_seconds": 2,
@@ -323,13 +325,22 @@ The MDS API accepts this AIBrix extension block:
    }
 
 ``client`` controls per-job smart-client behavior. ``max_concurrency`` is an
-absolute job-global in-flight cap, hard limited to 256 (requests above are
+absolute job-global in-flight cap, hard limited to 1024 (requests above are
 rejected). If adaptive concurrency is enabled, the cap limits adaptive growth;
 if adaptive concurrency is disabled, it becomes the fixed concurrency. Omitted
 fields fall back to metadata-service environment
-defaults and then built-in defaults. This public block intentionally does not
-expose telemetry interval, QPS, request timeout, or fine-grained adaptive
-controller internals.
+defaults and then built-in defaults. ``adaptive_healthy_window`` is the number
+of healthy completions required for each growth step (default 8).
+During initial capacity probing, the controller increases by the greater of
+``adaptive_additive_increase`` and 25% of the current limit. After the first
+overload or latency-driven decrease, probing ends permanently and the
+controller uses ``adaptive_additive_increase`` as the fixed AIMD
+congestion-avoidance step (default 1). Smaller windows and larger increases
+ramp up faster but can overshoot backend capacity. Operators can set their
+process-wide fallbacks with
+``AIBRIX_BATCH_ADAPTIVE_HEALTHY_WINDOW`` and
+``AIBRIX_BATCH_ADAPTIVE_ADDITIVE_INCREASE``. Other fine-grained adaptive
+controller internals remain private.
 
 The known upstream runtime targets are:
 
--- a/apps/console/api/gen/console/v1/console.pb.go
+++ b/apps/console/api/gen/console/v1/console.pb.go
@@ -2209,13 +2209,15 @@ func (x *JobResourceRequest) GetProviderConfig() *structpb.Struct {
 // fields use proto3 presence so "unset" (fall back to env defaults) is
 // distinguishable from an explicit zero.
 type JobClientConfig struct {
-	state               protoimpl.MessageState `protogen:"open.v1"`
-	MaxConcurrency      *int32                 `protobuf:"varint,1,opt,name=max_concurrency,json=maxConcurrency,proto3,oneof" json:"max_concurrency,omitempty"`                // absolute in-flight cap, 1..1024
-	AdaptiveConcurrency *bool                  `protobuf:"varint,2,opt,name=adaptive_concurrency,json=adaptiveConcurrency,proto3,oneof" json:"adaptive_concurrency,omitempty"` // grow concurrency adaptively
-	AdaptiveMaxFactor   *float64               `protobuf:"fixed64,3,opt,name=adaptive_max_factor,json=adaptiveMaxFactor,proto3,oneof" json:"adaptive_max_factor,omitempty"`    // adaptive growth factor, >= 1
-	RetryPolicy         *JobClientRetryPolicy  `protobuf:"bytes,4,opt,name=retry_policy,json=retryPolicy,proto3" json:"retry_policy,omitempty"`
-	unknownFields       protoimpl.UnknownFields
-	sizeCache           protoimpl.SizeCache
+	state                    protoimpl.MessageState `protogen:"open.v1"`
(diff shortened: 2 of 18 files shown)


docs/source/features/batch-model-deployment-templates.rst
```diff
@@ -312,6 +312,8 @@ The MDS API accepts this AIBrix extension block:
          "max_concurrency": 256,
          "adaptive_concurrency": true,
          "adaptive_max_factor": 16,
+         "adaptive_healthy_window": 2,
```
Jeffwan, line 315: this is partial change? how do you configure it from backend console? seems this is not exposed yet? I am ok to merge this one first if that's on your plan

Jeffwan (approved):

Jeffwan:
Below is the AI review summary

---

Reviewed the change. The plumbing is clean and the tests track the new behavior well. Two things need a decision from you before this lands, and a few smaller items.

### 1. The growth window now scales with the current limit — this isn't in the PR description

`_growth_window_size()` returns `max(settings.healthy_window, self._limit)`, so the number of healthy completions required per increase is no longer a constant. The PR description lists three changes (expose two knobs, 25% proportional probe, switch to additive after first decrease) but not this one, and it is the change with the largest production impact.

Concretely, after a decrease to 115 with the default `additive_increase=1`:

- before: window 8, so ~104 healthy completions to climb back to 128
- after: window 115, so ~1495 healthy completions to climb back to 128

That is roughly a 14x slowdown in recovery. The renamed test makes this explicit — `test_llm_controller_recovers_quickly_after_small_decrease` became `test_llm_controller_uses_additive_increase_after_overload`, and the assertion flipped from "recovers to 128" to "stays at 115". For a batch job of a few hundred to a few thousand requests, one transient decrease now means the job never recovers its throughput.

Textbook AIMD does increase by 1 per RTT, and one RTT is roughly cwnd acks, so `window == limit` is the faithful translation. But TCP congestion avoidance is tuned for continuous flows fairly sharing a link. A batch job against a dedicated deployment has no competing flow to be fair to, and it has a finite request budget — aggressive recovery is the better tradeoff there. Could you confirm whether the slowdown is intentional? Either way it should be called out in the PR description and explained in the docs, because a reader of the current docs ("minimum number of healthy completions in a growth window") will not infer that the effective window equals the current limit.

### 2. `_probing` is a one-way door that never resets

`_decrease()` sets `self._probing = False` permanently. Any single decrease — for the rest of the job — costs you the proportional probe and drops you into `+additive_increase` per `limit` completions.

Two ways this fires earlier than you would want:

- **Startup.** `_is_capacity_error()` treats `retryable is not False` as a capacity error, which covers `NO_ENDPOINT`. While a per-job Deployment is still warming up, it is easy to accumulate the 2 errors / 10% rate that trips the overload window, before any real traffic has flowed.
- **Latency outliers.** `_is_slow()` decreases on a *single* outcome when TTFT exceeds `relative_slowdown_factor` (1.75) x EWMA. This path is mostly dormant today because vLLM does not surface TTFT in the OpenAI response body, so `ttft_seconds` is usually `None` — but any backend that does report it will trip this on normal prompt-length variance, not on real overload.

Combined with (1), a job that hits one early hiccup is pinned at a low limit for its entire lifetime. Worth giving probing a way back: re-enter when the limit falls to (or near) `min_limit`, or after K consecutive windows with no decrease, or exempt a warm-up period from ending the probe.

### 3. `adaptive_healthy_window` is inert as a per-job knob in most cases

Because the effective window is `max(healthy_window, limit)`, this field has no effect at all once `limit > healthy_window`. The new test demonstrates it: `test_llm_controller_adapts_probe_increase_to_current_limit` sets `healthy_window=2` with `initial_limit=4`, and the effective window is 4, not 2. So the documented benefit — "Lower minimum windows speed up low-concurrency probing" — only holds while the limit is under 8.

Exposing a public per-job API field that silently stops mattering above single-digit concurrency is a bit of a trap. Either document the exact range where it applies, or reconsider whether this is the right knob to expose.

### 4. The knob that actually drives the ramp is hardcoded

`_PROBE_INCREASE_RATIO = 0.25` is the dominant term during probing: `increase = max(additive_increase, ceil(limit * 0.25))`, so `additive_increase` only wins while `limit < 4 * additive_increase`. For a PR titled "add adaptive AIMD ramp-up controls", the ramp-up rate during the probe phase is the one thing not controllable. Fine to keep it private, but please spell out the interaction in the `LLMAdaptiveConcurrencySettings` docstring so users are not surprised that `additive_increase` appears to do nothing during ramp-up.

### 5. Duplicated window logic

`_growth_window_size()` is the same expression already inlined in `_observe_overload_window()`:

```python
self._overload_window_size = max(
    self._settings.healthy_window,
    self._limit,
)
```

Two copies of the same policy will drift. Have `_observe_overload_window()` call `_growth_window_size()`.

### 6. The Console contract was not updated, so these fields are unreachable from Console

`AibrixExtension` says the authoritative shape lives in `console.proto` and to "keep these two in sync", but neither side was updated:

- `apps/console/api/proto/console/v1/console.proto` — `JobClientConfig` still has only fields 1-3 plus `retry_policy`
- `apps/console/api/planner/api/jobs.go` — the `ClientConfig` struct is missing both fields

Since the planner projects that struct into `extra_body.aibrix.client`, a Console-submitted job silently drops both knobs. Only a direct MDS `POST /v1/batches` can set them today. The `extra_body` example in the `AibrixExtension` docstring in `metadata/api/v1/batch.py` was also not updated.

### 7. Pre-existing doc inconsistency worth fixing here

The rst says `max_concurrency` is "hard limited to 256", but `MAX_CLIENT_CONCURRENCY = 1024` and the proto comment says `1..1024`. The proto also contradicts itself — the field comment says 1..1024 while the prose block below says 256. Since you are already touching this paragraph, good chance to make all three agree.

### 8. Test gaps

- The `max(..., 1)` clamp in `_adaptive_healthy_window()` / `_adaptive_additive_increase()` is untested — nothing covers `AIBRIX_BATCH_ADAPTIVE_HEALTHY_WINDOW=0` or a non-integer value.
- Precedence is tested only in isolation (env-only in one test, client-only in another). Worth one test with both set, asserting the per-job value wins.

### Context on multi-tenancy

For what it's worth, I traced the per-job isolation and it holds: `create_job_driver()` is per job, `DispatchEngine` is constructed per job in `_execute_job_in_runtime`, `run()` keeps all concurrency state in locals, and `concurrency.py` has no module-level mutable state. So two concurrent jobs genuinely get independent controllers with independent settings.

What is *not* isolated is the thing being controlled. In the injected-endpoint-source path, `_endpoint_source` is a process singleton, `GatewayEndpointSource` holds a single `HttpChannel`, and that channel lazily creates a single `httpx.AsyncClient`. So all jobs share one connection pool, and each job computes its own `max_limit` from the same advertised `capacity` times its own `adaptive_max_factor` — N jobs at factor 16 is N x capacity x 16 with no global ceiling. Two consequences: pool queueing shows up to the controller as latency rather than as overload, which is a misleading signal; and making ramp-up per-job tunable also makes the amplification per-job tunable. Not introduced by this PR, but this PR increases the exposure, so a process-level global cap may be worth a follow-up.

Jeffwan:
Let's merge this one

```
{
  "input_dataset": "file-xxx",
  "endpoint": "/v1/chat/completions",
  "completion_window": "24h",
  "name": "adaptive-batch-job",
  "model_id": "your-model",
  "model_template_name": "your-template",
  "resource_request": {
    "replicas": 1
  },
  "client": {
    "adaptive_concurrency": true,
    "max_concurrency": 256,
    "adaptive_max_factor": 16,
    "adaptive_healthy_window": 8,
    "adaptive_additive_increase": 2
  }
}
```

Since you mainly use the API, I think you can directly try it first, no need to implement it in the UI

PR #2544, opened 2026-08-11
feat(gateway-plugin): add lockedRoutingStrategy to pin per-model routing strategy

Yang1032:
## Summary

Add `lockedRoutingStrategy` to the `model.aibrix.ai/config` annotation so admins can pin a routing strategy model-wide and prevent clients from overriding it via the `routing-strategy` request header.

### New resolution priority (highest → lowest)

1. `lockedRoutingStrategy` (top-level) — always wins when set
2. `routing-strategy` request header
3. `routingStrategy` from the resolved config profile
4. `ROUTING_ALGORITHM` environment variable

Behavior is unchanged when `lockedRoutingStrategy` is unset.

### Usage

```yaml
annotations:
  model.aibrix.ai/config: |
    {
      "lockedRoutingStrategy": "pd",
      "defaultProfile": "default",
      "profiles": {
        "default": { "routingStrategy": "pd" },
        "batch":   { "routingStrategy": "random", "requestsPerSecond": 50 }
      }
    }
```

Profile selection via `config-profile` stays alive — the lock only pins `routingStrategy`. The `batch` profile above can still tune `requestsPerSecond`; its `routingStrategy` is simply ignored while the lock is set.

### Use cases

- **Pin PD for disaggregated deployments** — a model served with separate prefill/decode pools must always go through `pd`. Clients sending `routing-strategy: random` can no longer break the prefill/decode split.
- **Per-model strategy on mixed clusters** — when a cluster mixes PD and non-PD models, each model pins its own strategy via its own annotation, independently of the global `ROUTING_ALGORITHM`.
- **Predictable routing for SLO/audit** — latency-sensitive models can be locked to `slo` / `least-latency` / `prefix-cache` so routing stays auditable.

## Changes

- `pkg/plugins/gateway/configprofiles/configprofiles.go`: add `LockedRoutingStrategy` to `ModelConfigProfiles`; split `ResolveProfile` into `ResolveConfig` (returns profile + locked strategy) and keep `ResolveProfileFromPod` for callers that only need the profile.
- `pkg/types/router_context.go`: add `LockedRoutingStrategy` to `ResolvedConfigProfile`.
- `pkg/plugins/gateway/util.go`: propagate the locked strategy in `applyConfigProfile`; check it first in `deriveRoutingStrategyFromContext`.
- Tests: unit tests for `ResolveConfig` and `deriveRoutingStrategyFromContext`, plus four integration tests covering lock-wins-over-header, lock-without-resolvable-profile, invalid-lock-returns-400, and a regression check that the header still wins when no lock is set.
- Docs: `gateway-plugins.rst` (field reference, priority table, example) and `pd-disaggregation.rst` (how to pin `pd`).

## Non-goals / Limitations

- **Backend compatibility is not validated.** The lock bypasses the per-request override, so the admin must ensure the pinned strategy matches the deployment: `pd` requires disaggregated prefill/decode pools; cache-aware strategies depend on the engine exposing prefix-cache state. Forcing an unsupported strategy will fail or degrade routing.
- **An invalid `lockedRoutingStrategy` returns 400 for every request** rather than silently falling back. This is intentional (fail loud over silently routing somewhere unintended), but a typo in the annotation makes the model unavailable until corrected.
- **Models without the annotation are unaffected.** Existing deployments need no migration.
- **`pd_disaggregation.go` keeps using `ResolveProfileFromPod`** (not `ResolveConfig`) because PD routing-config resolution happens inside the algorithm, after the locked strategy has already been enforced upstream in `deriveRoutingStrategyFromContext`.

### DIFF
--- a/docs/source/features/gateway-plugins.rst
+++ b/docs/source/features/gateway-plugins.rst
@@ -542,9 +542,35 @@ Add the ``model.aibrix.ai/config`` annotation to the pod template of your ``Depl
           }
         }
 
+Optionally pin a single routing strategy model-wide so it cannot be overridden at
+request time (see Routing strategy priority below). ``lockedRoutingStrategy`` only
+locks the strategy: the remaining per-profile knobs (``requestsPerSecond``,
+``routingConfig``) still follow the profile selected by ``config-profile``:
+
+.. code-block:: yaml
+
+    annotations:
+      model.aibrix.ai/config: |
+        {
+          "lockedRoutingStrategy": "pd",
+          "defaultProfile": "default",
+          "profiles": {
+            "default": {
+              "routingStrategy": "pd"
+            },
+            "batch": {
+              "routingStrategy": "random",
+              "requestsPerSecond": 50
+            }
+          }
+        }
+
 **Selecting a profile at request time**
 
-Add the ``config-profile`` header. If the header is absent, the ``defaultProfile`` is used:
+Two request headers drive the config at request time:
+
+* ``config-profile`` selects a named profile; when absent, ``defaultProfile`` (or ``"default"``) is used.
+* ``routing-strategy`` overrides the selected profile's ``routingStrategy``, unless ``lockedRoutingStrategy`` is set (see Routing strategy priority below).
 
 .. code-block:: bash
 
@@ -553,6 +579,21 @@ Add the ``config-profile`` header. If the header is absent, the ``defaultProfile
       -H "config-profile: batch" \
       -H "Content-Type: application/json" \
       -d '{"model": "my-model", "messages": [{"role": "user", "content": "Summarize..."}]}'
+
+**Top-level fields**
+
+.. list-table::
+   :header-rows: 1
+   :widths: 30 70
+
+   * - Field
+     - Description
+   * - ``lockedRoutingStrategy``
+     - Pins a single routing strategy model-wide. When set, it takes precedence over the ``routing-strategy`` header, the per-profile ``routingStrategy`` and the ``ROUTING_ALGORITHM`` env. The remaining per-profile knobs (``requestsPerSecond``, ``routingConfig``) are still applied normally.
+   * - ``defaultProfile``
+     - Profile name used when no ``config-profile`` header is sent. Falls back to ``"default"`` when omitted.
+   * - ``profiles``
+     - Map of named profiles, each carrying the per-profile fields below.
 
 **Profile fields**
 
@@ -571,11 +612,12 @@ Add the ``config-profile`` header. If the header is absent, the ``defaultProfile
 
 **Routing strategy priority** (highest to lowest):
 
-1. ``routing-strategy`` request header — always wins, even if a profile is active.
-2. ``routingStrategy`` from the resolved config profile.
-3. ``ROUTING_ALGORITHM`` environment variable on the gateway plugin.
+1. ``lockedRoutingStrategy`` pinned model-wide in the config — always wins when set, even over the ``routing-strategy`` header.
+2. ``routing-strategy`` request header.
+3. ``routingStrategy`` from the resolved config profile.
+4. ``ROUTING_ALGORITHM`` environment variable on the gateway plugin.
 
-**Backward compatibility**: if a pod has no ``model.aibrix.ai/config`` annotation, the gateway falls back to the ``model.aibrix.ai/routing-strategy`` pod label and the ``ROUTING_ALGORITHM`` env. No migration is required for existing deployments.
+**Backward compatibility**: if a pod has no ``model.aibrix.ai/config`` annotation, the gateway falls back to the ``routing-strategy`` request header and then the ``ROUTING_ALGORITHM`` env (steps 2 and 4 above). No migration is required for existing deployments.
 
 .. _prometheus-api-access:
 
--- a/docs/source/features/pd-disaggregation.rst
+++ b/docs/source/features/pd-disaggregation.rst
@@ -184,6 +184,21 @@ To make ``pd`` the default for a model, add the config annotation to the pod tem
           }
         }
 
+To prevent clients from overriding ``pd`` with a ``routing-strategy`` header (e.g.
+``random``), pin it model-wide with ``lockedRoutingStrategy``:
+
+.. code-block:: yaml
+
+    annotations:
+      model.aibrix.ai/config: |
+        {
+          "lockedRoutingStrategy": "pd",
+          "defaultProfile": "default",
+          "profiles": {
+            "default": { "routingStrategy": "pd" }
+          }
+        }
+
 
 Step 3 — Add Standard Inference Pods (Optional)
 -------------------------------------------------
--- a/pkg/plugins/gateway/configprofiles/configprofiles.go
+++ b/pkg/plugins/gateway/configprofiles/configprofiles.go
@@ -44,8 +44,12 @@ type ModelConfigProfile struct {
 
 // ModelConfigProfiles is the root JSON structure from model.aibrix.ai/config.
 type ModelConfigProfiles struct {
-	DefaultProfile string                        `json:"defaultProfile"`
-	Profiles       map[string]ModelConfigProfile `json:"profiles"`
+	// LockedRoutingStrategy, when set, pins the routing strategy model-wide.
+	// It takes precedence over the routing-strategy request header, the per-profile
+	// routingStrategy and the ROUTING_ALGORITHM environment variable.
+	LockedRoutingStrategy string                        `json:"lockedRoutingStrategy,omitempty"`
+	DefaultProfile        string                        `json:"defaultProfile"`
+	Profiles              map[string]ModelConfigProfile `json:"profiles"`
 }
 
 // GetProfile returns the profile for the given name, or the default profile.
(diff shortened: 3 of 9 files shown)


Yang1032:
@varungup90  @Jeffwan  could you take a look when you have a moment?

Yang1032:
Hi @googs1025 , thanks for the Copilot light review — both comments are addressed
Could you take another look when you have a moment? Thanks!

googs1025:
Thanks for adding this. The precedence change is clear, well documented, and covered by both helper-level and request-path tests. LGTM.

googs1025 (approved):

varungup90:
### Review Comments

* **Security-design (confirmed) — `configprofiles.go:54**`
`RoutingForced` is enforced per-profile, not per-model. If a model has both a forced profile and a second, non-forced profile, a client can send `config-profile: <other-name>` and fully regain header-based routing control. This is disclosed in the PR's docs and covered by a test (`TestHandleRequestBody_ConfigProfileBypassesForcedProfile`), so it's intentional. However, there's no guard in `ParseModelConfig` warning an admin who mixes forced/non-forced profiles for the same model, making the protection easy to silently defeat by future config changes. Worth confirming this is the desired security posture.
* **Simplification (confirmed) — `util.go:671**`
The nil-guard + `TrimSpace(routingCtx.ConfigProfile.RoutingStrategy)` sequence is duplicated between the new forced check and the existing fallback check a few lines later. This is low risk today, but the two copies could drift if a third profile-derived field is added later.
* **Style (minor) — `configprofiles.go:63` and `router_context.go**`
Struct field alignment for the new fields looks like it hasn't been formatted. It's worth doing a quick `gofmt -l`/`gofmt -w` pass before merging.

Yang1032:
@varungup90  — thanks for the review; this is the design question I wanted to surface, and I'd appreciate your take on the right posture.

The choice: should the pin be scoped per-profile or per-model?

Option 1 — keep per-profile `routingForced `v(current PR, the pr's name should update).
- Semantics: when a forced profile is active, the client's routing-strategy header is ignored.
- Pros: granular (different profiles can have different policies) and backward-compatible.
- Known gap: if any profile is non-forced, a client can send config-profile to select it and regain the header override.
- Mitigation at the config-authoring layer: validate at pod add/update — the informer hook, i.e. the moment a config is loaded into the cache. Warn when (a) a config mixes forced and non-forced profiles, and (b) pods of the same model carry different annotations. The effective profile is currently taken from whichever pod is first in the list, and that order isn't deterministic, so a silent divergence can even flip the effective config between requests. Pod events are low-frequency, so this stays off the request hot path.
- Documented posture: for a locked strategy, define only forced profiles.

Option 2 — move the pin to model level (`forced: true` at the model/`defaultProfile` level).
- Stronger: the strategy is admin-locked regardless of profile — no per-profile loophole, simpler to reason about.
- Costs: schema change, loss of per-profile granularity, and it forces a semantic decision: should config-profile also be ignored for a forced model (true pin), or only the routing-strategy header (profile selection can still change the strategy)?

My leaning is Option 1 + validation at pod add/update: it keeps the feature granular and backward-compatible while surfacing the "silently bypassable" gap the moment the config is written or changed, rather than when it's exploited. 

What's your take — does one of these match how you'd want routing lock-in to behave?

varungup90:
> Option 2 — move the pin to model level (`forced: true` at the model/`defaultProfile` level).

Thanks for laying out the tradeoff so clearly. I'd go with model-level, but shaped a bit differently than `"forced: true` at the model/`defaultProfile` level" — lock the *field*, not the profile object:

```json
{
  "defaultProfile": "default",
  "lockedRoutingStrategy": "pd",
  "profiles": {
    "default": {"routingStrategy": "pd"},
    "burst":   {"routingStrategy": "random", "requestsPerSecond": 50}
  }
}

```

`lockedRoutingStrategy` sits outside profiles, so it directly resolves the "should `config-profile` also be ignored" question: profile selection stays alive for the knobs that are legitimately per-profile (`requestsPerSecond`, `routingConfig`) — the `burst` profile above can still tune RPS — but `routingStrategy` resolution short-circuits to the locked value regardless of which profile is active, before the header or the profile's own `routingStrategy` is even consulted.

New priority becomes:

1. `lockedRoutingStrategy` (top-level, if set)
2. `routing-strategy` header
3. Profile `routingStrategy`
4. `ROUTING_ALGORITHM` env

This also lets `routingForced` on individual profiles go away rather than living alongside a second force mechanism in the same struct — one precedence rule instead of two overlapping ones.

Yang1032:
> `lockedRoutingStrategy` sits outside profiles, so it directly resolves the "should `config-profile` also be ignored" question: profile selection stays alive for the knobs that are legitimately per-profile (`requestsPerSecond`, `routingConfig`) — the `burst` profile above can still tune RPS — but `routingStrategy` resolution short-circuits to the locked value regardless of which profile is active, before the header or the profile's own `routingStrategy` is even consulted.
> 
> New priority becomes:
> 
> 1. `lockedRoutingStrategy` (top-level, if set)
> 2. `routing-strategy` header
> 3. Profile `routingStrategy`
> 4. `ROUTING_ALGORITHM` env
> 
> This also lets `routingForced` on individual profiles go away rather than living alongside a second force mechanism in the same struct — one precedence rule instead of two overlapping ones.

@varungup90 — agree, this is cleaner. A top-level lockedRoutingStrategy resolves the question I was stuck on: the strategy short-circuits to the locked value before the header or the profile is even consulted, while profile selection keeps governing the genuinely per-profile knobs (requestsPerSecond, routingConfig). And dropping per-profile routingForced leaves exactly one precedence rule instead of two overlapping ones — that's a better design.

Very happy to implement this — I'll update the PR and follow up.

varungup90:
### Nit Review Comments

#### 1. REUSE — New test helper duplicates existing inline pattern

**File:** `pkg/plugins/gateway/gateway_req_body_test.go` (near lines ~882–890)

The PR adds `registerTestRouter(mockRouter)` to DRY up router registration for its four new tests. However, the existing `TestHandleRequestBody_ModelRPSNotConsumedOnRoutingFailure` (around lines 799–810) still contains the same inline setup:

```go
mockRouterProvider := func() (types.Router, error) { ... }
routingalgorithms.Register(...)
routingalgorithms.Init()
```

This leaves the file with two implementations of the same setup logic: the new helper and the original inline version.

**Suggestion:** Reuse `registerTestRouter()` in the existing test as well so there is a single registration pattern throughout the file.

**Failure scenario:** If router registration logic changes in the future (for example, adding cleanup or deduplication), both implementations must be updated independently, increasing maintenance cost and the chance of inconsistencies.

---

#### 2. SIMPLIFICATION — Four near-identical tests could be table-driven

**File:** `pkg/plugins/gateway/gateway_req_body_test.go` (lines ~960–1105)

The following tests share nearly identical setup:

- `TestHandleRequestBody_LockedRoutingStrategyOverridesHeader`
- `TestHandleRequestBody_WithoutResolvableProfile`
- `TestHandleRequestBody_InvalidReturns400`
- `TestHandleRequestBody_HeaderWinsWithoutLockedStrategy`

Each repeats roughly 35–40 lines of identical scaffolding:

- `mockCache` / `mockRouter` setup
- `registerTestRouter`
- request creation
- `routingCtx` construction
- `mockCache.On(...)`
- final assertions

The only differences are:

- `profileJSON`
- request header value
- expected routing strategy
- expected HTTP status

**Suggestion:** Convert these into a single table-driven test, for example:

```go
[]struct {
    name
    profileJSON
    headerValue
    wantStrategy
    wantStatus
}
```

This would reduce roughly 150 lines of duplicated test code to a single reusable test loop.

**Failure scenario:** Any future change to `HandleRequestBody`, mock expectations, or header names must currently be applied to four nearly identical tests, making it easy for one copy to drift.

---

#### 3. EFFICIENCY — `LockedRoutingStrategy` is trimmed twice

**Files:**

- `pkg/plugins/gateway/configprofiles/configprofiles.go`
- `pkg/plugins/gateway/util.go`

`ResolveConfig()` already trims the locked routing strategy before returning it:

```go
return cfg.GetProfile(headerProfile), strings.TrimSpace(cfg.LockedRoutingStrategy)
```

`applyConfigProfile()` stores this already-trimmed value into:

```go
routingCtx.ConfigProfile.LockedRoutingStrategy
```

Later, `deriveRoutingStrategyFromContext()` trims the same value again before using it.

This differs from the sibling `RoutingStrategy` field, which is stored untrimmed and normalized only once at read time.

**Suggestion:** Choose a single ownership model for normalization (normalize on write or normalize on read) and keep both routing strategy fields consistent.

**Failure scenario:** Every request using a config profile performs an unnecessary second `TrimSpace()`. While inexpensive, it is redundant work on a hot request path, and the inconsistent normalization behavior makes future maintenance less predictable.

Yang1032:
Thanks for the review @varungup90 — all three nits are addressed:

1. **REUSE** — `TestHandleRequestBody_ModelRPSNotConsumedOnRoutingFailure` now calls the shared `registerTestRouter(mockRouter)` helper instead of its inline registration, so there's a single registration pattern in the file.
2. **SIMPLIFICATION** — the four tests are collapsed into one table-driven `TestHandleRequestBody_LockedRoutingStrategy` using `[]struct{ name, profileJSON, headerValue, wantStrategy, wantStatus }` plus a `t.Run` loop.
3. **EFFICIENCY** — removed the double trim. I kept the "normalize-on-read" model: `ResolveConfig` now returns the raw `LockedRoutingStrategy` (consistent with the sibling `RoutingStrategy` field), and `deriveRoutingStrategyFromContext` trims it once at read time.

Could you take another look when you have a moment? Thanks!

varungup90 (approved):

abe-servescale:
@varungup90 @googs1025 @Jeffwan can I ask if this PR will be scheduled for being part of aibrix v0.8.0? I think that it is really useful! Thanks in advance for you time!

varungup90:
@abe-servescale yes, it will be part of 0.8.0 release.

PR #2576, opened 2026-08-17
feat(brixbench): Add brixbench routing topology comparison matrix

JinKim48:
## Pull Request Description

This PR expands brixbench routing benchmarks to support consistent comparisons across AIBrix, Dynamo, and llm-d.

### Changes

- Add Qwen3-8B routing scenarios for:
  - 4p4d single-node and multi-node
  - 4p8d multi-node
  - 8p4d multi-node
  - 8p8d multi-node
- Standardize workloads at request rates 8 and 16 with 1,000 prefix-repetition prompts.
- Add framework-specific and combined comparison scenarios.
- Add AIBrix, Dynamo, and llm-d deployment fixtures for the supported topologies.
- Improve Dynamo RDMA reliability by detecting each Pod's UCX device and applying the validated UCX transport configuration.
- Extend aggregate CSV output with topology-aware labels, run dates, and prefix-repetition workload fields.
- Replace the native TOS client with the S3-compatible API to support public and internal TOS endpoints.
- Update brixbench documentation for the new scenarios, provenance handling, and publishing configuration.

### Testing

- All benchmark package unit tests
- End-to-end AIBrix Qwen3-8B 4p4d single-node scenarios at rates 8 and 16
- End-to-end AIBrix Qwen3-8B 4p4d multi-node scenarios at rates 8 and 16

The larger 4p8d, 8p4d, and 8p8d fixtures are covered by topology and configuration tests but were not all executed end to end.


brixbench/internal/deployers/llmd.go
```diff
@@ -281,17 +273,20 @@ func (d *LLMdDeployer) ensureLLMdImagePullSecret(ctx context.Context) (bool, err
...
 	}
-	output, err := d.captureLLMdCommand(ctx, "validate-llmd-release-tag", "git", "-C", repoPath, "rev-parse", d.version+"^{tag}")
+	output, err := d.captureLLMdCommand(ctx, "validate-llmd-release-tag", "git", "ls-remote", "--tags", "--refs", llmdRepoURL, "refs/tags/"+version)
 	if err != nil {
-		return fmt.Errorf("failed to validate LLM-d release tag %s in %s: %w", d.version, repoPath, err)
+		return fmt.Errorf("failed to query LLM-d release tag %s in %s: %w", version, llmdRepoURL, err)
 	}
-	if strings.TrimSpace(output) == "" {
-		return fmt.Errorf("LLM-d release tag %s was not found in %s", d.version, repoPath)
+	if !lsRemoteOutputHasExactTag(output, version) {
```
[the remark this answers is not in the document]
JinKim48: This is a false positive.

`lsRemoteOutputHasExactTag` already lives in the same `deployers` package
(`brixbench/internal/deployers/dynamo_release.go`) and is shared with the
Dynamo release-tag check. `validateLLMdReleaseTag` compiles and
`go test ./internal/deployers` passes on this branch.

Keeping the helper (instead of inlining `strings.Contains`) avoids accepting
prefix matches like `v1.2.10-rc0` when looking for `v1.2.1` — see the existing
unit coverage in `dynamo_test.go`.

brixbench/benchmark/tos_go_uploader.go
```diff
@@ -177,26 +206,32 @@ func (u *goTOSUploader) AppendBytes(remoteURI string, data []byte) error {
...
+		defer out.Body.Close()
+		existing, readErr := io.ReadAll(out.Body)
+		if readErr != nil {
+			return fmt.Errorf("tos get body %s: %w", remoteURI, readErr)
 		}
-		input.Offset = head.ContentLength
-		input.PreHashCrc64ecma = head.HashCrc64ecma
-	} else if !isTOSNotFound(headErr) {
-		return fmt.Errorf("tos head %s: %w", remoteURI, headErr)
+		body = append(existing, data...)
```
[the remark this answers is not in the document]
JinKim48: Addressed. Follow-up commit `d8c49360` closes the GetObject body
immediately after `io.ReadAll`, before PutObject.

brixbench/benchmark/aggregate_csv_test.go
```diff
@@ -46,6 +47,7 @@ var aggregateCSVHeader = []string{
 	"category",
 	"status",
 	"started_at",
+	"run_date",
```
whalepark, line 50: Was this schema change accounted for in the dashboard CSV parser? Existing benchmark_metrics.csv objects retain the old header, while newly appended rows insert run_date after started_at. A header-based parser could therefore map run_date to finished_at and shift every subsequent column. Does the dashboard explicitly support both layouts, or is the aggregate object recreated during rollout?

brixbench/benchmark/tos_go_uploader.go
```diff
@@ -177,26 +206,33 @@ func (u *goTOSUploader) AppendBytes(remoteURI string, data []byte) error {
...
+	} else if isTOSNotFound(getErr) {
+		body = data
+	} else {
+		return fmt.Errorf("tos get %s: %w", remoteURI, getErr)
 	}
 
-	_, err = u.client.AppendObjectV2(ctx, input)
+	_, err = u.client.PutObject(ctx, &s3.PutObjectInput{
```
whalepark, line 228: This keeps the implementation simple, and contention should be low for current scenarios. Still, concurrent publishers could overwrite rows, so it would be good to add ETag-based CAS and retries in a follow-up.

whalepark (commented):
Looks good to me if `run_date` is not messing up csv header. The other comment is about some known issues about data race.

brixbench/benchmark/aggregate_csv_test.go
JinKim48: Good catch, that risk is real for this append path.

If an older aggregate CSV still had the pre-`run_date` header and we appended the newer wider rows, a header-based parser would shift columns from `finished_at` onward.

The live TOS aggregate is already migrated to the current schema (`run_date` included, row widths consistent), and CI on this branch is appending against that object, so we’re fine for this PR.

I’ll harden this in a follow-up PR so schema drift is handled safely.

brixbench/benchmark/tos_go_uploader.go
JinKim48: Agreed. Get+concat+Put is intentionally simple and is not concurrency-safe: two publishers can race and drop rows. Current usage is effectively single-writer / sequential (one CI or one interactive suite at a time), so we’re accepting that for now.

I’ll track ETag/If-Match CAS with retries as a follow-up rather than blocking this PR on it.

whalepark (commented):
Few concerns, but after being clarified looks good to me.

Jeffwan (approved):

PR #2586, opened 2026-08-20
feat(chart): add configurable gateway TLS support

NelZyhh:
## Summary
- add configurable Gateway `http` / `tls` mode selection in the Helm chart
- support HTTPS listeners with optional `hostname`, `secretName`, and self-signed certificate generation
- add value validation and schema updates for TLS settings, plus regression tests for hostname and self-signed Secret rendering

## Details
- update `dist/chart/templates/gateway-instance/gateway.yaml` to render either HTTP or HTTPS listeners
- add `dist/chart/templates/gateway-instance/tls_secret.yaml` to generate and reuse a self-signed TLS Secret when `gateway.tls.autoGenerate=true`
- add validation in `dist/chart/templates/_helpers.tpl` to enforce HTTP/TLS exclusivity and require `gateway.tls.hostname` when auto-generating certs
- update `dist/chart/values.yaml`, `dist/chart/stable.yaml`, `dist/chart/vke.yaml`, and `dist/chart/values.schema.json` with the new TLS fields
- add Helm unit tests for TLS hostname rendering and self-signed Secret rendering

## Test Plan
- [x] `helm lint dist/chart`
- [x] `helm template aibrix dist/chart --namespace aibrix-system --show-only templates/gateway-instance/tls_secret.yaml --show-only templates/gateway-instance/gateway.yaml --set gateway.enable=true --set gateway.envoyAsSideCar=false --set gateway.http.enabled=false --set gateway.tls.enabled=true --set gateway.tls.autoGenerate=true --set gateway.tls.hostname='*.ctmaas.cn' --set gateway.tls.secretName='aibrix-gateway-tls'`
- [ ] `helm unittest dist/chart` (`helm-unittest` plugin is not installed in the current environment)


varungup90:
Thanks for adding TLS support! I’ve reviewed the changes and noted a few areas that need adjustment, particularly around listener parity and certificate validation.

**Correctness Bugs**

* **Missing `preserve_external_request_id` patch (`gateway.yaml:122`):** The `EnvoyPatchPolicy` is currently gated on `gateway.http.enabled`. In an HTTPS-only mode (where `http.enabled=false` and `tls.enabled=true`), this patch is silently dropped, which changes request-ID handling behavior with zero errors or warnings.
* **Stale TLS secret reuse (`tls_secret.yaml:9`):** Secret reuse only checks if `tls.crt` and `tls.key` exist, but never validates if the certificate matches the current hostname or is still valid. If the hostname changes or the auto-generated cert's `validityDays` expire, the mismatched certificate will be kept indefinitely until manually deleted.

**Efficiency & Design**

* **Unconditional cert generation (`tls_secret.yaml:11`):** `genSelfSignedCert` runs unconditionally on every render, even when `reuseExisting=true`. This wastes CPU during every install, upgrade, lint, or CI run since the output is just discarded.
* **Duplicated RouteConfiguration patch (`gateway.yaml:91`):** The HTTPS route configuration patch is a near-verbatim copy of the HTTP one. Factoring this into a shared named template would prevent drift (which likely caused the missing patch issue above).
* **Forced HTTP/TLS exclusivity (`_helpers.tpl:34`):** Forcing `http.enabled` and `tls.enabled` to be mutually exclusive blocks standard dual-listener setups, such as using HTTP for health checks or redirects alongside HTTPS for live traffic.
* **Ineffective schema requirement (`values.schema.json:252`):** The schema `if/then` checks only for key presence, not non-empty strings. Because `values.yaml` defaults to `secretName: ""`, the JSON-schema requirement never actually triggers. Only the runtime Helm `required` call catches the empty case.

NelZyhh:
@varungup90 
 Thanks for the review and suggestions.

I updated the TLS/chart changes to address the follow-up points from the review.

For the Envoy patch configuration, I added the missing preserve_external_request_id patch for the HTTPS listener so TLS-only deployments preserve the same request ID behavior as the HTTP path. I also factored the duplicated HTTP/HTTPS RouteConfiguration patch into a shared named template to reduce drift between the two listener paths.
`{{- define "aibrix.gateway.routeConfigurationPatch" -}}
- type: type.googleapis.com/envoy.config.route.v3.RouteConfiguration
  name: {{ .root.Release.Namespace }}/{{ include "aibrix.fullname" .root }}-eg/{{ .listenerName }}
  operation:
    op: add
    path: "/virtual_hosts/0/routes/0"
    value:
      name: original_route
      ...
{{- end -}}
`

`{{- if .Values.gateway.http.enabled }}
{{- include "aibrix.gateway.routeConfigurationPatch" (dict "root" . "listenerName" "http") | nindent 4 }}
{{- end }}
{{- if .Values.gateway.tls.enabled }}
{{- include "aibrix.gateway.routeConfigurationPatch" (dict "root" . "listenerName" "https") | nindent 4 }}
{{- end }}`

For the self-signed certificate flow, I changed the rendering logic so genSelfSignedCert only runs when the existing Secret cannot be reused, instead of running unconditionally on every render. I also added chart-managed annotations for hostname, validity, and expiry metadata so reuse decisions can account for configuration drift and certificate expiration while still leaving pre-existing user-managed Secrets untouched.
`{{- $reuseExisting := and $hasKeyPair $managedHostnameMatches $managedValidityMatches $managedNotExpired }}
{{- $generatedCert := "" }}
{{- if not $reuseExisting }}
{{- $generatedCert = genSelfSignedCert $hostname nil $dnsNames $validityDays -}}
{{- end -}}`

I also relaxed the previous HTTP/TLS exclusivity rule. The chart now allows dual-listener configurations and only requires that at least one of gateway.http.enabled or gateway.tls.enabled be enabled.
`{{- if and (not $httpEnabled) (not $tlsEnabled) -}}
{{- fail "at least one of gateway.http.enabled or gateway.tls.enabled must be true when gateway.enable=true and gateway.envoyAsSideCar=false." -}}
{{- end -}}`

On the validation side, I tightened both schema and runtime checks so gateway.tls.secretName must be non-empty whenever TLS is enabled, and gateway.tls.hostname must be non-empty when autoGenerate=true .

varungup90 (approved):

PR #2587, opened 2026-08-20
fix(kvcache): decode vLLM map-format KV events; tier-aware prefix scoring

### LINKED ISSUE #2285: KV event decoder drops group_idx/medium/lora_name, breaking prefix matching for hybrid and multi-tier models

wuyan-zs:
## Summary

Fixes #2285. While investigating, I found a larger compatibility break:
vLLM switched KV event encoding from positional arrays to **msgpack maps** in
[vllm-project/vllm#42892](https://github.com/vllm-project/vllm/pull/42892)
(merged 2026-06-09). `DecodeEventBatch` still expected `[ts, events]` arrays,
so **every** KV event payload from current vLLM fails to unmarshal
(`msgpack: invalid code=0x88 decoding array length`) — kv-events prefix
routing is effectively dead against current vLLM. The legacy publisher also
sent one event per ZMQ message (tag-first array), which the batch-shaped
decoder never handled either.

## Changes

- `pkg/cache/kvcache/msgpack_decoder.go`: accept all three wire shapes —
  map single event (current vLLM), tag-first array single event (legacy
  vLLM), and `[ts, events]` batch (older code paths / test encoder).
- Parse every `BlockStored`/`BlockRemoved` field: `lora_id`, `medium`,
  `lora_name`, `extra_keys`, `group_idx`, `kv_cache_spec_kind`,
  `kv_cache_spec_sliding_window`, `locality`.
- `pkg/utils/syncprefixcacheindexer`: record the storage `medium` per
  (prefix, pod) and weight `MatchPrefix` scores (GPU=1.0, CPU=0.5,
  STORAGE=0.25; unspecified keeps full score for backward compatibility)
  so pods holding a prefix only on a slower tier are not over-credited.
- Plumb `medium`/`lora_name`/`group_idx` through the `kvevent` handler and
  `store_providers` adapter (previously dropped at the handler).

## Tests

New `decoder_vllm_compat_test.go` verifies decoding against **byte-exact
payloads generated with the real msgspec encoder** for both current map and
legacy array formats (single-event and batch shapes). New
`medium_scoring_test.go` covers tier weighting via `AddPrefixWithMedium` and
the event ingestion path. All existing `kvcache` and
`syncprefixcacheindexer` tests keep passing.

## Follow-up

The `(model, lora_name, group_idx)` index keying part of #2285 needs a
request-side adapter id (`RoutingContext` has no LoRA field today); see the
issue thread.

### DIFF
--- a/pkg/cache/kvcache/decoder_vllm_compat_test.go
+++ b/pkg/cache/kvcache/decoder_vllm_compat_test.go
@@ -1,0 +1,157 @@
+package kvcache
+
+// decoder_vllm_compat_test.go — decoder compatibility with the real vLLM wire
+// formats, verified against byte-exact payloads generated with msgspec
+// (vllm/distributed/kv_events.py):
+//
+//   - map encoding: vLLM >= #42892 (2026-06), single event per ZMQ message
+//   - array encoding: vLLM < #42892, single event per ZMQ message
+//   - [ts, events] batch: legacy batch wrapper
+
+import (
+	"encoding/hex"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+// Real vLLM main (msgspec map encoding, post-#42892) fixtures.
+const (
+	vllmMapMinimal   = "88a474797065ab426c6f636b53746f726564ac626c6f636b5f68617368657391c420000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1fb1706172656e745f626c6f636b5f68617368c0a9746f6b656e5f6964739401020304aa626c6f636b5f73697a6504a76c6f72615f6964c0a66d656469756dc0a96c6f72615f6e616d65c0"
+	vllmMapHybrid    = "89a474797065ab426c6f636b53746f726564ac626c6f636b5f68617368657391c420000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1fb1706172656e745f626c6f636b5f68617368c0a9746f6b656e5f6964739401020304aa626c6f636b5f73697a6504a76c6f72615f6964c0a66d656469756da3475055a96c6f72615f6e616d65c0a967726f75705f69647801"
+	vllmMapExtraKeys = "8aa474797065ab426c6f636b53746f726564ac626c6f636b5f68617368657391c420000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1fb1706172656e745f626c6f636b5f68617368c0a9746f6b656e5f6964739401020304aa626c6f636b5f73697a6504a76c6f72615f6964c0a66d656469756da3475055a96c6f72615f6e616d65a9616461707465722d31aa65787472615f6b6579739192a4696d6730a4656d6230a967726f75705f69647802"
+	vllmMapMaximal   = "8da474797065ab426c6f636b53746f726564ac626c6f636b5f68617368657391c420000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1fb1706172656e745f626c6f636b5f68617368c0a9746f6b656e5f6964739401020304aa626c6f636b5f73697a6504a76c6f72615f696401a66d656469756da3475055a96c6f72615f6e616d65a9616461707465722d31aa65787472615f6b6579739192a4696d6730a4656d6230a967726f75705f69647805b26b765f63616368655f737065635f6b696e64a6687962726964bc6b765f63616368655f737065635f736c6964696e675f77696e646f77cc80a86c6f63616c697479a54c4f43414c"
+	vllmMapRemoved   = "84a474797065ac426c6f636b52656d6f766564ac626c6f636b5f68617368657391c420000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1fa66d656469756da3475055a967726f75705f69647801"
+	vllmMapCleared   = "81a474797065b0416c6c426c6f636b73436c6561726564"
+)
+
+// Legacy array fixtures (pre-#42892). batch-wrapped and bare forms.
+const (
+	legacyArrayHybrid = "92cb41d954fc40000000919aab426c6f636b53746f72656491c420000102030405060708090a0b0c0d0e0f101112131415161718191a1b1c1d1e1fc0940102030404c0a3475055c0c001"
+	legacyArrayShort  = "92cb41d954fc400000009195ab426c6f636b53746f726564926566c0940102030404"
+)
+
+func decodeHex(t *testing.T, s string) []byte {
+	t.Helper()
+	b, err := hex.DecodeString(s)
+	if err != nil {
+		t.Fatalf("bad hex: %v", err)
+	}
+	return b
+}
+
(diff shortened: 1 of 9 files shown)


varungup90:
Thanks for the PR. The tier-aware scoring and format compatibility improvements look solid, but there are a few edge cases involving state updates and logic duplication that need addressing:

* **`sync_hash.go:438` (Stale Medium State):** `PodInfo.Medium` is never updated when a repeat `BlockStored` event arrives for an existing (prefix, pod) entry; only `LastAccessTime` refreshes. If a block moves tiers (e.g., GPU → CPU offload), the pod stays permanently scored at its original medium, undermining the tier-aware scoring logic.
* **`sync_hash.go:340` (Overly Broad Removals):** `ProcessBlockRemoved` ignores `event.Medium` and `event.GroupIdx` and deletes the entire prefix-map entry unconditionally. Since the `BlockRemovedEvent` doc comment claims these fields scope removals to a specific tier/group, a tier-scoped removal (like an eviction from GPU that leaves the block on CPU) will incorrectly erase all knowledge of that pod's cached prefix.
* **`msgpack_decoder.go:226` (Duplicated Parsing Logic):** `parseEventArray` and `parseEventMap` duplicate roughly 30 lines of optional-field parsing logic. Consolidating this into a shared helper function will prevent silent format-specific decoding gaps if one path is updated in the future but the other is missed.

varungup90:
**1. Stale `Medium` state on existing `(prefix, pod)` entries (`sync_hash.go` ~438 and ~579)**

Currently, `AddPrefixWithMedium` and `addPrefixToPodLocked` write `Medium` only when creating a new `PodInfo`. If the pod is already in the map, they only refresh `LastAccessTime`.

This breaks the feature this PR is adding. ZMQ is at-least-once, and vLLM CPU offload can emit another `BlockStored` for the same block (sometimes after `BlockRemoved`). The first event may be GPU or empty; a later one may be CPU/STORAGE. `MatchPrefix` keeps scoring at 1.0, which is the over-credit this PR aims to fix.

Please update the medium on the existing path, ensuring an empty value doesn't overwrite a known tier:

```go
if podInfo, exists := pods[podName]; exists {
    podInfo.LastAccessTime.Store(now)
    if medium != "" {
        podInfo.Medium = medium
    }
}

```

*Suggestion:* Add a test that stores the same `(prefix, pod)` as GPU then CPU, and asserts the score drops from 100 to 50.

**2. "Current vLLM" fixtures do not match the real ZMQ shape (`decoder_vllm_compat_test.go`)**

The live `ZmqEventPublisher` encodes an `EventBatch`, which is still `array_like`: `[ts, [event, ...], dp_rank?]`. PR #42892 only changed the inner `KVCacheEvent` structs to maps. The production payload is an outer array with map events inside.

The new tests cover:

* A top-level single map (`vllmMap*`)
* An outer batch with array inner events (`legacyArray*`)

They never cover `[ts, [{type: BlockStored, ...}]]`. On that path, the old decoder fails with `expected msgpack array, got map[...]`, not the `invalid code=0x88 decoding array length` cited in the PR description (which happens when unmarshaling a map into a slice). While `decodeBatchArray` looks like it can parse inner maps, the production shape lacks a regression test.

*Suggestion:* Add a fixture encoded as a real `KVEventBatch` (`msgspec.msgpack.Encoder().encode(...)`), including an optional `data_parallel_rank` third element, rather than a bare `BlockStored`.

henrichter:
Bumping this. Running into the same issue

bolubo (commented):
I went through the current head (`059fe5ec`) and the follow-up commit. The map-format decoding and the fixtures look right; three things are still open.

**1. `lint` is red because of the duplicated optional-field parsing.**

`lint` is the only failing check on this PR, and the annotation says:

```
pkg/cache/kvcache/msgpack_decoder.go:159: cyclomatic complexity 38 of func `parseEventArray` is high (> 30) (gocyclo)
```

I think most of that complexity sits in the eight optional-field blocks for `BlockStored` (lines 229-268). They account for about 16 of the 38 decision points, and they are the same duplication the earlier review asked to consolidate. Pulling them into one helper shared by `parseEventArray` (line 159) and `parseEventMap` (line 320) should bring the function back under the threshold. Job log: https://github.com/vllm-project/aibrix/actions/runs/32333976535/job/96326109310

**2. A repeat `BlockStored` does not update the tier (reproduced on `059fe5ec`).**

Both `addPrefixToPodLocked` (line 579) and `AddPrefixWithMedium` (line 438) write `Medium` only when the `(prefix, pod)` entry is created; on an existing entry they only refresh `LastAccessTime`. So a same-pod `GPU` then `CPU` report for the same block keeps scoring 100 instead of 50. With `store(m)` wrapping `ProcessBlockStored` for the same block and pod:

```go
store(MediumGPU)                                            // creates the entry with GPU
store(MediumCPU)                                            // repeat BlockStored, same block/pod
matched, _ := table.MatchPrefix(model, -1, block1, readyPods)
// got matched["p1"] == 100, expected 50
```

The tests in this PR give each tier its own pod, which is why this one stays green. Two notes for the fix: only write the tier when the event carries one, and the write needs to land in `addPrefixToPodLocked` too, since that is the path the ZMQ events take (the `AddPrefix` callers never pass a tier).

**3. A question about the new max-weighting in `MatchPrefix`.**

`MatchPrefix` now keeps the maximum weighted value per pod (`sync_hash.go:241`), so a shorter GPU match can outrank a longer CPU match: 50% on GPU gives 50, 60% on CPU gives 30, and 50 wins. Before this change the value was simply the longest matched prefix as a percent. If the intent is still "longest match, discounted by its tier", I think the code should pick the longest matched prefix first and weight that, instead of maxing across prefixes. Since `ScoreAll` feeds these numbers straight into routing (`prefix_cache.go:616`), could you confirm which one is intended?

PR #2592, opened 2026-08-20
[Bug] stop KPA/APA scaling down on unknown metrics

### LINKED ISSUE #2591: KPA/APA should not scale down on unknown metrics and should reject invalid subTargetSelector targets
### Bug description

While smoke-testing PodAutoscaler KPA / APA strategies on minikube, I found two related validation / error-handling issues:

1. Unknown metric names are accepted and then treated as metric value `0`, causing KPA / APA to scale down.
2. `subTargetSelector.roleName` is accepted for a Deployment target under KPA, even though role-level sub-targeting appears to be intended only for StormService.

### What worked

Using a fake vLLM-style Prometheus endpoint and the current central metric registry key `num_requests_running`, both KPA and APA worked:

- KPA + POD metric scaled a Deployment from `2` to `4`.
- APA + POD metric scaled a Deployment up to the configured `maxReplicas: 6`.
- KPA / APA did not create HPA resources; they directly patched the target workload's `spec.replicas`.

#### 1. Unknown metric name is treated as zero and causes scale-down

Example PodAutoscaler:

```yaml
spec:
  scalingStrategy: APA
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: custom-pa-target
  minReplicas: 1
  maxReplicas: 6
  metricsSources:
  - metricSourceType: pod
    protocolType: http
    port: "8000"
    path: /metrics
    targetMetric: running_requests
    targetValue: "1"
```

Observed controller log:

```text
Failed to fetch metric running_requests from pod ...: metric running_requests not found in central registry. Returning zero value.
```

Observed behavior:

- The PodAutoscaler is accepted by admission.
- The controller records metric values as `[0, 0, ...]`.
- KPA / APA computes a scale-down recommendation.
- The workload scales down to `minReplicas=1`.
- PodAutoscaler still reports `ValidSpec=True` and `Ready=True`.

This is dangerous because a typo, stale example, or registry mismatch can silently scale down a live workload. Metric collection/config errors should not be converted into a healthy metric value of `0`.

Expected behavior:

- Unknown `targetMetric` should be rejected during admission if the set of supported central registry metrics is known to the controller/webhook.
- If the metric cannot be validated at admission time, metric fetch failures should make scaling invalid for that reconciliation.
- PodAutoscaler should report something like `FailedComputeScale` / `AbleToScale=False`, and should not apply a scale-down based on synthetic zero values.

#### 2. Existing test/example metric key may not match the current registry

I noticed existing integration coverage uses `running_requests`, while the current central registry key is `num_requests_running`.

With `targetMetric: running_requests`:

- KPA / APA accepts the PodAutoscaler.
- Controller logs `metric running_requests not found in central registry`.
- The metric is treated as `0`.

With `targetMetric: num_requests_running`:

- KPA / APA successfully fetch metrics from a vLLM-style endpoint.
- Scaling behaves as expected.

This suggests test fixtures/examples should be updated to use current metric keys, and regression tests should cover the unknown-metric behavior.

#### 3. KPA accepts `subTargetSelector.roleName` for a Deployment target

Example:

```yaml
spec:
  scalingStrategy: KPA
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: custom-pa-target
  subTargetSelector:
    roleName: prefill
  minReplicas: 1
  maxReplicas: 6
  metricsSources:
  - metricSourceType: pod
    protocolType: http
    port: "8000"
    path: /metrics
    targetMetric: num_requests_running
    targetValue: "1"
```

Observed behavior:

- Server-side dry-run accepts it.
- Live creation accepts it.
- PodAutoscaler status reports `ValidSpec=True` and `Ready=True`.
- No clear invalid-spec signal is surfaced.

Based on the controller code and behavior, `subTargetSelector.roleName` appears intended for StormService role-level autoscaling, not generic Deployment scaling. This should be rejected by admission and/or controller fallback validation unless Deployment sub-targeting is intentionally supported.

### Expected validation / test coverage

Please consider adding:

- Webhook / controller validation for unknown or unsupported `targetMetric` names when possible.
- Regression tests ensuring metric fetch/config errors do not produce synthetic `0` values that trigger scale-down.
- Test fixture/example cleanup from `running_requests` to `num_requests_running`, or a compatibility alias if `running_requests` is still intended to be supported.
- Admission/controller tests rejecting `subTargetSelector.roleName` when `scaleTargetRef.kind` is not `StormService`.

### Environment

- AIBrix version: upstream `main`, commit `136197c8db5ab353052234749a53d0534b3590de`
- Deployment environment: minikube
- Controller image: `aibrix/controller-manager:nightly`
- CRDs and controller deployed from `config/crd` and `config/default`

wieghx:
## Pull Request Description

KPA/APA treated a missing pod metric as `0`, so an unknown `targetMetric` (for example `running_requests` instead of `num_requests_running`) looked like idle load and scaled the workload down. `subTargetSelector.roleName` was also accepted on a Deployment even though role-level scaling is StormService-only.

This change:

- Makes `RestMetricsFetcher` return fetch errors instead of a synthetic `0`. Collector/autoscaler already skip failed scrapes; if every source fails, reconcile reports `FailedComputeScale` and does not change replica count.
- Does **not** reject `metricSourceType=pod` `targetMetric` names against the central registry. Admission and controller spec validation stay open so unregistered (but otherwise valid) pod metrics are not blocked. Kubernetes `external`/`domain`/`custom` metrics are unchanged.
- Rejects `subTargetSelector.roleName` unless `scaleTargetRef.kind` is `StormService` (HPA + roleName remains forbidden).

The current registry key is `num_requests_running`. `running_requests` is not aliased; a typo still fails at scrape time rather than scaling down.

## Related Issues
Resolves: #2591

### DIFF
--- a/pkg/controller/podautoscaler/metrics/fetcher.go
+++ b/pkg/controller/podautoscaler/metrics/fetcher.go
@@ -97,10 +97,9 @@ func (f *RestMetricsFetcher) fetchFromPod(ctx context.Context, pod v1.Pod, sourc
 	// Use the centralized engine fetcher with real pod information
 	metricValue, err := f.engineFetcher.FetchTypedMetric(ctx, endpoint, engineType, identifier, source.TargetMetric)
 	if err != nil {
-		klog.Warningf("Failed to fetch metric %s from pod %s: %v. Returning zero value.",
+		klog.Warningf("Failed to fetch metric %s from pod %s: %v",
 			source.TargetMetric, identifier, err)
-		// Return zero value with warning instead of error - business logic can decide how to handle
-		return 0.0, nil
+		return 0.0, err
 	}
 
 	return metricValue.GetSimpleValue(), nil
@@ -276,9 +275,9 @@ func (f *ExternalMetricsFetcher) fetchFromGPUOptimizer(ctx context.Context, pod 
 	// This gives us a global value that we need to adapt to per-pod semantics
 	metricValue, err := f.engineFetcher.FetchTypedMetric(ctx, source.Endpoint, "external", "gpu-optimizer", source.TargetMetric)
 	if err != nil {
-		klog.Warningf("Failed to fetch metric %s from GPU-Optimizer %s: %v. Returning zero value.",
+		klog.Warningf("Failed to fetch metric %s from GPU-Optimizer %s: %v",
 			source.TargetMetric, source.Endpoint, err)
-		return 0.0, nil
+		return 0.0, err
 	}
 
 	// Adaptation: Global metric -> per-pod value
--- a/pkg/controller/podautoscaler/metrics/fetcher.go
+++ b/pkg/controller/podautoscaler/metrics/fetcher.go
@@ -134,15 +134,14 @@ func (f *ResourceMetricsFetcher) fetchResourceMetric(ctx context.Context, pod v1
 		"metric", source.TargetMetric)
 
 	if f.metricsClient == nil {
-		klog.Warningf("Kubernetes resource metrics client not initialized for metric %s", source.TargetMetric)
-		return 0.0, nil
+		return 0.0, fmt.Errorf("kubernetes resource metrics client not initialized for metric %s", source.TargetMetric)
 	}
 
 	// Use existing ResourceMetricsFetcher logic
 	podMetrics, err := f.metricsClient.MetricsV1beta1().PodMetricses(pod.Namespace).Get(ctx, pod.Name, metav1.GetOptions{})
 	if err != nil {
-		klog.Warningf("Failed to fetch resource metrics for pod %s: %v. Returning zero value.", pod.Name, err)
-		return 0.0, nil
+		klog.Warningf("Failed to fetch resource metrics for pod %s: %v", pod.Name, err)
+		return 0.0, err
 	}
 
 	var total float64
@@ -192,8 +191,7 @@ func (f *CustomMetricsFetcher) fetchCustomMetric(ctx context.Context, pod v1.Pod
 		"metric", source.TargetMetric)
 
 	if f.customMetricsClient == nil {
-		klog.Warningf("Kubernetes custom metrics client not initialized for metric %s", source.TargetMetric)
-		return 0.0, nil
+		return 0.0, fmt.Errorf("kubernetes custom metrics client not initialized for metric %s", source.TargetMetric)
 	}
 
 	// Use existing CustomMetricsFetcher logic
@@ -209,8 +207,8 @@ func (f *CustomMetricsFetcher) fetchCustomMetric(ctx context.Context, pod v1.Pod
 
 	metricList, err := f.customMetricsClient.NamespacedMetrics(pod.Namespace).GetForObject(podGK, podRef.Name, source.TargetMetric, labels.Everything())
 	if err != nil {
-		klog.Warningf("Failed to fetch custom metric %s for pod %s: %v. Returning zero value.", source.TargetMetric, pod.Name, err)
-		return 0.0, nil
+		klog.Warningf("Failed to fetch custom metric %s for pod %s: %v", source.TargetMetric, pod.Name, err)
+		return 0.0, err
 	}
 
 	return float64(metricList.Value.Value()), nil
--- a/pkg/controller/podautoscaler/metrics/fetcher.go
+++ b/pkg/controller/podautoscaler/metrics/fetcher.go
@@ -101,6 +101,9 @@ func (f *RestMetricsFetcher) fetchFromPod(ctx context.Context, pod v1.Pod, sourc
 			source.TargetMetric, identifier, err)
 		return 0.0, err
 	}
+	if metricValue == nil {
+		return 0.0, fmt.Errorf("metric value is nil for %s", source.TargetMetric)
+	}
 
 	return metricValue.GetSimpleValue(), nil
 }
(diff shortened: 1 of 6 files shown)


pkg/webhook/podautoscaler_webhook.go
```diff
@@ -269,6 +273,15 @@ func validatePodMetricSource(ms *autoscalingv1alpha1.MetricSource, msPath *field
 	if ms.Path == "" {
 		errs = append(errs, field.Required(msPath.Child("path"), "required for metricSourceType=pod"))
 	}
+	if ms.TargetMetric != "" {
```
googs1025: I don’t think we should reject POD `targetMetric` at admission time based on `metrics.Metrics`.

  This makes the API accept only metrics that are already present in the central registry, which can block otherwise valid pod metrics that have not been registered yet.

pkg/controller/podautoscaler/podautoscaler_controller.go
```diff
@@ -533,6 +537,12 @@ func (r *PodAutoscalerReconciler) validatePodMetricSource(ms *autoscalingv1alpha
 	if ms.Path == "" {
 		return invalid(ReasonMetricsConfigError, "path is required for metricSourceType=pod.")
 	}
+	if ms.TargetMetric != "" {
```
googs1025: dtto

wieghx: Same on the controller fallback path: spec validation no longer requires POD `targetMetric` to be in the central registry. Fetch errors still surface as `FailedComputeScale` and do not apply a scale-down. Addressed in 82b187c.

pkg/webhook/podautoscaler_webhook.go
wieghx: Admission no longer looks up POD `targetMetric` in `metrics.Metrics`. That check would reject otherwise valid pod metrics that are not (yet) in the central registry — including names already used in integration tests such as `requests_per_second`.

Scrape failures still return an error instead of a synthetic 0, so KPA/APA will not scale down on an unknown or unfetchable metric. Addressed in 82b187c.

pkg/controller/podautoscaler/metrics/fetcher.go
```diff
@@ -97,10 +97,9 @@ func (f *RestMetricsFetcher) fetchFromPod(ctx context.Context, pod v1.Pod, sourc
...
 	if err != nil {
-		klog.Warningf("Failed to fetch metric %s from pod %s: %v. Returning zero value.",
+		klog.Warningf("Failed to fetch metric %s from pod %s: %v",
 			source.TargetMetric, identifier, err)
-		// Return zero value with warning instead of error - business logic can decide how to handle
-		return 0.0, nil
+		return 0.0, err
 	}
 
 	return metricValue.GetSimpleValue(), nil
```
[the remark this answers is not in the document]
googs1025: check you this review comment?  Check the code to see if this is the case.
The current FetchTypedMetric contract should already return a non-nil MetricValue when err is nil, but adding a guard is cheap and prevents a panic if that contract regresses.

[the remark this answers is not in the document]
wieghx: Checked `FetchTypedMetric`: every error path returns `nil, err`, and `parseMetricInstance` / `GetCounterGaugeValue` / `GetHistogramValue` never return a nil value without an error. So `metricValue.GetSimpleValue()` is not reachable with a nil receiver on the current contract.

Added the nil guard anyway at both `FetchTypedMetric` call sites (`fetchFromPod` and `fetchFromGPUOptimizer`). If that contract regresses, the autoscaler reports a fetch error instead of panicking; it still does not treat the missing value as 0.

The lint job failure on this PR is unrelated: golangci-lint install hit a transient `sum.golang.org` INTERNAL_ERROR (`ginkgolinter@v0.16.2`). `test` / `verify` / `check-coverage` passed. This push will re-run lint.

pkg/controller/podautoscaler/podautoscaler_controller.go
googs1025: thanks for this fix. The POD/GPU-Optimizer paths now propagate fetch errors instead of turning them into synthetic `0` values, which addresses the reported POD metric typo case.

  Could you also take a look at the remaining fetchers before we close this issue? `ResourceMetricsFetcher` and `CustomMetricsFetcher` still return `0.0, nil` when the Kubernetes
  metrics/custom metrics client is unavailable or when metric fetch fails:

  - `pkg/controller/podautoscaler/metrics/fetcher.go`: resource metrics client nil / fetch error
  - `pkg/controller/podautoscaler/metrics/fetcher.go`: custom metrics client nil / fetch error

  If KPA/APA can use `metricSourceType=resource` or `custom`, those paths may still convert metric collection failures into healthy zero values and potentially scale down.

wieghx: KPA/APA collect RESOURCE and CUSTOM through `collectFromPods`, so a `0.0, nil` fetch is treated as idle load the same way a missing POD scrape was.

`ResourceMetricsFetcher` and `CustomMetricsFetcher` now return an error when the Kubernetes metrics/custom metrics client is nil or the fetch fails, instead of a synthetic 0. Collector still skips failed scrapes; if every source fails, reconcile reports `FailedComputeScale` and does not change replica count. Covered by unit tests for both nil-client and fetch-failure paths.

googs1025 (approved):
Thanks for the updates. The direction LGTM: metric fetch failures now propagate as errors instead of synthetic zero values, which avoids accidental scale-down, and role-level subTargetSelector validation is clearer.

  I would wait for the remaining CI checks to finish and pass before merging. Once CI is green, this should be ready from my side.

Issue #2534, opened 2026-08-07
Add progressDeadlineSeconds support to StormService rollout progress


googs1025:
## 🚀 Feature Description and Motivation

StormService rollouts for LLM inference workloads can get stuck for common serving-system reasons, including long model weight downloads, image pull or startup delays, engine bootstrap failures, CUDA graph compilation stalls, P/D role coordination issues, or one role becoming ready while another remains blocked.

RayClusterFleet already exposes `spec.progressDeadlineSeconds` to detect rollout progress stalls. StormService has a `Progressing` condition, but it currently acts as a generic “not ready yet” signal. There is no API-level deadline for deciding that rollout progress has stalled.

As a result, a StormService can remain progressing indefinitely without a clear timeout condition for users or automation.

## Use Case

In production model serving, rollout stalls often require operator action, such as checking model artifacts, image availability, runtime logs, or role coordination.

This feature would give users and automation a clear signal when a StormService rollout is stuck instead of leaving it indefinitely Progressing.

## Proposed Solution

Add `spec.progressDeadlineSeconds` to `StormServiceSpec`.

Suggested semantics:

- Default: `600`
- If unset, preserve default rollout timeout behavior.
- If rollout progress does not happen within this deadline, update the StormService `Progressing` condition to indicate failure, for example:
  - `type: Progressing`
  - `status: False`
  - `reason: ProgressDeadlineExceeded`
- The controller should continue reconciling after timeout.
- Paused StormServices should not be considered timed out while paused.

Expected work:

- Add `progressDeadlineSeconds` to `StormServiceSpec`.
- Regenerate CRDs / deepcopy / client code as needed.
- Update StormService progress condition logic.
- Add unit tests.
- Add integration test cases under `test/integration/controller/stormservice_test.go`.

Required integration test cases:

- Default deadline is applied when `progressDeadlineSeconds` is unset.
- Rollout progress refreshes the progress timestamp / condition.
- A stuck rollout transitions `Progressing` to failed after the deadline.
- Paused StormService does not time out while paused.
- After progress resumes, the timeout condition should recover according to the chosen condition semantics.

googs1025:
We can determine whether need this first, and then work on this feature later.  cc @Jeffwan

PR #2604, opened 2026-08-23
feat: add StormService progress deadline

zupengwang:
## Summary

- add `spec.progressDeadlineSeconds` to `StormService`, defaulting to 600 seconds
- track rollout progress without refreshing the deadline on no-op reconciles
- exclude paused time, surface `ProgressDeadlineExceeded`, and continue reconciling so later progress can recover
- regenerate deepcopy, apply-configuration, and CRD artifacts
- cover defaulting, progress, pause, timeout, recovery, and RoleSet identity in unit and controller integration tests

Related to #2534.

## Semantics

The implementation follows the existing RayClusterFleet/Kubernetes-style progress contract:

- increases in updated or ready replicas, or decreases in old replicas, count as progress
- paused StormServices use `DeploymentPaused`; resuming resets the deadline with `DeploymentResumed`
- timeout is status-only (`Progressing=False`, reason `ProgressDeadlineExceeded`)
- the controller keeps reconciling after timeout and recovers when rollout progress resumes
- timeout does not replace the existing RoleSet

This is a Draft PR so maintainers can confirm the desired semantics and pause/resume reason names before it is marked ready.

## Validation

- `go test -race ./pkg/controller/stormservice -count=1`
- state-machine stress: four scenarios x 1,000 repetitions
- race repetition: 20 runs
- `make test`
- `make lint-all`
- `go vet ./pkg/controller/stormservice`
- `bash hack/verify-codegen.sh`
- `hack/verify-crd-sync.sh`
- focused controller envtest for defaulting, progress refresh, pause, timeout, recovery, and RoleSet UID preservation: 3/3 passed
- full controller integration suite: 46/46 passed

The integration workload uses the existing vLLM Pod shape but does not start the container; this change is confined to Kubernetes control-plane rollout status and requeue behavior.


api/orchestration/v1alpha1/stormservice_types.go
```diff
@@ -54,6 +54,14 @@ type StormServiceSpec struct {
 	// +optional
 	Paused bool `json:"paused,omitempty" protobuf:"varint,7,opt,name=paused"`
 
+	// The maximum time in seconds for a StormService to make progress before it
+	// is considered to be failed. The controller will continue to process failed
+	// StormServices and surface a condition with a ProgressDeadlineExceeded reason.
+	// Progress is not estimated while a StormService is paused. Defaults to 600s.
+	// +kubebuilder:default=600
```
googs1025, line 69: `progressDeadlineSeconds` should be validated as a positive value. Right now the CRD/API only sets `default: 600`, but does not add a minimum, so users can create a StormService with `progressDeadlineSeconds: 0` or a negative value. can we valid in webhook ?

googs1025 (commented):
Could you clarify the intended behavior after `ProgressDeadlineExceeded`?

  From the current implementation, it looks like the StormService only surfaces `Progressing=False` with reason `ProgressDeadlineExceeded`, while the controller continues reconciling and does not roll back or replace the existing RoleSet. Is that the expected behavior, i.e. the rollout stays where it is until later progress/ready state clears it?

  Could you clarify the intended pause/resume semantics for the progress deadline?

  The API comment says progress is not estimated while a StormService is paused. In the current implementation, when a StormService resumes from `DeploymentPaused`, the condition is updated to `DeploymentResumed` with a fresh `LastUpdateTime`, which appears to restart the progress deadline from resume time rather than preserving elapsed pre-pause time. Is that intentional?

zupengwang:
@googs1025 Thanks for checking these semantics. Yes, both behaviors are intentional and modeled after Kubernetes Deployment progress-deadline handling:

- `ProgressDeadlineExceeded` is a status signal, not an automatic rollback or replacement policy. The controller keeps reconciling the existing RoleSet, and later observable progress or readiness can update the condition.
- For a rollout that has not already timed out, paused time is excluded. Resuming records `DeploymentResumed` with a fresh `LastUpdateTime`, so the progress deadline restarts from resume time. If the rollout has already timed out, pausing preserves `ProgressDeadlineExceeded` instead of clearing it.

This matches Kubernetes v1.31.8 [`checkPausedConditions`](https://github.com/kubernetes/kubernetes/blob/v1.31.8/pkg/controller/deployment/sync.go#L75-L94), including preserving an existing timeout and using `ConditionUnknown` for pause/resume. The unit and integration tests intentionally cover timeout recovery and paused-time exclusion.

api/orchestration/v1alpha1/stormservice_types.go
zupengwang: Thanks — addressed in `3b3b5f7`. I added structural CRD validation with `+kubebuilder:validation:Minimum=1`, so both zero and negative values are rejected by the API server on create and update. This is enforced independently of the custom webhook availability or its failure policy. I also regenerated the module and Helm CRDs and added integration cases verifying that `0` and `-1` return Kubernetes `Invalid` errors.

pkg/controller/stormservice/progress.go
```diff
@@ -0,0 +1,144 @@
...
+
+func progressDeadline(stormService *orchestrationv1alpha1.StormService) time.Duration {
+	seconds := defaultProgressDeadlineSeconds
+	if stormService.Spec.ProgressDeadlineSeconds != nil {
+		seconds = *stormService.Spec.ProgressDeadlineSeconds
+	}
+	return time.Duration(seconds) * time.Second
+}
+
+func syncStormServiceProgressingCondition(stormService *orchestrationv1alpha1.StormService, oldStatus *orchestrationv1alpha1.StormServiceStatus, now time.Time) {
```
googs1025, line 68: Could you help double-check whether replacing the whole `Status.Conditions` slice here can cause any issue?

  In `syncStormServiceProgressingCondition`, we assign `stormService.Status.Conditions = orchestrationv1alpha1.Conditions{condition}`. I noticed `StormService` defines multiple
  condition types (`Ready`, `Progressing`, `ReplicaFailure`), and there is also a `SetStormServiceCondition` helper that preserves other condition types while updating one condition.

  Could you verify whether the current StormService status model intentionally keeps only one active condition at a time, or whether this path might accidentally drop another condition that should be preserved?

pkg/controller/stormservice/progress.go
```diff
@@ -0,0 +1,144 @@
...
+		} else if currentCondition != nil && currentCondition.Reason == PausedReason {
+			condition = *currentCondition
+		} else {
+			condition = newProgressingCondition(
+				corev1.ConditionUnknown,
+				PausedReason,
+				fmt.Sprintf("StormService %q is paused.", stormService.Name),
+				now,
+			)
+		}
```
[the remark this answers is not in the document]
googs1025: plz check this too

pkg/controller/stormservice/progress.go
zupengwang: Thanks for flagging this. I traced every current StormService Status.Conditions writer. Although the Ready, Progressing, and ReplicaFailure paths currently happen to replace the slice, the API defines multiple condition types and does not document a single-active-condition contract, so replacing the whole slice here can discard an unrelated condition.

I reproduced that risk on the previous head: a focused regression test failed because updating Progressing removed an existing Ready condition. Commit 92687e7 fixes this by filtering out only existing Progressing entries and appending the newly computed Progressing condition, preserving all other condition types unchanged and collapsing malformed duplicate Progressing entries.

I did not call SetStormServiceCondition directly because its same-status/same-reason early return would suppress the LastUpdateTime refresh required when rollout progress is observed. After merging current main, the focused regression, StormService unit and race tests, controller envtest (47/47), webhook envtest (66/66), lint, codegen, manifests, and CRD sync all pass at ae4a360.

pkg/controller/stormservice/progress.go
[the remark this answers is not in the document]
zupengwang: Thanks for asking me to verify this as well. I checked Kubernetes v1.31.8 checkPausedConditions directly. It intentionally returns without replacing an existing ProgressDeadlineExceeded condition when a timed-out Deployment is paused. For a non-timed-out rollout, both DeploymentPaused and DeploymentResumed use ConditionUnknown; resume refreshes LastUpdateTime, while LastTransitionTime remains unchanged because the status stays Unknown.

The current StormService implementation follows those semantics, including excluding paused time from the deadline, so I do not recommend changing this path. Kubernetes reference: https://github.com/kubernetes/kubernetes/blob/v1.31.8/pkg/controller/deployment/sync.go#L75-L94

googs1025 (commented):
I'am OK, for this pr. LGTM. 
cc @varungup90 @Jeffwan  to final review

varungup90:
@zupengwang can you resolve the merge conflict

zupengwang:
Resolved the merge conflict in `2843fa5`. The resolution keeps the deadline-aware `Progressing` state machine for non-ready StormServices, uses current main's type-specific `Ready` update for ready StormServices, and retains the new gang-scheduling condition updates. This preserves unrelated conditions and progress timestamp semantics while incorporating current main.

Local validation passed:
- StormService unit and race tests
- focused progress controller envtests (4/4)
- full controller integration suite (49/49)
- webhook integration suite (77/77)
- `make lint-all`, codegen verification, manifests/CRD sync, client build, and diff checks

varungup90:
**Fatal (must fix)**

**[`[pkg/controller/stormservice/sync.go](https://github.com/vllm-project/aibrix/pull/2604)`](https://github.com/vllm-project/aibrix/pull/2604)**: `Ready` stays `True` after a ready `StormService` starts a rollout

* **Rule**: PP-75 (implement status contract) + CC-153 (boundary: ready → not-ready)
* **Cause**: Before this PR, `setStormServiceAvailabilityCondition(false)` always removed `Ready`. The new branch only runs the `Progressing` state machine and leaves existing conditions in place. After a completed rollout (`Ready=True`, `Progressing` removed), a new template change makes `stormServiceReady` false, so `Progressing` comes back—but `Ready=True` is never cleared. `kubectl print-columns`, dashboards, and status monitors will report ready while the rollout is stuck or timed out.
* **Fix**: When not ready, update `Ready` alongside `Progressing`. To match the existing contract:
```go
if stormServiceReady {
    setStormServiceAvailabilityCondition(&stormService.Status, true)
} else {
    RemoveStormServiceCondition(&stormService.Status, orchestrationv1alpha1.StormServiceReady)
    syncStormServiceProgressingCondition(stormService, checkpoint, time.Now())
}

```


*(Alternative: Set `Ready=False` with `MinimumReplicasUnavailable` instead of deleting it).* Add an envtest verifying: ready `StormService` → template change → `Ready` is not `True`, deadline exceeded, `RoleSet` UID unchanged.
* **Cost / Benefit**: Low cost / High benefit—`Ready` is the primary availability signal; a stale `True` masks rollout failures.

---

**Serious (should fix)**

**[`[pkg/controller/stormservice/sync.go](https://github.com/vllm-project/aibrix/pull/2604)`](https://github.com/vllm-project/aibrix/pull/2604)**: `ReplicaFailure` replaces the entire `Conditions` slice, resetting the deadline clock

* **Rule**: PP-36 (defend new invariant) + CC-153
* **Cause**: On scaling or rollout errors, `Status.Conditions` is overwritten directly with `ReplicaFailure`. The subsequent successful reconcile finds no `Progressing` condition, causing `syncStormServiceProgressingCondition` to generate a fresh `LastUpdateTime`. Intermittent errors (e.g., image pull blips, quota limits) will continuously reset the clock, preventing `ProgressDeadlineExceeded` from triggering.
* **Fix**: Use `SetStormServiceCondition` for `ReplicaFailure` (and clear it on success) instead of assigning `Status.Conditions = []Condition{failure}`. Run progress sync on the error path so the deadline clock continues advancing.
* **Cost / Benefit**: Low cost / Medium benefit—prevents intermittent controller errors from indefinitely delaying rollout timeout detection.

zupengwang:
Thanks for the detailed review. I addressed both items in `034301dd`.

- The status boundary now removes the previous `Ready` condition whenever the StormService is not ready, including reconcile-error paths. `syncStormServiceProgressingCondition` remains responsible only for replacing `Progressing` and preserving unrelated conditions.
- `ReplicaFailure` is now updated type-specifically, cleared after a successful reconcile, and the progress state machine also runs on error paths so transient errors do not reset the deadline clock.

I added a controller envtest covering ready StormService -> template change -> stalled rollout -> `ProgressDeadlineExceeded`; `Ready` is no longer `True`, and the rollout RoleSet UID remains unchanged through timeout. A deterministic unit test covers error/recovery condition ownership and verifies the progress timestamp is not reset.

Local validation passed: focused and full StormService unit tests, StormService race tests, the focused controller envtest, full controller and webhook integration suites, `make lint-all`, codegen verification, manifests/CRD sync, client build, and diff checks.

PR #2617, opened 2026-08-26
[Feat] Consume StormService spec.mode in the controller update path and PodAutoscaler

rishabhsinha17:
## Summary

Implements the follow-ups googs1025 outlined in #2449 after #2450 merged:

- The StormService controller consumes `spec.mode` / `ResolvedMode()` for the update path instead of reading `updateStrategy.type` directly, with webhook validation for conflicting declared combinations.
- PodAutoscaler mode detection moves from the `autoscaling.aibrix.ai/storm-service-mode` annotation to `StormService.spec.mode`, with the annotation kept as a deprecated compatibility fallback during migration.

## Changes

**Controller** (`pkg/controller/stormservice`): `rollout()` now selects the path through a new `EffectiveUpdateStrategyType` helper. A declared `mode: Replica` takes the rolling path and a declared `mode: Pooled` takes the in-place path. When `spec.mode` is unset, the legacy `updateStrategy.type` switch is kept unchanged, so manifests without `spec.mode` behave exactly as before (`ResolvedMode()`'s replicas-based inference is intentionally not used there because it does not always match the declared strategy, e.g. `replicas: 1` with `RollingUpdate`).

**Webhook** (`pkg/webhook`): `validateStormServiceMode` additionally rejects a declared `mode: Replica` combined with a declared `updateStrategy.type: InPlaceUpdate`, on create and update. The reverse combination (`Pooled` + `RollingUpdate`) is deliberately not rejected: the CRD defaults `spec.updateStrategy.type` to `RollingUpdate` whenever the `updateStrategy` block is present, so at admission a user-written `RollingUpdate` is indistinguishable from the default and rejecting it would break plain `mode: Pooled` manifests. The controller resolves that case in favor of the declared mode and logs the override.

**PodAutoscaler** (`pkg/controller/podautoscaler`): new `stormServiceScalingMode` resolves the mode as declared `spec.mode` first, then the `storm-service-mode` annotation (`"replica"`), then the legacy pooled default. Both the role-level read (`getCurrentReplicasForRole`) and write (`setDesiredReplicasForRole`) paths use it. The annotation constant is marked deprecated. A nil `spec.replicas` is no longer dereferenced on the replica-mode read path.

**Docs and samples**: design doc notes that a declared `spec.mode` drives the update path and autoscaler routing; the metric-based autoscaling doc points to `spec.mode` with the annotation as deprecated fallback; the sample PodAutoscalers drop the annotation since the StormServices in those samples already declare `spec.mode` after #2450.

**Tests**: unit tables for `EffectiveUpdateStrategyType`, the new mode/strategy webhook validation, `stormServiceScalingMode` precedence (including a declared mode winning over a stale annotation), and mode-driven get/set scaling paths; integration webhook cases for the new rejection plus the accepted combinations.

Backward compatibility:

- Unset `spec.mode`: no behavior change in controller path selection, validation, or autoscaler routing.
- Declared `mode: Pooled` with a defaulted or declared `RollingUpdate` now takes the in-place path; declared `mode: Replica` with `InPlaceUpdate` is now rejected at admission. Both combinations can only exist on objects that declared `spec.mode`, which shipped in #2450 (merged 2026-08-25, unreleased).

## Validation

```
$ KUBEBUILDER_ASSETS=.../bin/k8s/1.29.0-darwin-arm64 go test -count=1 ./pkg/controller/stormservice/... ./pkg/controller/podautoscaler/... ./pkg/webhook/... ./test/integration/webhook/
ok  	github.com/vllm-project/aibrix/pkg/controller/stormservice	5.607s
ok  	github.com/vllm-project/aibrix/pkg/controller/podautoscaler	0.697s
?   	github.com/vllm-project/aibrix/pkg/controller/podautoscaler/aggregation	[no test files]
ok  	github.com/vllm-project/aibrix/pkg/controller/podautoscaler/algorithm	0.263s
ok  	github.com/vllm-project/aibrix/pkg/controller/podautoscaler/context	0.262s
ok  	github.com/vllm-project/aibrix/pkg/controller/podautoscaler/metrics	0.616s
ok  	github.com/vllm-project/aibrix/pkg/controller/podautoscaler/monitor	0.437s
ok  	github.com/vllm-project/aibrix/pkg/controller/podautoscaler/types	0.259s
ok  	github.com/vllm-project/aibrix/pkg/webhook	0.562s
ok  	github.com/vllm-project/aibrix/test/integration/webhook	7.255s
```

`golangci-lint run` (v1.57.2, whole repo): exit 0, no findings. `go build ./...`: exit 0.

## Related Issues

Part of #2449. The `/scale` subresource enforcement for `mode: Pooled` from the same follow-up list is left as a separate change.

### DIFF
--- a/pkg/controller/podautoscaler/workload_scale.go
+++ b/pkg/controller/podautoscaler/workload_scale.go
@@ -39,7 +39,33 @@ import (
 	"github.com/vllm-project/aibrix/pkg/controller/constants"
 )
 
+// AutoscalingStormServiceModeAnnotationKey is the PodAutoscaler annotation that
+// historically selected how role-level scaling applies to a StormService target:
+// "replica" scales StormService.spec.replicas, anything else scales the targeted
+// role's replicas.
+//
+// Deprecated: declare StormService.spec.mode instead. The annotation is only
+// honored as a compatibility fallback when the target StormService does not
+// declare spec.mode, and may be removed once spec.mode is broadly adopted.
 const AutoscalingStormServiceModeAnnotationKey = "autoscaling.aibrix.ai/storm-service-mode"
+
+// stormServiceScalingMode resolves the deployment mode used to route role-level
+// scaling for a StormService target. A declared StormService.spec.mode is the
+// source of truth. When spec.mode is unset, the deprecated
+// autoscaling.aibrix.ai/storm-service-mode annotation on the PodAutoscaler is
+// honored as a compatibility fallback ("replica" selects replica mode), and any
+// other value keeps the legacy pooled default. spec.replicas is intentionally
+// not used for inference here: replicas == 1 cannot distinguish a pooled
+// StormService from a scaled-down replica-mode one.
+func stormServiceScalingMode(pa *autoscalingv1alpha1.PodAutoscaler, ss *orchestrationv1alpha1.StormService) orchestrationv1alpha1.StormServiceMode {
+	if ss.Spec.Mode != "" {
+		return ss.Spec.Mode
+	}
+	if pa.Annotations[AutoscalingStormServiceModeAnnotationKey] == "replica" {
+		return orchestrationv1alpha1.StormServiceReplicaMode
+	}
+	return orchestrationv1alpha1.StormServicePooledMode
+}
 
 // WorkloadScale provides scaling operations for different workload types.
 // It provides the mechanism to get/set replica counts on workload resources,
@@ -152,9 +178,11 @@ func (s *workloadScale) getCurrentReplicasForRole(ctx context.Context, pa *autos
 		return 0, err
 	}
 
-	// replica mode, return the replicas directly
-	// we can not easily use `*ss.Spec.Replicas > 1` as condition since 1 could be pool or replica both case under autoscaling scenarios
-	if pa.Annotations[AutoscalingStormServiceModeAnnotationKey] == "replica" {
+	// Replica mode scales the whole StormService, so report spec.replicas directly.
+	if stormServiceScalingMode(pa, ss) == orchestrationv1alpha1.StormServiceReplicaMode {
+		if ss.Spec.Replicas == nil {
+			return 0, nil
+		}
 		return *ss.Spec.Replicas, nil
 	}
 
@@ -246,9 +274,9 @@ func (s *workloadScale) setDesiredReplicasForRole(ctx context.Context, pa *autos
 			return err
 		}
 		upd := cur.DeepCopy()
-		// TODO: tricky part. it's hard to know replica=1 is pooling or replica mode, we can use autoscaling to limit it.
-		// we can extract the method and fallback to ss object annotation to check as well.
-		if pa.Annotations[AutoscalingStormServiceModeAnnotationKey] == "replica" {
+		// Replica mode scales the whole StormService through spec.replicas; pooled mode
+		// scales the targeted role. See stormServiceScalingMode for the resolution order.
+		if stormServiceScalingMode(pa, cur) == orchestrationv1alpha1.StormServiceReplicaMode {
 			upd.Spec.Replicas = ptr.To(replicas)
 			return s.client.Patch(ctx, upd, client.MergeFrom(cur))
 		}
--- a/pkg/controller/podautoscaler/workload_scale.go
+++ b/pkg/controller/podautoscaler/workload_scale.go
@@ -178,12 +178,11 @@ func (s *workloadScale) getCurrentReplicasForRole(ctx context.Context, pa *autos
 		return 0, err
 	}
 
-	// Replica mode scales the whole StormService, so report spec.replicas directly.
+	// Replica mode scales the whole StormService, so report spec.replicas. An omitted
+	// spec.replicas resolves to the documented default of 1 instead of reading as a
+	// scaled-to-zero workload, matching how the stormservice controller reconciles it.
 	if stormServiceScalingMode(pa, ss) == orchestrationv1alpha1.StormServiceReplicaMode {
-		if ss.Spec.Replicas == nil {
-			return 0, nil
-		}
-		return *ss.Spec.Replicas, nil
+		return ss.Spec.ResolvedReplicas(), nil
 	}
 
 	roleName := pa.Spec.SubTargetSelector.RoleName
--- a/pkg/controller/stormservice/sync.go
+++ b/pkg/controller/stormservice/sync.go
@@ -145,10 +145,7 @@ func (r *StormServiceReconciler) scaling(ctx context.Context, stormService, curr
 	}
 	// skip scaling when there are terminating roleSets
 	activeRoleSets, _ := filterTerminatingRoleSets(allRoleSets)
-	var expectReplica int32
-	if stormService.Spec.Replicas != nil {
-		expectReplica = *stormService.Spec.Replicas
-	}
+	expectReplica := stormService.Spec.ResolvedReplicas()
 	minAvailable := MinAvailable(stormService)
 	maxSurge := MaxSurge(stormService)
 	diff := len(activeRoleSets) - int(expectReplica)
@@ -264,10 +261,7 @@ func (r *StormServiceReconciler) rollout(ctx context.Context, stormService, curr
 	if err != nil {
 		return err
 	}
-	var expectReplica int32
-	if stormService.Spec.Replicas != nil {
-		expectReplica = *stormService.Spec.Replicas
-	}
+	expectReplica := stormService.Spec.ResolvedReplicas()
 	updated, _ := filterRoleSetByRevision(allRoleSets, updateCR.Name)
 	if len(updated) == int(expectReplica) {
 		return nil
@@ -318,10 +312,7 @@ func (r *StormServiceReconciler) rollingUpdate(allRoleSets []*orchestrationv1alp
 	}
 
 	// 2. create roleset, follow the max surge rule
-	var expectedReplica int
-	if stormService.Spec.Replicas != nil {
-		expectedReplica = int(*stormService.Spec.Replicas)
-	}
+	expectedReplica := int(stormService.Spec.ResolvedReplicas())
 	surge := utils.MinInt(expectedReplica+int(maxSurge)-len(allRoleSets), expectedReplica-len(updated))
 	if surge < 0 {
 		surge = 0
@@ -396,13 +387,10 @@ func (r *StormServiceReconciler) updateStatus(ctx context.Context, stormService 
 	stormService.Status.ReadyReplicas = int32(len(ready))
 	stormService.Status.NotReadyReplicas = int32(len(notReady))
 	// set conditions
-	var specReplica int32
-	if stormService.Spec.Replicas != nil {
-		specReplica = *stormService.Spec.Replicas
-	}
+	specReplica := stormService.Spec.ResolvedReplicas()
 	stormServiceReady := stormService.Status.ReadyReplicas >= specReplica &&
-		stormService.Status.UpdatedReplicas == *stormService.Spec.Replicas &&
-		stormService.Status.Replicas == *stormService.Spec.Replicas &&
+		stormService.Status.UpdatedReplicas == specReplica &&
+		stormService.Status.Replicas == specReplica &&
 		stormService.Status.CurrentRevision == stormService.Status.UpdateRevision
 	if stormServiceReady {
 		setStormServiceAvailabilityCondition(&stormService.Status, true)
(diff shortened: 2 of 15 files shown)


varungup90:
### Code Review Suggestions

#### 1. `IsRollingUpdate()` out of sync with `EffectiveUpdateStrategyType`

* **File**: `pkg/controller/stormservice/utils.go:93`
* **Violations**: [[PP-15](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself) (DRY) • [[PP-37](https://en.wikipedia.org/wiki/Design_by_contract)](https://en.wikipedia.org/wiki/Design_by_contract) (Design by Contract) • [[CC-153](https://en.wikipedia.org/wiki/Boundary_testing)](https://en.wikipedia.org/wiki/Boundary_testing) (Incorrect Boundary Behavior)
* **Why**: `rollout()` now follows a declared `spec.mode` (`Pooled` $\rightarrow$ in-place even when `type` is the CRD-defaulted `RollingUpdate`). However, `MaxSurge`, `MinAvailable`, and `MaxUnavailable` still read through `IsRollingUpdate()`, which `scaling()` uses on every reconcile. For `mode: Pooled` + defaulted `RollingUpdate` + `maxSurge > 0`, scale-out can create extra `RoleSets`, violating the pooled "one `RoleSet`" contract. The PR's design note already treats this type combination as the common case.
* **Fix**: Update `IsRollingUpdate` (or `MaxSurge`/`MinAvailable`) to query `EffectiveUpdateStrategyType`. In `Pooled` mode, surge and unavailable values must evaluate to `0`. Add a test for `mode: Pooled` + defaulted `RollingUpdate` + `maxSurge: 2` asserting that a second `RoleSet` is never created.
* **Cost**: **Low** — one helper update in the same package and a controller/utils test.
* **Benefit**: **High** — `spec.mode` is the primary path this PR introduces. Extra `RoleSets` duplicate the entire role set, disrupting capacity allocation and traffic splitting with no user workaround.

---

#### 2. Excessive logging on steady-state CRD defaults in `rollout()`

* **File**: `pkg/controller/stormservice/utils.go:122`
* **Violations**: [[PP-72](https://en.wikipedia.org/wiki/KISS_principle)](https://en.wikipedia.org/wiki/KISS_principle) (Keep It Simple) • [[CC-21](https://en.wikipedia.org/wiki/Single-responsibility_principle)](https://en.wikipedia.org/wiki/Single-responsibility_principle) (Do One Thing)
* **Why**: When the `updateStrategy` block is present, the CRD defaults `type` to `RollingUpdate`. This represents steady-state behavior for pooled objects, not an anomaly. Because `sync()` enters `rollout()` whenever it isn't scaling, this `klog.Infof` line will flood operator logs and obscure real rollout errors. Additionally, the `Replica` + `InPlaceUpdate` override log is effectively unreachable (since the webhook rejects this combination) and shouldn't log at `Info` level either.
* **Fix**: Lower the log level for steady-state overrides to `klog.V(4)` or emit the log only when the strategy type actually changes.
* **Cost**: **Low** — simple log-level change in a single file.
* **Benefit**: **Medium** — prevents production log pollution across every active, ready pooled `StormService` during routine reconcile loops.

rishabhsinha17:
Thanks for the review. Both addressed in d419b4b.

**1. `IsRollingUpdate` out of sync with `EffectiveUpdateStrategyType`**

Good catch on the scaling path. `IsRollingUpdate` now resolves the strategy through `EffectiveUpdateStrategyType` instead of reading `updateStrategy.type` directly, so `MaxSurge`, `MaxUnavailable`, and `MinAvailable` all evaluate to 0 in Pooled mode (and for an unresolvable mode, which `rollout()` already rejects). I audited the remaining direct readers of `updateStrategy.type`: outside test helpers the only one left is the webhook's Replica + InPlaceUpdate rejection, which intentionally validates the declared type at admission; `rollout()` already goes through `EffectiveUpdateStrategyType`, and the `roleset`/`rayclusterfleet` hits are different types without a mode field.

Added `TestScalingPooledModeNeverCreatesSecondRoleSet` (sync_test.go), driving `scaling()` with `mode: Pooled`, the CRD-defaulted `RollingUpdate`, and `maxSurge: 2`. It asserts exactly one RoleSet exists after scaling across three cases: steady state, scale-out from zero, and a terminating RoleSet awaiting replacement. The last one is the regression: the old helper handed `scaling()` a surge budget of 2 and it created a second RoleSet next to the terminating one (verified the test fails against the previous code with 2 RoleSets). `TestPooledModeZeroesSurgeAndUnavailable` and an extended `TestIsRollingUpdate` cover the helper contract, including `maxSurge: 2` as a plain integer.

**2. Steady-state override logging**

Both override logs in `EffectiveUpdateStrategyType` are now `klog.V(4)`. The Pooled + RollingUpdate one is steady state as you note, so it is verbose-only. The Replica + InPlaceUpdate one is unreachable past the webhook guard; kept at V(4) with a comment saying it only fires for objects that bypassed admission, rather than removing it.

`go test -count=1` passes on `pkg/controller/stormservice`, `pkg/controller/podautoscaler`, `pkg/webhook`, and `test/integration/webhook`; `golangci-lint run` is clean.

pkg/controller/stormservice/sync.go
```diff
@@ -272,17 +272,16 @@ func (r *StormServiceReconciler) rollout(ctx context.Context, stormService, curr
...
-	switch stormService.Spec.UpdateStrategy.Type {
-	case "":
-		// By default use RollingUpdate strategy
-		fallthrough
-	case orchestrationv1alpha1.RollingUpdateStormServiceStrategyType:
-		return r.rollingUpdate(allRoleSets, stormService, current, currentCR, updateCR)
-	case orchestrationv1alpha1.InPlaceUpdateStormServiceStrategyType:
+	// The update path follows the declared spec.mode when it is set and falls back to
+	// the legacy updateStrategy.type selection otherwise, see EffectiveUpdateStrategyType.
+	strategyType, err := EffectiveUpdateStrategyType(stormService)
```
googs1025, line 271: Could you please double-check the `spec.replicas == nil` handling around the reconcile path? `spec.replicas` is optional, and the generated CRD does not seem to set a real `default: 1`. If omitted replicas can reach reconcile, `scaling()` / `rollout()` appear to derive `expectReplica` as `0`, while `updateStatus()` still compares against `*stormService.Spec.Replicas`. Could you verify whether replicas is actually defaulted before reconcile in this codebase, or whether these paths should resolve nil to the documented default of `1`?

pkg/controller/podautoscaler/workload_scale.go
```diff
@@ -152,9 +178,11 @@ func (s *workloadScale) getCurrentReplicasForRole(ctx context.Context, pa *autos
...
 	}
 
-	// replica mode, return the replicas directly
-	// we can not easily use `*ss.Spec.Replicas > 1` as condition since 1 could be pool or replica both case under autoscaling scenarios
-	if pa.Annotations[AutoscalingStormServiceModeAnnotationKey] == "replica" {
+	// Replica mode scales the whole StormService, so report spec.replicas directly.
+	if stormServiceScalingMode(pa, ss) == orchestrationv1alpha1.StormServiceReplicaMode {
+		if ss.Spec.Replicas == nil {
+			return 0, nil
+		}
```
googs1025: Could you please double-check this nil branch? Since `spec.replicas` is optional and documented as defaulting to `1`, returning `0` here may make the autoscaler interpret an omitted replica count as zero replicas. Could you verify whether this field is defaulted before the autoscaler reads it, or whether this should resolve nil to the documented default of `1`?

pkg/controller/stormservice/sync.go
rishabhsinha17: Verified: nothing defaults spec.replicas before reconcile. The CRD schema carries no default (config/crd/orchestration/orchestration.aibrix.ai_stormservices.yaml:61-63) and the mutating webhook only does annotation gated sidecar injection (pkg/webhook/stormservice_webhook.go:50-70), so nil reaches the controllers.

The split you describe was real and actually worse than expectReplica == 0. scaling() calls MinAvailable and MaxSurge on every reconcile, and both dereferenced the pointer unconditionally (old pkg/controller/stormservice/utils.go:63,79,88), so a StormService created without replicas panicked the controller on its first reconcile. That predates this PR (the sites blame to e3bb459a), but both of your comments trace to the same root cause, so I fixed it here.

cd9add82 adds StormServiceSpec.ResolvedReplicas() (api/orchestration/v1alpha1/stormservice_types.go:103-114), which resolves nil to the documented default of 1 and preserves explicit 0, and routes scaling(), rollout(), the rolling update budgets and updateStatus() through it (sync.go:148,264,315,385, utils.go:63,80,89). The raw derefs at the old sync.go:399-400 are gone. TestScalingNilReplicasResolvesToDefault segfaults on the previous code and now asserts exactly one RoleSet is created for an omitted replicas.

pkg/controller/podautoscaler/workload_scale.go
rishabhsinha17: You are right, 0 was the wrong resolution. Nothing defaults the field before the autoscaler reads it: the CRD schema has no default for spec.replicas (config/crd/orchestration/orchestration.aibrix.ai_stormservices.yaml:61-63) and the mutating webhook only injects the runtime sidecar.

Fixed in cd9add82: the branch now returns ss.Spec.ResolvedReplicas() (pkg/controller/podautoscaler/workload_scale.go:185), a helper next to ResolvedMode() that resolves nil to the documented default of 1 and preserves explicit 0 (api/orchestration/v1alpha1/stormservice_types.go:103-114). Covered by TestGetCurrentReplicasReplicaModeNilReplicas. The stormservice reconcile path now goes through the same helper, details in the sync.go thread.

rishabhsinha17:
Pushed cd9add82 addressing both nil replicas review threads from @googs1025.

Root cause shared by both: spec.replicas is optional with a documented default of 1, but neither the CRD schema nor the mutating webhook materializes that default, so an omitted value reaches the controllers as nil. scaling() and rollout() derived expectReplica as 0, MinAvailable/MaxSurge/MaxUnavailable and the readiness comparison in updateStatus() dereferenced the nil pointer (a pre-existing panic on the first reconcile of any StormService created without replicas), and the autoscaler replica mode read reported 0.

The commit adds StormServiceSpec.ResolvedReplicas() (nil resolves to 1, explicit 0 preserved, sibling of ResolvedMode()) and routes every StormService replica read in the reconcile and autoscaler paths through it.

New tests: TestScalingNilReplicasResolvesToDefault (segfaults without the fix), TestGetCurrentReplicasReplicaModeNilReplicas, TestStormServiceSpecResolvedReplicas. go test for api/orchestration/v1alpha1, pkg/controller/stormservice, pkg/controller/podautoscaler, pkg/webhook and test/integration/webhook plus golangci-lint all pass locally.

rishabhsinha17:
Rebased onto main (991fb82c). The branch is linear again: 99b91d46, 6807852c, ddf7cf1d.

Conflict resolution:
- `pkg/webhook/stormservice_webhook.go` and `test/integration/webhook/stormservice_webhook_test.go` vs #2629: kept the Volcano gang validation (`validateStormServiceSchedulingStrategy`, `schedulingConfigChanged`, the new `ValidateUpdate` signature) and layered the Replica + InPlaceUpdate rejection in `validateStormServiceMode` on top. The four mode entries sit next to the gang entries in the same table.
- `pkg/controller/stormservice/sync_test.go` vs #2604: the same package was imported under two aliases; unified on upstream's `appsv1`. Upstream's progress deadline and gang condition tests are kept alongside `TestScalingPooledModeNeverCreatesSecondRoleSet` and `TestScalingNilReplicasResolvesToDefault`.
- `sync.go` replica reads go through `ResolvedReplicas()` on top of the #2604 / #2632 condition handling. #2628 only overlapped the design doc and merged cleanly.

All three commits are signed off and DCO is green. Locally: `go build ./...`, `go test` for `api/orchestration/v1alpha1`, `pkg/controller/stormservice`, `pkg/controller/podautoscaler`, `pkg/webhook` and `test/integration/webhook` (81/81 against regenerated CRDs), and golangci-lint v1.57.2 all pass.

@varungup90 @googs1025 could you take another look? I could not re-request review from my side.

googs1025 (approved):
LGTM overall. the previous IsRollingUpdate/effective strategy mismatch is addressed, nil spec.replicas now resolves through ResolvedReplicas(), and the controller/autoscaler/webhook paths look consistent with the intended spec.mode precedence.

googs1025:
cc @varungup90

varungup90 (approved):

PR #2625, opened 2026-08-27
fix(modeladapter): surface missing backend scheduling reasons

### LINKED ISSUE #1885: Disclose exact pending reason for failed to schedule lora
### 🐛 Describe the bug

Image

if we do not have qualified base models, the error message shows like this.

However, the information is not that clear to end users, we need to improve and tell there's no qualified backend pods etc.

Image

### Steps to Reproduce

```
kubectl apply -f samples/adapter/adapter.yaml
```

### Expected behavior

It should expose more information

### Environment

nightly

mikemikimike:
## Summary

When a single-replica ModelAdapter has no ready backend pods, or fewer ready pods than required, the controller only logs the situation and leaves users without an actionable status message.

This change records a `Scheduled=False` condition with a specific reason and message for both cases, while preserving the existing requeue behavior.

## Validation

- `gofmt` on changed Go files
- `go test ./pkg/controller/modeladapter/...`
- `go vet ./pkg/controller/modeladapter/...`
- `git diff --check`

Fixes #1885

### DIFF
--- a/pkg/controller/modeladapter/modeladapter_controller.go
+++ b/pkg/controller/modeladapter/modeladapter_controller.go
@@ -88,6 +88,10 @@ const (
 	ModelAdapterLoadingErrorReason = "ModelAdapterLoadingError"
 	// ValidationFailedReason is added when model adapter object fails the validation
 	ValidationFailedReason = "ValidationFailed"
+	// NoReadyPodsReason is added when no backend pods are ready for scheduling.
+	NoReadyPodsReason = "NoReadyPods"
+	// InsufficientReadyPodsReason is added when fewer backend pods are ready than required.
+	InsufficientReadyPodsReason = "InsufficientReadyPods"
 	// StableInstanceFoundReason is added if there's stale pod and instance has been deleted successfully.
 	StableInstanceFoundReason = "StableInstanceFound"
 	// ConditionNotReason is added when there's no condition found in the cluster.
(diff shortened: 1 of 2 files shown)


varungup90:
### Code Review Feedback

Thanks for submitting this PR! Here are a few critical findings regarding test coverage, logic execution, and API performance:

* **Dead Test File (`pkg/controller/modeladapter/modeladapter_controller_unit_tests.go:527`)**
* The newly added `TestNewSchedulingPendingCondition` (along with the pre-existing `TestScheduledPodsManagement`) is located in a file named `*_unit_tests.go` instead of standard Go convention `*_test.go`.
* As a result, `go test ./...` ignores these tests entirely and they are not being executed in CI. Renaming the file to end in `_test.go` will ensure they run properly.


* **Unreachable Logic Branch (`pkg/controller/modeladapter/modeladapter_controller.go:679`)**
* The `InsufficientReadyPodsReason` logic currently can never trigger.
* Because `desiredReplicas` is hardcoded to `1` (and `Spec.Replicas` is not actively read), `neededReplicas` evaluates to `1`. Consequently, there is no integer value satisfying `0 < candidatePods < neededReplicas`, making this branch dead code.


* **Redundant API Subresource Writes (`pkg/controller/modeladapter/modeladapter_controller.go:682`)**
* `updateStatus` is being invoked unconditionally during every 5-second reconcile retry when pods remain unavailable.
* Although a `changed` flag is calculated internally, it isn't utilized to skip redundant writes, which risks generating excessive status updates to the API server at scale.

varungup90 (approved):

PR #2633, opened 2026-08-29
fix(batch): scale concurrency with live capacity

scarlet25151:
## Summary
- scale fixed and adaptive dispatch concurrency with live discovered capacity
- keep one recovery probe when live capacity reaches zero so EOF, errors, and deadlines still progress
- propagate configured runtime capacity for Kubernetes discovery backends and bound no-endpoint retries by the authoritative job deadline

## Validation
- 114 targeted pytest cases passed
- ruff format/check passed
- git diff --check passed


zhangjyr (approved):
LGTM, please resolve conflict first before merge.

PR #2646, opened 2026-09-01
[Bug] Batch: validate scheduled jobs concurrently

scarlet25151:
## Pull Request Description

Batch job admission currently runs inside the scheduling loop before the job is registered in the execution pool. A slow validation therefore blocks later jobs even when `pool_size` has available capacity.

This change moves admission into the tracked per-job task so that:
- independent jobs can validate concurrently;
- `pool_size` bounds the full admission and execution lifecycle;
- declined or failed admissions release their pool slot without blocking the scheduler;
- duplicate active job IDs are skipped instead of overwriting tracked tasks.

The concurrency regression test blocks four admissions simultaneously with `pool_size=4` and verifies that all four start before execution is released, matching the multi-job scheduling scenario shown in the PR discussion.

## Related Issues

N/A

## Testing

- `poetry run pytest -q tests/batch/test_scheduler.py tests/batch/test_batch_manager.py tests/batch/job_driver/test_deployment_driver.py` (70 passed)
- `poetry run pytest -q tests/batch/test_e2e_openai_batch_api.py -k "respects_default_job_pool_size or second_job_does_not_stay_validating_when_pool_has_capacity"` (2 passed)
- `poetry run ruff check .`
- `poetry run ruff format --check .`
- `poetry run mypy aibrix/batch/batch_scheduler.py`
- `poetry run pytest -q tests` (1040 passed, 25 skipped, 1 xfailed)

### DIFF
--- a/python/aibrix/aibrix/batch/batch_scheduler.py
+++ b/python/aibrix/aibrix/batch/batch_scheduler.py
@@ -25,7 +25,7 @@ like an OS scheduler); the per-job execution itself lives in the JobDriver.
 
 import asyncio
 import time
-from typing import TYPE_CHECKING, Optional, Tuple
+from typing import TYPE_CHECKING, Optional
 
 import aibrix.batch.constant as constant
 from aibrix.batch.job_entity import BatchJobState
@@ -56,8 +56,8 @@ class BatchScheduler:
         the scheduler keeps no duplicated due-time / inactive bookkeeping — the
         registry is the single source of truth for job state.
 
-        ``pool_size`` caps the number of concurrently executing jobs while the
-        policy still decides which pending job is admitted next.
+        ``pool_size`` caps the number of jobs being admitted or executed while
+        the policy still decides which pending job is admitted next.
         """
         self._context = context
         self._job_progress_manager = job_progress_manager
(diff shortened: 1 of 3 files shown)


Jeffwan:
did we encounter any issues recently

scarlet25151:
@Jeffwan this is regarding to the scheduling part, 
image
in the current implementation even we increase the provision concurrency, the scheduling is sequentially and one job at a time, which will cause a long waiting time. this pr fix the issue.

python/aibrix/aibrix/batch/batch_scheduler.py
```diff
@@ -168,21 +177,20 @@ async def jobs_running_loop(self) -> None:
...
 
-            if scheduled:
-                one_job, job_driver = scheduled
+            if job_id is not None:
                 task = self._serve_loop.create_task(
-                    self._execute_scheduled_job(one_job, job_driver)
+                    self._admit_and_execute_scheduled_job(job_id)
                 )
-                self._job_execution_tasks[one_job] = task
+                self._job_execution_tasks[job_id] = task
```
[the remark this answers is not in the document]
scarlet25151: Addressed in e911e69b. The scheduler now skips a job ID that is already present in `_job_execution_tasks`, so an active admission/execution task cannot be overwritten. Added `test_jobs_running_loop_skips_duplicate_active_job` to verify the original task remains tracked. I also expanded the concurrency regression test to block four admissions simultaneously (`pool_size=4`), matching the multi-job scheduling scenario in the screenshot.

python/aibrix/aibrix/batch/batch_scheduler.py
```diff
@@ -168,21 +177,26 @@ async def jobs_running_loop(self) -> None:
...
 
-            if scheduled:
-                one_job, job_driver = scheduled
-                task = self._serve_loop.create_task(
-                    self._execute_scheduled_job(one_job, job_driver)
-                )
-                self._job_execution_tasks[one_job] = task
+            if job_id is not None:
+                if job_id in self._job_execution_tasks:
+                    logger.warning(
```
zhangjyr, line 191: Why is this check necessary? It's ok for a safeguard. I'm just curious whether there has been a real failure case previously seen.

zhangjyr (approved):
Overall, LGTM

python/aibrix/aibrix/batch/batch_scheduler.py
scarlet25151: there is no lock mechanism in job submit and there would be a possible for the job and duplicate, here is some defensive program.

Issue #766, opened 2025-02-28
Loading base model/lora adapters from s3 fails and throws HF error


robert-moyai:
### 🐛 Describe the bug

Im trying to initialize AIBrix using model files hosted on s3. But I am not able to get vllm-openai container to run as its incorrectly identifying the s3 link as being a HF repo.

`huggingface_hub.errors.HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name': 's3://s3-bucket/llama-3.1-nemoguard-8b-topic-control'. Use `repo_type` argument if needed.`

### Steps to Reproduce

1. Move a model `llama-3.1-nemoguard-8b-topic-control` to a s3 bucket
2. Deploy AIBrix following the docs: https://aibrix.readthedocs.io/latest/features/lora-dynamic-loading.html#create-base-model. Be sure to reference the model using its s3 path: `s3://s3-bucket/llama-3.1-nemoguard-8b-topic-control`
3. Inspect the pod logs and find the error. Pod will crash when applying.
4. Same issue if you just load LoRA from s3

### Expected behavior

Pod should just start running like when downloading model from HF

### Environment

- AIBrix version: 0.20.0
- k8s deployment
- AWS s3 bucket with model files

PR #2654, opened 2026-09-02
[Bug] Fail fast when direct-path LoRA adapter uses s3/gcs/tos URL

pujitha24:
## Pull Request Description

When a `ModelAdapter`'s `ArtifactURL` points to `s3://`, `gcs://`, or `tos://` and the target pod does **not** have the `aibrix-runtime` sidecar (the default setup, unless the pod has the `model.aibrix.ai/sidecar-injection: "true"` annotation and the controller is started with `--enable-runtime-sidecar`), `loadAdapterCall`'s direct-engine branch in `pkg/controller/modeladapter/lora_client.go` only transforms `huggingface://`/`hf://` URLs and forwards every other scheme unchanged as `lora_path` straight to vLLM/SGLang's native load-adapter API. Those engines can't resolve a URL like that and misread it as a HuggingFace repo id, producing the crash from the issue:

```
huggingface_hub.errors.HFValidationError: Repo id must be in the form 'repo_name' or 'namespace/repo_name'
```

**This PR does not implement S3/GCS/TOS downloading in the Go controller, and does not make it possible to load such adapters without the runtime sidecar.** As a maintainer pointed out in review, issue #766's reporter wants to load from S3 *without* sidecar configuration, and their base-model S3 path crashes the vLLM container directly — neither of those is addressed here, so issue #766 should stay open.

What this PR actually does: it makes the non-sidecar (direct) `ModelAdapter` load path fail fast with a clear, actionable error instead of silently forwarding a URL the engine cannot use, so the failure is diagnosable at the controller instead of surfacing as a cryptic HF error deep in the vLLM/SGLang container logs. The sidecar-delegation branch of the same function already handles `s3://`/`gcs://`/`tos://` correctly (it forwards the URL to the `aibrix-runtime` sidecar, which has real scheme-aware download logic in `python/aibrix/aibrix/runtime/downloaders.py`) and is unaffected. `huggingface://`/`hf://` URLs and pre-mounted local paths continue to work unchanged.

The scheme check generalizes to any URL containing `://` (not just `s3`/`gcs`/`tos`) per review feedback, and `hf://` (the alias `ValidateArtifactURL` already allows) is now stripped the same way as `huggingface://` so it doesn't incorrectly hit the new check. The error message was reworded to state that the engine can't fetch the scheme directly and to list the actual alternatives: enabling the runtime sidecar (annotation *and* the controller's `--enable-runtime-sidecar` flag), a `huggingface://`/`hf://` URL, or a pre-mounted local path.

**Validation**
- `go build ./...` passes.
- `go vet ./pkg/controller/modeladapter/...` passes.
- `gofmt -l` on changed files reports nothing.
- The artifact-URL scheme check now runs in `LoadAdapter` before the list-models HTTP call, not just inside `loadAdapterCall`, so a `ModelAdapter` the direct path can never load fails immediately instead of after a pointless round trip to the pod. Extracted the check into a standalone `resolveDirectPathArtifactURL` helper so both call sites share it.
- `TestLoadAdapter`'s `s3://` case now asserts the specific error substring and that neither the list-models nor the load-adapter endpoint is ever called, instead of only asserting `wantErr: true`. Added an `hf://` case to `TestExtractHuggingFacePath` covering the new alias.
- Wrote a temporary scratch test (run, then removed — not part of this commit) driving `LoadAdapter` end-to-end via `config.RuntimeConfig{DebugMode: true}` against a real `httptest`-backed listener on `localhost:30081` (avoiding the port-8000 conflict below) to confirm: an `s3://` URL is rejected with the new message and neither endpoint is hit; a separate scratch test against `loadAdapterCall` directly confirmed an `hf://` URL resolves and reaches the engine like `huggingface://`, and a pre-mounted local path (`/models/...`) still passes through unchanged.
- Could **not** execute the full `TestLoadAdapter` table locally in this sandbox: `mockServer()` hardcodes `127.0.0.1:8000`, and this machine has an unrelated process already bound to port 8000, so every subtest in that table fails with "address already in use" regardless of this change (confirmed via `lsof -i :8000`). A maintainer separately suggested switching `mockServer()` to `httptest.NewServer` for a dynamic port; I did not make that change here because `BuildURLs` hardcodes the target port from `config.RuntimeConfig`, so every existing test case in this file relies on the mock listening on that fixed port — switching to a dynamic port would need a broader change to how the test wires ports to `BuildURLs`, which felt out of scope for this fix and risked breaking unrelated passing cases. CI runs in a clean container and is unaffected by the local port conflict.
- Could **not** run `make lint`/golangci-lint locally due to an unrelated local toolchain/Go-version mismatch; used `go vet` + `gofmt` as a substitute. `make lint` runs on the pinned Go version in CI.
- No generated code, CRDs, or manifests were touched.

## Related Issues
Relates to #766 (does not resolve it — see note above).

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
    <li>[ ] Documentation updated to reflect changes (if applicable) — not applicable, no user-facing docs describe this direct-path behavior</li>
</ul>

</details>

### DIFF
--- a/pkg/controller/modeladapter/lora_client.go
+++ b/pkg/controller/modeladapter/lora_client.go
@@ -259,11 +259,12 @@ func (c *loraClient) loadAdapterCall(ctx context.Context, url string, instance *
 				klog.ErrorS(err, "Invalid artifact URL", "artifactURL", artifactURL)
 				return err
 			}
-		} else if strings.HasPrefix(artifactURL, "s3://") || strings.HasPrefix(artifactURL, "gcs://") || strings.HasPrefix(artifactURL, "tos://") {
-			// These schemes need the artifact downloaded before the inference engine can
-			// load it; the engine only understands local paths and HuggingFace repo ids,
-			// so forwarding the raw URL causes it to be misread as a HuggingFace repo id
-			// (e.g. HFValidationError). Fail fast with an actionable error instead.
+		} else if strings.Contains(artifactURL, "://") {
+			// Any remaining scheme (s3://, gcs://, tos://, oss://, https://, ...) needs the
+			// artifact downloaded before the inference engine can load it; the engine only
+			// understands local paths and HuggingFace repo ids, so forwarding the raw URL
+			// causes it to be misread as a HuggingFace repo id (e.g. HFValidationError).
+			// Fail fast with an actionable error instead.
 			err := fmt.Errorf("artifact URL %q requires downloading via the aibrix runtime sidecar, "+
 				"which is not enabled for ModelAdapter %q; enable it with the "+
 				"\"model.aibrix.ai/sidecar-injection: true\" annotation on the target pod, "+
--- a/pkg/controller/modeladapter/lora_client.go
+++ b/pkg/controller/modeladapter/lora_client.go
@@ -259,6 +259,17 @@ func (c *loraClient) loadAdapterCall(ctx context.Context, url string, instance *
 				klog.ErrorS(err, "Invalid artifact URL", "artifactURL", artifactURL)
 				return err
 			}
+		} else if strings.HasPrefix(artifactURL, "s3://") || strings.HasPrefix(artifactURL, "gcs://") || strings.HasPrefix(artifactURL, "tos://") {
+			// These schemes need the artifact downloaded before the inference engine can
+			// load it; the engine only understands local paths and HuggingFace repo ids,
+			// so forwarding the raw URL causes it to be misread as a HuggingFace repo id
+			// (e.g. HFValidationError). Fail fast with an actionable error instead.
+			err := fmt.Errorf("artifact URL %q requires downloading via the aibrix runtime sidecar, "+
+				"which is not enabled for ModelAdapter %q; enable it with the "+
+				"\"model.aibrix.ai/sidecar-injection: true\" annotation on the target pod, "+
+				"or use a huggingface:// URL", artifactURL, instance.Name)
+			klog.ErrorS(err, "Unsupported artifact URL for direct engine loading", "ModelAdapter", klog.KObj(instance))
+			return err
 		}
 		// TODO: Add support for other URL transformations if needed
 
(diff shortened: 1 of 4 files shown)


pujitha24:
Good catch, that's a better check. Switched to `strings.Contains(artifactURL, "://")` so it covers any scheme the engine can't handle (oss://, cos://, https://, etc.) instead of just the three I'd hardcoded, and it's no longer case-sensitive. huggingface:// is still handled first and has its scheme stripped before this branch runs, so it's unaffected. Existing s3:// test case still passes under the new condition.

varungup90:
**Pull Request Comment**

Thanks for the contribution! While failing fast with diagnostics is an improvement over a cryptic HuggingFace validation error, a few points need addressing before merging:

* **Issue reference:** This doesn't actually resolve [[#766](https://github.com/vllm-project/aibrix/issues/766)](https://github.com/vllm-project/aibrix/issues/766), since the reporter's expected behavior is being able to load from S3 without sidecar configuration, and base model S3 paths crash the vllm container directly. Please drop `Fixes #766` so the issue isn't prematurely closed.
* **`hf://` prefix support:** `ValidateArtifactURL` allows `hf://`, but the Go direct path only handles `huggingface://`. As a result, `hf://org/model` hits the new scheme check and incorrectly tells the user to enable sidecar injection. Please ensure `hf://` is stripped the same way as `huggingface://`.
* **Error message accuracy:** The error text overstates the effect of the annotation alone, ignoring `--enable-runtime-sidecar` and valid local mounted paths (`/models/...`). It would be clearer to state that the engine cannot fetch the scheme directly, and list alternative workarounds (sidecar injection with the appropriate flag, `huggingface://` / `hf://`, or a pre-mounted local path).
* **Test assertions & execution:** The new `TestLoadAdapter` case only asserts `wantErr: true` and still hits `/v1/models` first because `loadAdapterCall` calls `getModels` prior to checking. Moving the scheme check to the top of the `!useSidecar` branch and asserting the specific error text / uncalled mock endpoint would make the test robust. Additionally, switching `mockServer()` to use `httptest.NewServer` would fix the hardcoded port 8000 conflict so tests can run locally.

varungup90 (approved):

Issue #832, opened 2025-03-09
Some prompts with special character fail the benchmark script


Jeffwan:
### 🐛 Describe the bug

This is a follow up issue of https://github.com/vllm-project/aibrix/issues/783

Seems some specific prompt data will fail the test

```
python benchmark_serving.py --backend vllm  --model deepseek-ai/deepseek-r1 --trust-remote-code --served-model-name deepseek-r1-671b --base-url http://localhost:8888 --endpoint /v1/completions --num-prompts 100 --request-rate 2 --metric_percentiles '50,90,95,99' --goodput ttft:1000 tpot:100 --max-concurrency 200 --random-input-len 2048 --random-output-len 200 --dataset-name random --ignore-eos
WARNING 03-09 14:32:12 _custom_ops.py:19] Failed to import from vllm._C with ModuleNotFoundError("No module named 'vllm._C'")
INFO 03-09 14:32:12 importing.py:10] Triton not installed; certain GPU-related functions will not be available.
Namespace(backend='vllm', base_url='http://localhost:8888', host='127.0.0.1', port=8000, endpoint='/v1/completions', dataset_name='random', dataset_path=None, max_concurrency=200, model='deepseek-ai/deepseek-r1', tokenizer=None, best_of=1, use_beam_search=False, num_prompts=100, logprobs=None, request_rate=2.0, burstiness=1.0, seed=0, trust_remote_code=True, disable_tqdm=False, profile=False, save_result=False, metadata=None, result_dir=None, result_filename=None, ignore_eos=True, percentile_metrics='ttft,tpot,itl', metric_percentiles='50,90,95,99', goodput=['ttft:1000', 'tpot:100'], sonnet_input_len=550, sonnet_output_len=150, sonnet_prefix_len=200, sharegpt_output_len=None, random_input_len=2048, random_output_len=200, random_range_ratio=1.0, random_prefix_len=0, hf_subset=None, hf_split=None, hf_output_len=None, tokenizer_mode='auto', served_model_name='deepseek-r1-671b', lora_modules=None)
Starting initial single prompt test run...
Traceback (most recent call last):
  File "/Users/bytedance/workspace/vllm/benchmarks/benchmark_serving.py", line 1314, in <module>
    main(args)
  File "/Users/bytedance/workspace/vllm/benchmarks/benchmark_serving.py", line 950, in main
    benchmark_result = asyncio.run(
[...]

PR #2655, opened 2026-09-02
[Bug] Fix benchmark prompts silently exceeding target token length

pujitha24:
## Pull Request Description

`generate_synthetic_prompt` and `adjust_prompt_length` in
`benchmarks/generator/dataset_generator/synthetic_prompt.py` build a prompt of
a target token length by truncating a token-id list and calling
`tokenizer.decode()`, then send the resulting **text** as the prompt (as a
real client/server would). Decoding a truncated token sequence and
re-encoding it is not guaranteed to round-trip to the same token count for
BPE/sentencepiece tokenizers, so the text handed back could silently be
longer than the requested `target_token_length` once the serving engine
re-tokenizes it. This is the same class of bug as the reported symptom (a
benchmark run requested `--random-input-len 2048` but the engine logged the
actual prompt as 2104 tokens and rejected it as too long), though see the
scope note below.

This PR adds a `_truncate_to_target_length` helper that truncates, decodes,
and — if the re-encoded length still overshoots — keeps dropping the last
token and re-decoding until the text's re-encoded length fits (or the id list
is exhausted). Both call sites now use it, and `generate_synthetic_prompt`
also recomputes `token_count` from the final (possibly further-truncated)
text instead of returning the stale pre-truncation count.

`adjust_prompt_length` also had a separate bug flagged in review: its
pad/truncate branches were `if`/`elif`, so a prompt padded past
`target_token_length` in the padding loop was never truncated back down
afterwards. This is now two independent `if` checks (matching
`generate_synthetic_prompt`'s existing structure), and truncation now
operates on the (possibly padded) `adjusted_prompt` instead of the original
`prompt`. Added a regression test
(`test_truncates_after_padding_overshoots_target`) that fails against the old
`elif` logic.

I verified the round-trip mismatch is real using a HuggingFace `gpt2`
tokenizer: decoding 200 random 50-token sequences and re-encoding them
produced a token-count mismatch in 139/200 cases (up to +7 tokens over
target). After the fix, 80+ trials of `generate_synthetic_prompt` and
several trials of `adjust_prompt_length` against the same tokenizer showed
zero overshoots and zero stale-count mismatches.

**Scope:** this only fixes aibrix's own benchmark prompt generator. The
original issue's traceback and reported log (`U+FFFD` replacement
characters, mixed-script gibberish) come from vLLM's upstream
`benchmark_serving.py --dataset-name random` path
(`/workspace/vllm/benchmarks/benchmark_serving.py`), which is not vendored in
this repository and is not touched by this PR — that upstream path already
has its own `gen_prompt_decode_to_target_len` helper. This PR instead fixes
the same *class* of round-trip-overshoot bug in AIBrix's own synthetic
prompt/workload generator (used by `benchmarks/benchmark.py`). It does not
resolve the linked issue's original report, so it should not auto-close it.

Also note: `benchmarks/` is not currently wired into any CI workflow in
`.github/workflows/` (only `python/aibrix/**` is covered by
`python-aibrix-tests.yml`), so there is no lint/test job for this change to
mirror. Validation below was run manually.

## Related Issues
Related to #832

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>

Related to #832

### DIFF
--- a/benchmarks/generator/dataset_generator/synthetic_prompt.py
+++ b/benchmarks/generator/dataset_generator/synthetic_prompt.py
@@ -1,6 +1,9 @@
+import logging
 import random
-from typing import Tuple
+from typing import List, Tuple
 from transformers import PreTrainedTokenizer
+
+logger = logging.getLogger(__name__)
 
 # A collection of realistic text templates for generating prompts
 REALISTIC_TEMPLATES = [
@@ -65,7 +68,7 @@ LOCATIONS = [
 
 
 def _truncate_to_target_length(tokenizer: PreTrainedTokenizer,
-                                token_ids: list,
+                                token_ids: List[int],
                                 target_token_length: int) -> Tuple[str, int]:
     """
     Truncate token ids to target_token_length and decode to text, making sure
@@ -81,11 +84,18 @@ def _truncate_to_target_length(tokenizer: PreTrainedTokenizer,
     """
     truncated_ids = token_ids[:max(0, target_token_length)]
     text = tokenizer.decode(truncated_ids, skip_special_tokens=True)
-    encoded = tokenizer.encode(text)
+    encoded = tokenizer.encode(text, add_special_tokens=True)
     while truncated_ids and len(encoded) > target_token_length:
         truncated_ids = truncated_ids[:-1]
         text = tokenizer.decode(truncated_ids, skip_special_tokens=True)
-        encoded = tokenizer.encode(text)
+        encoded = tokenizer.encode(text, add_special_tokens=True)
+    if not truncated_ids and len(encoded) > target_token_length:
+        logger.warning(
+            "_truncate_to_target_length drained to an empty string but the "
+            "re-encoded length (%d) still exceeds target_token_length (%d); "
+            "the tokenizer likely adds special tokens even for empty text.",
+            len(encoded), target_token_length,
+        )
     return text, len(encoded)
 
 
@@ -204,8 +214,9 @@ def adjust_prompt_length(tokenizer: PreTrainedTokenizer,
             ]
             adjusted_prompt += random.choice(additional_content)
             token_count = len(tokenizer.encode(adjusted_prompt))
-    elif token_count > target_token_length:
-        adjusted_prompt_tokenized = tokenizer.encode(prompt)
+
+    if token_count > target_token_length:
+        adjusted_prompt_tokenized = tokenizer.encode(adjusted_prompt)
         adjusted_prompt, _ = _truncate_to_target_length(tokenizer, adjusted_prompt_tokenized, target_token_length)
     return adjusted_prompt
     
--- a/benchmarks/generator/dataset_generator/synthetic_prompt.py
+++ b/benchmarks/generator/dataset_generator/synthetic_prompt.py
@@ -66,7 +66,7 @@ LOCATIONS = [
 
 def _truncate_to_target_length(tokenizer: PreTrainedTokenizer,
                                 token_ids: list,
-                                target_token_length: int) -> str:
+                                target_token_length: int) -> Tuple[str, int]:
     """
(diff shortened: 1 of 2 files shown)


varungup90:
**Blocker:** Do not close issue #832. Using `Fixes #832` or `Resolves: #832` will automatically close an issue that this change does not actually fix.

* **Issue Scope:** The linked report is for vLLM’s `benchmark_serving.py --dataset-name random` (traceback located under `/workspace/vllm/benchmarks/`). The engine log shows mixed-script gibberish with `U+FFFD` replacement characters from random vocabulary IDs decoded to text—not AIBrix’s English templates. It requested 2048 tokens while the engine counted 2104; upstream vLLM already provides `gen_prompt_decode_to_target_len` for that path.
* **Linkage:** The PR description admits this root cause is out of scope. Please update the issue linkage to **`Related to #832`**.

---

### Core Bug: `adjust_prompt_length` Overshoots After Padding

`generate_synthetic_prompt` pads and then truncates using a separate `if` block. In contrast, `adjust_prompt_length` uses `if / elif`, which prevents a short prompt that grows past its target during padding from ever being truncated:

`synthetic_prompt.py` (Ln 196–209):

```python
if token_count < target_token_length:
    while token_count < target_token_length:
        ...
        adjusted_prompt += random.choice(additional_content)
        token_count = len(tokenizer.encode(adjusted_prompt))
elif token_count > target_token_length:
    adjusted_prompt_tokenized = tokenizer.encode(prompt)
    adjusted_prompt, _ = _truncate_to_target_length(...)

```

**Required Changes:**

1. **Fix Truncation Logic:** Drop the `elif` and truncate whenever the final text ends up over length.
2. **Encode the Correct Prompt:** When truncating, encode `adjusted_prompt` rather than the original `prompt`.
3. **Add Tests:** Add a unit test starting with a prompt shorter than the target, asserting that the padded result remains `<= target`.

---

### Test Coverage Updates

* `generate_synthetic_prompt` does not currently hit this bug in test execution.
* `test_returned_token_count_matches_returned_prompt` uses `_WordTokenizer`, which round-trips 1:1 and masks the bug.
* `AdjustPromptLengthTest` already uses `_ExpandingTokenizer`. Update `test_returned_token_count_matches_returned_prompt` to use `_ExpandingTokenizer` as well, and assert `token_count == len(tokenizer.encode(prompt))` against it to catch the stale-count bug noted in the description.
* *Note:* `benchmarks/` is not currently included in CI (which only runs against `python/aibrix/**`). This matches existing generator tests and is worth addressing in a follow-up PR, but is not a merge blocker.

---

### Helper Nits (`synthetic_prompt.py`, Ln 82–89)

```python
truncated_ids = token_ids[:max(0, target_token_length)]
text = tokenizer.decode(truncated_ids, skip_special_tokens=True)
encoded = tokenizer.encode(text)
while truncated_ids and len(encoded) > target_token_length:
    ...
return text, len(encoded)

```

* **Explicit Special Tokens:** `encode(text)` defaults to `add_special_tokens=True` while `decode` strips them. This aligns with how token counting is handled elsewhere in this file, but pass `add_special_tokens` explicitly so future updates do not cause a desync (vLLM’s helper does this explicitly).
* **Empty Drain Handling:** If every non-empty decode continues to re-encode over target, the loop can drain down to `""` and still return `len(encoded) > target` (e.g., if `encode("")` yields `[bos]`). While unlikely for standard tokenizers, adding a log on an empty drain clarifies the contract.
* **Overshoot Strategy:** Linear shrink-by-one is acceptable for typical BPE overshoot (a few tokens). No need to implement vLLM's retry-and-pad loop unless landing exactly on target (`== target`) rather than `<= target` becomes a requirement.
* **Type Annotations:** Annotate `token_ids: list` as `List[int]`.

varungup90 (approved):

PR #2656, opened 2026-09-02
ci: improve workflow check display

googs1025:
## Summary
- split Go unit, race, and integration checks into separate CI jobs
- clarify workflow job names for Python and release publishing checks
- make installation test cleanup and artifact deletion run on failure paths

## Test Plan
- git diff --check
- ruby -e 'require "yaml"; ARGV.each { |f| YAML.load_file(f); puts "ok #{f}" }' .github/workflows/*.yml .github/workflows/*.yaml
- make -n test ENVTEST=true
- make -n test-race-condition ENVTEST=true
- make -n test-integration ENVTEST=true

### DIFF
--- a/.github/workflows/chart-ci.yml
+++ b/.github/workflows/chart-ci.yml
@@ -50,6 +50,7 @@ jobs:
           retention-days: 1
 
   test:
+    name: Helm Chart Install Test
     needs: [build-images]
     runs-on: ubuntu-latest
     steps:
--- a/.github/workflows/installation-tests.yml
+++ b/.github/workflows/installation-tests.yml
@@ -102,6 +102,7 @@ jobs:
           retention-days: 1
 
   installation-test:
+    name: Installation E2E Tests
     needs: [build-images]
     runs-on: ubuntu-latest
     steps:
@@ -206,7 +207,7 @@ jobs:
           kind create cluster --name installation-test --config=./hack/ci/kind-config.yaml
           kubectl cluster-info --context kind-installation-test
 
-      - name: Load image into Kind
+      - name: Load images into kind
         run: |
           set -euo pipefail
           sha="${{ github.sha }}"
@@ -245,7 +246,7 @@ jobs:
             aibrix/inplace-e2e:v2 \
             --name installation-test
 
-      - name: Deploy controller with the built image
+      - name: Deploy AIBrix control plane
         run: |
           kubectl apply -k config/dependency --server-side
           kubectl apply -k config/crd --server-side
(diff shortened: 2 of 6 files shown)


varungup90:
### Suggested Review Comment

```markdown
Thanks for working on this! Overall, splitting the CI jobs improves visibility significantly. 

Here are a few findings and suggestions based on the latest changes:

### 🚨 Should Fix

#### 1. `delete-artifacts` still leaves tarballs behind and risks false-negative job failures
* **Incomplete Cleanup:** `build-images` uploads five distinct artifacts (`controller-manager-image`, `gateway-plugins-image`, `runtime-image`, `metadata-service-image`, and `vllm-mock-image`), but the cleanup job currently only deletes `vllm-mock-image`.
* **Error Handling on Skip/Failure:** `if: always()` runs even if `installation-test` is skipped (e.g., if a `build-images` matrix job fails). Since `geekyeggo/delete-artifact@v5` defaults to `failOnError: true`, unexpected deletion errors could mark the cleanup job red on top of an already failed build.

**Suggested Fix:**
```yaml
delete-artifacts:
  name: Delete image artifacts
  needs: [build-images, installation-test]
  if: ${{ always() && !cancelled() }}
  runs-on: ubuntu-latest
  steps:
    - uses: geekyeggo/delete-artifact@v5
      with:
        name: '*-image'
        failOnError: false

```

---

#### 2. Kind cleanup can fail if setup dies early

If the workflow fails before the "Install kind" step (e.g., during artifact download, disk cleanup, or cache restoration), `kind delete cluster` will fail with `kind: command not found`. Because GitHub-hosted `ubuntu-latest` VMs are ephemeral and discarded anyway, this step shouldn't introduce noisy job failures.

**Suggested Fix:**

```yaml
- name: Clean up
  if: always()
  continue-on-error: true
  run: kind delete cluster --name installation-test || true

```

---

### 💬 Non-blocking / Considerations

* **Incomplete Display Names:** To fully align with the PR objective:
* `check-coverage` and `delete-artifacts` are still missing explicit `name:` keys, so GitHub defaults to displaying their raw job IDs.
* The `test` job in `chart-ci.yml` shares the same issue.


* **Status Check Renaming Risk (`test` → `Go Unit Tests`):** Renaming this check changes the context name sent to GitHub. If classic branch protection rules require a status check specifically named `test`, this change will block merges until updated. Please verify repository branch protection settings.
* **Untested Python Workflow Rename:** The Python workflows trigger specifically on `python/aibrix/**` and `python/aibrix_kvcache/**`. Because they don't include their own workflow paths in `on.push.paths`/`on.pull_request.paths` (unlike `apps-chat-ci.yml`), these changes won't run on this PR and will remain untested until the next Python file change. Consider adding the workflow paths to the trigger list.
* **Runner Overhead:** The previous combined test job ran unit, race, and integration tests on a single VM in ~13 minutes. While splitting them into separate jobs is the right move for actionable signals, keep in mind each job now duplicates overhead (`actions/checkout`, `setup-go`, `apt`, and `make` tooling setup), roughly tripling total runner minutes for that workflow.

```

```

googs1025:
I addressed the actionable feedback in the latest push:
  - delete all image artifacts with `*-image` and avoid cleanup-only failures
  - make kind cleanup non-blocking on early setup failures
  - add missing display names for coverage/artifact cleanup/chart install checks
  - include Python workflow files in their own path filters so these changes are exercised by this PR

  I’m going to merge this now to unblock the CI visibility improvement. If follow-up issue shows up in CI or branch protection after merge, I’ll handle it in a follow-up PR.

PR #2661, opened 2026-09-03
[API] Reject StormService /scale above 1 in Pooled mode

### LINKED ISSUE #2449: Make StormService deployment mode explicit

roshpr:
## Pull Request Description

#2450 and #2617 added optional `StormServiceSpec.Mode` and consume it in the controller and PodAutoscaler. Create/update admission already rejects a declared `mode: Pooled` with `spec.replicas > 1`.

That check does not run for the `/scale` subresource. `kubectl scale`, HPA, and other external autoscalers write `spec.replicas` through `/scale`, so a pooled StormService could still be scaled above one RoleSet.

This change registers a validating webhook for `stormservices/scale`. The handler loads the parent StormService, applies the proposed replica count, and reuses `validateStormServiceMode`:

- Declared `mode: Pooled` cannot set replicas > 1 via `/scale`. The admission error tells operators to scale roles through `spec.template.spec.roles[].replicas` instead.
- Declared `mode: Replica` can still scale `spec.replicas`.
- Omitted `spec.mode` is not pinned to inferred Pooled, so existing replica scaling continues to work.

Helm and kubebuilder webhook manifests include the new `vstormservicescale.kb.io` rule.

### Tests

- Unit: reject/allow table for `/scale` (Pooled > 1, Pooled == 1, Replica > 1, omitted mode > 1), plus malformed Scale and missing StormService.
- Integration: envtest `/scale` subresource reject for declared Pooled, allow for declared Replica, allow when mode is omitted.

Verified locally:

```
go test -count=1 ./pkg/webhook/
ok  	github.com/vllm-project/aibrix/pkg/webhook	0.089s

make test-integration-webhook
Ran 84 of 84 Specs in 7.265 seconds
SUCCESS! -- 84 Passed | 0 Failed
```

## Related Issues
Part of #2449. This is the leftover `/scale` subresource enforcement from the follow-up list after #2450 and #2617. It does not close #2449; the original issue still has remaining items such as reporting the resolved mode in status.

### DIFF
--- a/config/webhook/manifests.yaml
+++ b/config/webhook/manifests.yaml
@@ -210,3 +210,22 @@ webhooks:
     resources:
     - stormservices
   sideEffects: None
+- admissionReviewVersions:
+  - v1
+  clientConfig:
+    service:
+      name: webhook-service
+      namespace: system
+      path: /validate-orchestration-aibrix-ai-v1alpha1-stormservice-scale
+  failurePolicy: Ignore
+  name: vstormservicescale.kb.io
+  rules:
+  - apiGroups:
+    - orchestration.aibrix.ai
+    apiVersions:
+    - v1alpha1
+    operations:
+    - UPDATE
+    resources:
+    - stormservices/scale
+  sideEffects: None
--- a/dist/chart/templates/webhook/webhooks.yaml
+++ b/dist/chart/templates/webhook/webhooks.yaml
@@ -208,6 +208,25 @@ webhooks:
     service:
       name: aibrix-webhook-service
       namespace: {{ .Release.Namespace }}
+      path: /validate-orchestration-aibrix-ai-v1alpha1-stormservice-scale
+  failurePolicy: Ignore
+  name: vstormservicescale.kb.io
+  rules:
+    - apiGroups:
+        - orchestration.aibrix.ai
+      apiVersions:
+        - v1alpha1
+      operations:
+        - UPDATE
+      resources:
+        - stormservices/scale
+  sideEffects: None
+- admissionReviewVersions:
+    - v1
+  clientConfig:
+    service:
+      name: aibrix-webhook-service
+      namespace: {{ .Release.Namespace }}
       path: /validate-apps-v1-deployment
   failurePolicy: Ignore
   name: vedeployment.aibrix.ai
--- a/docs/source/designs/aibrix-stormservice.rst
+++ b/docs/source/designs/aibrix-stormservice.rst
@@ -39,7 +39,7 @@ Stormservice supports two deployment modes: **Replica Mode** and **Pooled Mode**
 .. note::
     1. These two modes are mutually exclusive. The mode is declared through the `stormservice.spec.mode` field, which accepts `Replica` or `Pooled`.
     2. `spec.mode` is optional and is not defaulted. When it is omitted the mode is inferred for backward compatibility from `stormservice.spec.replicas`: replica mode when `replicas > 1`, otherwise pooled mode.
-    3. When `spec.mode` is set to `Pooled`, `spec.replicas` must stay at `1`; roles are scaled through `spec.template.spec.roles[].replicas`.
+    3. When `spec.mode` is set to `Pooled`, `spec.replicas` must stay at `1`; roles are scaled through `spec.template.spec.roles[].replicas`. The validating webhook enforces this on create/update and on the `/scale` subresource (`kubectl scale`, HPA, and other external autoscalers).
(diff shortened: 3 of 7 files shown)


googs1025:
@roshpr thanks for this fix. plz fix the conflict

varungup90 (approved):

roshpr:
Hi @googs1025 @varungup90 — thanks for the review. This is approved and CI is green; I don’t have merge rights on the repo. Could one of you merge when you have a moment?

PR #2664, opened 2026-09-03
[Bug] Fix /v1/embeddings input limits the gateway cannot measure

### LINKED ISSUE #2663: /v1/embeddings rejects all inputs over 8192 tokens
### 🐛 Describe the bug

It ignores the `truncate_prompt_tokens` so regardless if the real embedding model can ingest only 512 tokens or 32768 tokens, aibrix always checks against the hardcoded 8192 threshold. 

Now, it is technically impossible (or too expensive) to accurately calculate the input tokens based on the input text because we don't know at this point which model is gonna be routed to and which tokenizer that given model uses. Based on the input language and charset, the token count as calculated (by the cl100k_base tokenizer in aibrix) might be +- ~50% off.

The vllm itself handles the `truncate_prompt_tokens` (including `-1` as the model's max) accurately; and/or trims or rejects inputs which are too long. There is little value for the gateway to make this check, especially when it can't be accurate.

The proposal (and a following PR) is to drop this check alltogether to keep it simple.

### Steps to Reproduce

1. send a payload to `/v1/embeddings` with the input like 30 kilobytes - to a model which can ingest 32k input tokens (such as `qwen3-embedding-4b`)
2. observe a HTTP 400 being returned from aibrix (because it always compare to hardcoded 8192 tokens)

notes:
- it does not matter if the payload says `truncate_prompt_tokens:23768` or `truncate_prompt_tokens:512` or `truncate_prompt_tokens:-1`; this parameter is completely ignored.

### Expected behavior

Aibrix should honor the model's real cap - ideally by passing the payload to the model an let the model to decide on succeding, trimming or erroring out.

### Environment

aibrix 0.7.0

tomhol:
## Pull Request Description

`/v1/embeddings` requests are rejected at the gateway whenever a tiktoken (`cl100k_base`) estimate of the input exceeds `MaxInputTokensPerModel` (8192), before the request ever reaches the inference engine.

There are two independent problems with that check.

**`truncate_prompt_tokens` is invisible to it.** `embeddingReqMinimal` never parses the field, so a client that explicitly asks the backend to truncate oversized input receives a 400 that is byte-for-byte identical to one that did not ask. vLLM implements the parameter (including `-1` for "the model's maximum"), but the request never gets there.

**The units do not match.** The estimate is in cl100k tokens, while the limit that matters is denominated in the target model's own tokens. For sentencepiece models such as `bge-m3` or `multilingual-e5`, cl100k over-counts non-English text by roughly 2x, so valid requests are rejected with a misleading message. Measured on Czech prose, cl100k yields ~2.18 chars/token where XLM-R style tokenizers typically run 3.5–4.5.

The constant is also global despite its name, so it cannot match every served model simultaneously:

| model | real limit | gateway enforces | effect |
| --- | --- | --- | --- |
| `multilingual-e5-large-instruct` | 512 | 8192 | 16x too loose — guard protects nothing |
| `bge-m3` | 8192 | 8192 | same number, wrong unit |
| `qwen3-embedding-4b` | 32768 | 8192 | 4x too tight — rejects valid input |

There is no correct value to raise it to, because the gateway cannot know which model's limit applies. `validateRequestBody` runs before the pod list and engine label are resolved (`gateway_req_body.go`), so neither the addressed model's tokenizer nor its context length is reachable at that point.

### Change

**1. Drop the token-count checks.** Removed from all three call sites — `validateStringInputs`, `validateTokenInputs`, and `validateEmbeddingContentParts` (multimodal text parts). The backend already enforces the real context length using the model's own tokenizer and reports it accurately, and it is the only component that has both pieces of information. `MaxInputTokensPerModel` and `MaxTotalTokens` are dropped as unused.

**2. Correct `MaxArrayDimensions` to bound what it describes** (per @varungup90's review). OpenAI documents 2048 as `maxItems` for the input array — the batch size — but it was being compared against `len(tokens)`, the number of tokens in one pre-tokenized input. Because 2048 < 8192 that comparison always fired first, so pre-tokenized input was effectively capped at 2048 tokens: content a model accepts as a string was rejected at 2049 tokens when sent as an array. The check now applies to `len(inputs)` and `len(tokenArrays)` in both input paths, which is a property of the request the gateway can measure exactly.

Note this makes the string path slightly *stricter*: `validateStringInputs` previously had no dimensions check at all, so a >2048-string batch used to pass. It is now rejected, matching `maxItems`.

No request-parsing change is needed: `routingCtx.ReqBody` is forwarded verbatim in the `BodyMutation`, so `truncate_prompt_tokens` already reaches the backend once the gateway stops rejecting the request first. Removing the token-count check also takes a full tiktoken pass over every embeddings input off the request path.

### Reproduction (before this change)

`POST /v1/embeddings` with ~86 000 chars of Czech prose to a 512-token embedding model:

| `truncate_prompt_tokens` | result |
| --- | --- |
| `400` | 400 `input exceeds max tokens per model (8192), estimated tokens: 39480` |
| `-1` | 400, byte-for-byte identical |
| omitted | 400, byte-for-byte identical |

The identical response in all three cases is the symptom. With input small enough to clear the guard, vLLM honours both `-1` and an explicit value and returns 200 — the feature works, it is simply unreachable.

### Testing

`make test` passes: 63 packages ok, 0 failures. `go vet ./pkg/...` and `gofmt -l` clean.

- Oversized input is asserted to be forwarded rather than rejected, with `truncate_prompt_tokens` absent, set to `-1`, and set to an explicit value — the three cases must no longer collapse into an identical gateway-side 400.
- A non-ASCII case covers the over-count described above.
- Cases that previously locked in the per-input 2048 token cap now assert acceptance (`long_single_token_array_is_accepted`, `long_token_arrays_are_accepted`).
- New cases assert a batch beyond `maxItems` is rejected for both string and pre-tokenized inputs (`array_of_strings_exceeding_max_batch_size`, `batch_exceeding_max_dimensions`).

### Open question for reviewers

`validateEmbeddingContentParts` did **not** get the batch bound, because whether its array is one multimodal input's content parts or a batch of separate inputs is ambiguous in that code path: the OpenAI-style `{"type": ..., "text": ...}` shape reads like content parts of a single input, while sglang's flat `{"text": ...}` / `{"image": ...}` shape reads like a batch. Applying `maxItems` under the wrong reading would reject valid requests. Happy to add it if you can confirm which it is.

## Related Issues

Resolves: #2663

---

### Submission Checklist

- [x] Documentation updated to reflect changes (n/a — no user-facing docs cover this guard)

### DIFF
--- a/pkg/plugins/gateway/types.go
+++ b/pkg/plugins/gateway/types.go
@@ -95,9 +95,11 @@ const (
 
 	// Embedding Constraints
 	// https://github.com/openai/openai-go/blob/main/embedding.go#L126
-	MaxInputTokensPerModel = 8192
-	MaxTotalTokens         = 300000
-	MaxArrayDimensions     = 2048
+	//
+	// The token-count limits from that reference (8192 per input, 300000 per
+	// batch) are not enforced here: they are denominated in the target model's
+	// tokens, which the gateway cannot measure. See validateStringInputs.
+	MaxArrayDimensions = 2048
 
 	// Request Paths
 	PathChatCompletions     = "/v1/chat/completions"
--- a/pkg/plugins/gateway/util.go
+++ b/pkg/plugins/gateway/util.go
@@ -296,17 +296,15 @@ type embeddingContentPart struct {
 	Video    string          `json:"video"`
 }
 
-// validateEmbeddingContentParts validates a multimodal embeddings `input` array. Only
-// "text" parts are tokenized and counted against MaxInputTokensPerModel / MaxTotalTokens;
-// image/video parts are excluded because the gateway's text tokenizer cannot estimate
-// their cost, and the backend's own multimodal processor is responsible for that
-// accounting.
+// validateEmbeddingContentParts validates the shape of a multimodal embeddings
+// `input` array. Text length is not checked, for the reasons given on
+// validateStringInputs. Image and video parts were never counted anyway, since
+// the gateway's text tokenizer cannot estimate their cost.
 func validateEmbeddingContentParts(parts []embeddingContentPart) error {
 	if len(parts) == 0 {
 		return errors.New("input array cannot be empty")
 	}
 
-	totalEstimatedTokens := 0
 	for i, part := range parts {
 		typ := part.Type
 		if typ == "" {
@@ -328,16 +326,6 @@ func validateEmbeddingContentParts(parts []embeddingContentPart) error {
 			if part.Text == "" {
 				return fmt.Errorf("input at index %d cannot be an empty string", i)
 			}
-			tokens, err := utils.TokenizeInputText(part.Text)
-			if err != nil {
-				return fmt.Errorf("failed to tokenize input for validation: %w", err)
-			}
-			estimatedTokens := len(tokens)
-			if estimatedTokens > MaxInputTokensPerModel {
-				return fmt.Errorf("input at index %d exceeds max tokens per model (%d), estimated tokens: %d",
-					i, MaxInputTokensPerModel, estimatedTokens)
-			}
-			totalEstimatedTokens += estimatedTokens
 		case "image_url":
 			if len(part.ImageURL) == 0 {
 				return fmt.Errorf("input at index %d is missing image_url", i)
@@ -357,11 +345,6 @@ func validateEmbeddingContentParts(parts []embeddingContentPart) error {
 		default:
 			return fmt.Errorf("input at index %d has unsupported content type %q", i, typ)
 		}
-	}
-
-	if totalEstimatedTokens > MaxTotalTokens {
-		return fmt.Errorf("total tokens across all inputs exceeds maximum (%d), estimated total: %d",
-			MaxTotalTokens, totalEstimatedTokens)
 	}
 
 	return nil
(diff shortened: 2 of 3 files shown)


varungup90:
@tomhol sign the PR

varungup90:
Nice catch on cleaning up the gateway-side token limits! 

One issue worth fixing (now or as a fast follow-up): `MaxArrayDimensions` (2048) is still applied to `len(tokens)`—the length of a single token array—rather than the number of inputs. OpenAI’s 2048 limit specifies the maximum batch size (`maxItems: 2048`), not the token count per input. 

Currently, after this PR, a long string can be successfully forwarded to a 32k model, but the exact same input sent as a pre-tokenized array still triggers a 400 error at 2049 tokens because the tests lock that in. Moving the check to `len(inputs)` / `len(tokenArrays)` and removing the comparison against `len(tokens)` would align this behavior correctly.

tomhol:
Good catch! Thank you. The current commit has the sign-off now. And I also addressed the "validateStringInputs" / "MaxArrayDimensions" problem. The PR description has been updated in that sense too.

varungup90 (approved):

PR #2666, opened 2026-09-04
[Misc] Add unit tests for scaling algorithm factory and constraints

stywzn:
## Pull Request Description

Adds unit tests for two previously untested functions in `pkg/controller/podautoscaler/algorithm`. No production code is changed.

Both functions were at 0% statement coverage. The inner math (`computeTargetReplicas` for KPA and APA) is already well covered at ~95%, but the surrounding factory and constraint helpers had no tests at all.

**`applyConstraints`** — clamps the computed replica count into the `[minReplicas, maxReplicas]` range. Test covers below-range, above-range, within-range, and both boundary values, so the `<` / `<=` behaviour of the min and max comparisons is pinned down.

**`NewScalingAlgorithm`** — the strategy factory. Test covers all three strategies (KPA, APA, HPA) plus the `default` fallback for an unrecognized strategy, each verified through `GetAlgorithmType()`.

Package statement coverage goes from 59.4% to 65.6%; both functions reach 100%.

Verified locally with `gofmt`, `go vet`, `golangci-lint`, and `go test`.

## Related Issues

N/A — no tracking issue. This chips away at the existing coverage gap, in line with the 90% total-coverage goal noted in `.github/workflows/lint-and-tests.yml`.

### DIFF
--- a/pkg/controller/podautoscaler/algorithm/algorithm_test.go
+++ b/pkg/controller/podautoscaler/algorithm/algorithm_test.go
@@ -1,0 +1,81 @@
+/*
+Copyright 2026 The Aibrix Team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package algorithm
+
+import (
+	"testing"
+
+	autoscalingv1alpha1 "github.com/vllm-project/aibrix/api/autoscaling/v1alpha1"
+)
+
+func TestApplyConstraints(t *testing.T) {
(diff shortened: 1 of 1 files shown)


stywzn:
Thanks for the review — both suggestions are addressed in 898589a0:

- `applyConstraints` failures now include the `min`/`max` values from the test case, so a failing case is diagnosable without cross-referencing the table.
- `NewScalingAlgorithm` now asserts the concrete type (`*KPAAlgorithm` / `*APAAlgorithm` / `*HPAAlgorithm`) in addition to `GetAlgorithmType()`, so the factory mapping is verified directly rather than through the type string alone.

On the `Go Integration Tests` failure: I believe it is unrelated to this PR. It failed in `test/integration/controller/kvcache_test.go:293` — `KVCache controller test > pod-triggered reconciliation` — with 70 of 71 specs passing. This PR only adds a unit test file under `pkg/controller/podautoscaler/algorithm`, which the integration suite does not load, and no production code is touched. The same job passed on #2664 about 90 minutes after this failure, so it looks intermittent rather than a regression.

Could someone re-run the job? And if that kvcache spec isn't already tracked as flaky, I'm happy to open an issue and look into it separately.

googs1025 (approved):
LGTM. 
thanks for add more test coverage 😄 
hope to participate and contribute more to the community.

PR #2668, opened 2026-09-04
[Misc] Add unit tests for KPA panic mode and HPA algorithm

stywzn:
## Pull Request Description

Adds unit tests for three previously untested functions in `pkg/controller/podautoscaler/algorithm`. No production code is changed.

`hpa.go` had no test file at all, and `KPAAlgorithm.shouldEnterPanicMode` was at 0% statement coverage.

**`shouldEnterPanicMode`** — covers the `StableValue <= 0` guard, including a negative value so the `<=` comparison is actually exercised. That guard matters: without it the ratio below would divide by zero and produce `NaN`, and since any `NaN` comparison is false, panic mode would be skipped exactly when there is no stable history to fall back on. Also covers the ratio above and below the panic threshold, plus the exact-threshold boundary that pins down `>` versus `>=`.

**`HPAAlgorithm.ComputeRecommendation`** — exercised with several replica counts (0, 1, 7) so the test distinguishes "echoes `CurrentReplicas`" from "returns a constant", and asserts the fixed `Confidence`, `Algorithm` and `ScaleValid` fields along with a nil error.

**`HPAAlgorithm.GetAlgorithmType`** — covered.

Package statement coverage goes from 59.4% to 61.8%; all three functions reach 100%.

Verified locally with `gofmt`, `go vet`, `golangci-lint` and `go test`.

## Related Issues

N/A — no tracking issue. Continues chipping away at the coverage gap, in line with the 90% total-coverage goal noted in `.github/workflows/lint-and-tests.yml`. Follows the same approach as #2666.

### DIFF
--- a/pkg/controller/podautoscaler/algorithm/hpa_test.go
+++ b/pkg/controller/podautoscaler/algorithm/hpa_test.go
@@ -1,0 +1,68 @@
+/*
+Copyright 2026 The Aibrix Team.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package algorithm
+
+import (
+	"context"
+	"testing"
+)
+
+func TestHPAAlgorithm_ComputeRecommendation(t *testing.T) {
+	// HPA scaling is delegated to the Kubernetes HPA controller, so the
+	// recommendation must echo the current replica count unchanged.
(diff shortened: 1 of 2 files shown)


googs1025 (approved):
LGTM thank you for the help

stywzn:
Pushed ea752b47 to fix the `Go Lint` failure.

The failure was `goconst`: `string "hpa" has 3 occurrences, make it a constant` at `hpa.go:45`. It only appears once this branch is merged with `main` — `goconst` counts string literals across the whole package, and #2666 (merged shortly before this run) added its own `"hpa"` literals to `algorithm_test.go`. Each branch stays under the threshold on its own, which is why this was not visible when the PR was opened.

The fix introduces a single `hpaAlgorithmType` constant in `hpa_test.go` and uses it for the four assertions. No production code is changed.

Verified locally against the merged state (`main` + this branch): `goconst` clean, `golangci-lint` clean for the package, `gofmt`, `go vet` and `go test` all pass.

Happy to move the constant into `hpa.go` instead if you would rather have the identifier defined next to the implementation — `kpa.go` and `apa.go` have the same literal pattern, so that could be a small follow-up cleanup.

PR #2669, opened 2026-09-04
[CI] Stabilize KVCache pod-triggered reconciliation tests

### LINKED ISSUE #2667: [Bug] Flaky KVCache pod-triggered reconciliation integration test
## Summary

`Go Integration Tests` has a flaky KVCache controller integration spec:

`KVCache controller test / pod-triggered reconciliation / reconciles the KVCache named by the Pod's identifier label`

The failure is at `test/integration/controller/kvcache_test.go:293`.

## Evidence

Observed at least twice in recent GitHub Actions runs:

- https://github.com/vllm-project/aibrix/actions/runs/33837843112/job/100921328903
- https://github.com/vllm-project/aibrix/actions/runs/33772035855/job/100704428793

Both failures summarize the same spec:

```text
[FAIL] KVCache controller test pod-triggered reconciliation [It] reconciles the KVCache named by the Pod's identifier label
/home/runner/work/aibrix/aibrix/test/integration/controller/kvcache_test.go:293
```

In the 2026-09-04 run, the suite reported:

```text
Ran 71 of 71 Specs
70 Passed | 1 Failed
```

In the 2026-09-03 run, the suite reported:

```text
Ran 62 of 62 Specs
61 Passed | 1 Failed
```

## What appears to fail

The failing line is the `consistentlyAbsent(...)` assertion after deleting the generated StatefulSet and before creating the labeled Pod:

```go
gomega.Expect(k8sClient.Delete(ctx, sts)).To(gomega.Succeed())
consistentlyAbsent(kv.Name, &appsv1.StatefulSet{})

gomega.Expect(k8sClient.Create(ctx, makePod(ns.Name, "labeled-pod", kv.Name))).To(gomega.Succeed())

eventuallyGet(kv.Name, &appsv1.StatefulSet{})
```

The log shows:

```text
[FAILED] Failed after 0.255s.
Expected <bool>: false to be true
```

So the StatefulSet is recreated during the quiet window, before the labeled Pod is created.

## Why this looks flaky

The test assumes that deleting the StatefulSet will not trigger any self-healing reconcile because StatefulSets are not in the KVCache controller's `Owns()` list. However, the KVCache controller does own Services:

```go
Owns(&corev1.Service{})
```

The initial KVCache reconcile creates or updates owned Services. A delayed owner-triggered reconcile from those Service events can still be pending when the test deletes the StatefulSet. If that pending reconcile runs during `consistentlyAbsent(...)`, it recreates the StatefulSet and the test fails before the Pod-trigger path is exercised.

The Pod watch/mapping itself appears to be registered:

```go
Watches(&corev1.Pod{},
    handler.EnqueueRequestsFromMapFunc(kvCacheRequestsForPod),
    builder.WithPredicates(podWithLabelFilter(constants.KVCacheLabelKeyIdentifier))).
```

## Suggested direction

The test likely needs to drain or stabilize any initial/owned-resource-triggered reconciles before deleting the StatefulSet, then create the labeled Pod and assert that the StatefulSet is recreated.

Alternatively, the spec can be rewritten to verify the Pod-event mapping without relying on the absence of all other pending reconcile events in the shared envtest manager.

yigitcan-ozturk:
## Summary

Fixes the flaky KVCache pod-triggered reconciliation integration specs reported in #2667.

The tests deleted the generated StatefulSet as soon as it appeared and then used an absence window to prove that only the subsequent Pod event could recreate it. However, the initial reconcile also creates an owned Service. A delayed owner-triggered reconcile from that Service can still be queued and recreate the StatefulSet during the absence window, causing a false failure.

This change waits for both distributed resources (StatefulSet + headless Service) to exist and remain stable for the existing quiet window before removing the unwatched StatefulSet. That gives initial Service-owner events time to drain before the Pod event under test.

## Scope

- Test-only change
- No controller/runtime behavior changes
- Applies the stabilization helper to the create, delete, unlabeled, and routing Pod integration specs that rely on StatefulSet absence as their signal

## Why this keeps the test meaningful

StatefulSets are still not watched via `Owns()`. After the initial owned-resource events are allowed to settle, deleting the StatefulSet remains the observable state change used to verify the Pod watch behavior.

Fixes #2667

### DIFF
--- a/test/integration/controller/kvcache_test.go
+++ b/test/integration/controller/kvcache_test.go
@@ -178,6 +178,28 @@ var _ = ginkgo.Describe("KVCache controller test", func() {
 		}, kvCacheQuietWindow, kvCacheInterval).Should(gomega.BeTrue())
 	}
 
+	// waitForDistributedKVCacheToSettle ensures the initial reconcile has
+	// created both distributed resources, then gives owner-triggered Service
+	// events a quiet window to drain before tests remove the unwatched
+	// StatefulSet. Without this stabilization, a delayed Service-owner reconcile
+	// can recreate the StatefulSet before the Pod event under test arrives.
+	waitForDistributedKVCacheToSettle := func(name string) {
+		ginkgo.GinkgoHelper()
+
+		serviceName := fmt.Sprintf("%s-headless-service", name)
+		eventuallyGet(name, &appsv1.StatefulSet{})
+		eventuallyGet(serviceName, &corev1.Service{})
+
+		gomega.Consistently(func() error {
+			statefulSetKey := types.NamespacedName{Namespace: ns.Name, Name: name}
+			if err := k8sClient.Get(ctx, statefulSetKey, &appsv1.StatefulSet{}); err != nil {
+				return err
+			}
+			serviceKey := types.NamespacedName{Namespace: ns.Name, Name: serviceName}
+			return k8sClient.Get(ctx, serviceKey, &corev1.Service{})
+		}, kvCacheQuietWindow, kvCacheInterval).Should(gomega.Succeed())
+	}
+
 	ginkgo.Context("backend selection", func() {
 		ginkgo.It("defaults to the vineyard backend when no backend annotation is set", func() {
 			kv := makeKVCache(ns.Name, "default-backend", "")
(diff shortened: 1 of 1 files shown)


varungup90:
Sign the PR

yigitcan-ozturk:
Signed off the PR commits with DCO (`Signed-off-by: Yiğitcan Öztürk <318303091+yigitcan-ozturk@users.noreply.github.com>`) and squashed the branch to a single signed commit. Thanks.

varungup90 (approved):

PR #2676, opened 2026-09-07
[Gateway] Restructure E2E test suites

### LINKED ISSUE #2675: [TEST] Restructure E2E tests and add PD protocol contract coverage
## Background

The current `test/e2e` package mixes controller and gateway cases. Gateway tests cover routing outcomes and client-visible responses, but PD e2e does not verify the HTTP request bodies that gateway sends to prefill/decode backends.

For vLLM, SGLang, and TRT-LLM this leaves important regressions undetected:

- vLLM SHFS/NIXL handoff fields can be lost or mutated.
- SGLang bootstrap fields and nested raw JSON preservation are not verified end-to-end.
- TRT-LLM context/generation handoff shape is not verified against its OpenAI HTTP form.
- Prefill failures cannot be correlated to decode behavior by request ID.

## Goal

Build a stable, Kind-based gateway compatibility e2e suite that validates AIBrix gateway PD HTTP handoff contracts without requiring GPUs, Neuron, EFA, or real NIXL transfer.

Each contract is an AIBrix compatibility commitment. Upstream engine or connector changes must add a new named contract rather than silently changing existing e2e expectations.

## Test Organization

Split e2e tests by behavior ownership:

```text
test/e2e/
  framework/
  gateway/{api,routing,pd}/
  controller/{modeladapter,modelclaim,roleset,autoscaling}/
  testdata/
```

`ModelClaim` wake-up and `ModelAdapter` backend reachability remain controller e2e: the controller owns the resource state transition, while gateway is a downstream check.

The shared `e2eframework` package will provide environment config, Kubernetes/OpenAI clients, unique request IDs, mock recorder queries through Kubernetes Pod proxy, wait helpers, and failure diagnostics.

## PD Contracts

The default gateway suite will validate four successful handoff contracts:

- `vllm-aibrix-shfs`: prefill receives `kv_transfer_params.do_remote_decode=true`; decode receives the merged response fields, including an opaque sentinel.
- `vllm-aibrix-nixl`: prefill does not receive the SHFS skeleton; decode receives the complete prefill response under `disagg_prefill_resp`, including a sentinel.
- `sglang-http`: both legs receive top-level `bootstrap_host`, `bootstrap_port`, and the same `bootstrap_room`; user-supplied nested JSON remains byte-identical.
- `trtllm-openai`: prefill receives `request_type=context_only`; decode receives `request_type=generation_only`, `disagg_request_id`, `first_gen_tokens`, `encoded_opaque_state`, and prompt token IDs.

## Mock App Requirements

Extend the OpenAI HTTP mock app with:

- A request recorder and `GET /debug/requests?request_id=...`.
- Raw body, parsed JSON, headers, path, pod, engine, role, sequence, timestamps, and request outcome per record.
- `MOCK_PD_CONTRACT` to select strict contract validation.
- Strict validator/response builders for SHFS, NIXL, SGLang, and TRT-LLM.
- A NIXL mock StormService with `model.aibrix.ai/kv-connector-type: nixl`.
- Request-scoped fault injection: `x-aibrix-mock-fail: prefill|decode` and `x-aibrix-mock-delay-ms: <milliseconds>`.
- Shared SGLang handoff state keyed by request ID/bootstrap room so decode can expose an explicit handoff error after async prefill failure.

The mock must reject invalid protocol bodies and record the rejection; it must not only log the mismatch and return a successful completion.

## Delivery Plan

- [x] PR 1: Split existing e2e files, extract `e2eframework`, add suite selection, environment configuration, Pod proxy helpers, and failure diagnostics. Preserve existing test semantics.
- [x] PR 2: Add mock recorder, debug endpoint, strict PD contract support, NIXL manifest, fault injection, and SGLang shared handoff state. Cover mock behavior with its own tests.
- [x] PR 3: Add successful PD e2e for SHFS, NIXL, SGLang raw JSON preservation, and TRT-LLM HTTP handoff shape.
- [x] PR 4: Add `gateway-pd` failure e2e for vLLM/TRT synchronous prefill failure and SGLang async handoff failure.

## Acceptance Criteria

- `AIBRIX_E2E_SUITE=all|gateway|controller|gateway-pd` selects suites while using one deployed test cluster and sequential execution.
- Every PD e2e request has a unique `x-request-id` and asserts its observed prefill/decode legs from mock recorder data, not pod-log matching.
- The default gateway suite runs all four successful PD contracts.
- Failure cases identify the failed leg and never report a false successful decode.
- Failures collect gateway/mock logs, pod descriptions/events, and the relevant recorder dump.
- Local failures can preserve resources with `AIBRIX_E2E_KEEP_RESOURCES_ON_FAILURE=true`.

## Non-Goals

- Validate real GPU, Neuron, EFA, or NIXL KV block transfer.
- Track latest upstream engine behavior in the default e2e suite.
- Add PD streaming coverage in this work; it will be scoped separately.

googs1025:
## Summary
- split live-cluster e2e tests into gateway and controller packages
- add shared e2e framework configuration and suite selection
- update CI Dockerfile/testdata paths and failure diagnostics

Closes #2675

## Verification
- `go test ./test/e2e/framework -count=1`
- `go test ./test/e2e/... -run "^$" -count=1`
- `go test ./test/e2e/controller/autoscaling/external-metrics-adapter -count=1`
- `bash -n test/run-e2e-tests.sh test/run-external-metrics-e2e.sh`
- validated the moved RoleSet, external metrics adapter, and controller manager Docker build paths

Full Kind e2e is delegated to CI.


test/README.md
```diff
@@ -70,7 +73,7 @@ make test-e2e
 
 # Or run script directly
 # Note: Required port-forwards should be active in this mode
-go test ./test/e2e/ -v -timeout 0
+go test ./test/e2e/gateway/... ./test/e2e/controller/... -v -timeout 0
```
varungup90: This local-dev command omits `-p 1`, but `test/run-e2e-tests.sh` was updated to run `go test -p 1 "${E2E_PACKAGES[@]}" ...` specifically because, per that script's own new comment, "E2E packages share one cluster, fixed mock resources, and Redis state" and must run serially.

Now that the tests are split across multiple packages (`gateway/api`, `gateway/pd`, `gateway/routing`, `controller/...`), `go test` will run them concurrently by default unless `-p 1` is passed. A developer following this README instruction verbatim could get flaky/incorrect results from cross-package interference against the shared cluster — the same failure mode the script change is meant to prevent.

Suggest adding `-p 1` here to match the script.

test/e2e/controller/autoscaling/testdata/external-metrics-autoscaler/aibrix/kustomization.yaml
varungup90, line 1: This file was renamed from `test/e2e/testdata/external-metrics-autoscaler/aibrix/kustomization.yaml` (5 directories deep) to this path, which is 7 directories deep — but the relative resource path wasn't updated to match:

```yaml
resources:
- ../../../../../config/standalone/autoscaler-controller
```

`../../../../../` (5 levels) resolves to `test/e2e/controller/config/standalone/autoscaler-controller`, which doesn't exist. The real target (`config/standalone/autoscaler-controller`) is at repo root, so this needs 7 levels: `../../../../../../../config/standalone/autoscaler-controller`.

This breaks `test/run-external-metrics-e2e.sh`, which runs `kubectl apply -k test/e2e/controller/autoscaling/testdata/external-metrics-autoscaler/aibrix` — kustomize will fail to resolve the resource.

varungup90 (changes requested):
Reviewed the e2e test restructuring. The Go package split itself is solid — I checked every moved test file against its new package's `helpers_test.go` and all the symbols line up, so nothing there breaks compilation.

Two issues found, both left as inline comments:

1. **`test/e2e/controller/autoscaling/testdata/external-metrics-autoscaler/aibrix/kustomization.yaml`** — the relative resource path wasn't updated for the file's new, two-levels-deeper location, so it now points outside the repo. This breaks `test/run-external-metrics-e2e.sh`.
2. **`test/README.md`** — the manual local-dev `go test` command is missing the `-p 1` flag that `test/run-e2e-tests.sh` now uses specifically to keep the new e2e packages from running concurrently against the shared cluster/Redis state.

---

test/README.md
googs1025: Fixed in 16dc429: the README command now uses `go test -p 1` to preserve serial execution across shared-cluster e2e packages.

test/e2e/controller/autoscaling/testdata/external-metrics-autoscaler/aibrix/kustomization.yaml
googs1025: Fixed in addc537: updated the moved Kustomization resource path to reach the repository root; `kustomize build` now succeeds.

varungup90 (approved):

PR #2389, opened 2026-06-23
feat(pd): roleset eligibility extraction, load-imbalance fixes, and tracker registration ordering

varungup90:
## Pull Request Description
This PR refactors and hardens the prefill-decode disaggregation routing path in several related areas:

### 1. Roleset Eligibility (`pd_roleset.go` — *new file*)
Extracts the "group pods by roleset and filter to complete pairs" logic into a dedicated file. Key additions include:
* `eligiblePDRolesets`: Returns rolesets that have ≥1 routable prefill and ≥1 routable decode pod.
* `HasEligibleRoleset`: Public helper used by the gateway for an early-exit check.
* `roleReplicaCardinality`: Counts distinct replicas using `role-replica-index` labels (correct for multi-node tensor-parallel setups where one logical replica spans several pods).
* `DescribePDRolesetEligibility`: Diagnostic view of per-roleset and fleet-wide replica counts.
* `pdPodNames`: Stable pod name string for structured logs.

### 2. Load-Imbalance Fixes (`pd_disaggregation.go`)
* **Decode imbalance (`loadImbalanceSelectDecodePod`):**
    * Pods that have no `RealtimeNumRequestsRunning` metric are now excluded from the spread calculation (they still contribute their pending-decode count to the scoring pass). This prevents a thundering herd to freshly restarted or scrape-gapped pods, which previously appeared artificially idle.
    * Throughput spread similarly excludes pods that have no metric, rather than treating missing throughput as 0.
    * `decodePodKVCacheUsagePerc` falls back from `kv_cache_usage_perc` to `gpu_cache_usage_perc` for backward-compatible KV cache utilization on older engines.
* **Prefill imbalance (`loadImbalanceSelectPrefillPod`):** Guard added so it only narrows decode candidates when a non-nil pod is actually selected.
* **Bucket filtering (`collectAndBucketPods`):** Prompt-length bucket overrides are now applied atomically. Both prefill and decode must have bucket-eligible pods before either list is replaced, keeping them roleset-aligned.
* **Combined-pod collection:** Combined (non-PD) pod scanning is gated on `bucketingEnabled` and refactored to use `PDRolePrefill`/`PDRoleDecode` constants.

### 3. Tracker Registration Ordering
* `filterPrefillDecodePods` now calls `AddPrefillRequest` and `AddPendingDecode` immediately after a pair is selected and before returning. 
[...]

PR #2678, opened 2026-09-08
[Bug] Register PD prefill/decode selections atomically with the selection

### LINKED ISSUE #2677: [Bug] PD router scores prefill/decode candidates on a stale tracker snapshot under concurrent requests
### 🐛 Describe the bug

`pdRouter.filterPrefillDecodePods` reads `PrefillRequestTracker` / `PendingDecodeTracker` to score candidates, but the winning pods are only registered in `Route` after `podSelector.Select` returns (`AddPendingDecode` / `AddPrefillRequest`). There is no lock between the tracker read inside `scorePrefillPods` / `loadImbalanceSelectPrefillPod` / `scoreDecodePods` and that registration.

Under a burst of concurrent requests every request scores the same pre-burst snapshot: none of them sees the others' picks. With `least_request` (or `prefix_cache` on cold prompts, where the request-count term is the only load signal) the burst piles onto whichever pods looked idle at read time instead of being spread across them.

#2389 / #2394 (merged via #2395) moved registration to right after the pair is selected, which narrows the window but does not close it: the tracker read in the scorers and the write after selection still race across goroutines.

### Steps to Reproduce

1. Four idle prefill pods in one roleset plus one decode pod, `AIBRIX_PREFILL_SCORE_POLICY=least_request`.
2. Issue 256 concurrent `Route` calls, with the prefill HTTP call parked until the whole burst has been routed (so every in-flight prefill stays registered).
3. Count `prefill-target-pod` per pod. Expected 64 each. Observed in one run on `main` (fe7db93e):

```
prefill-0:59 prefill-1:62 prefill-2:76 prefill-3:59
```

A unit test that reproduces this deterministically (`TestPDRouter_ConcurrentBurstSeesPriorSelections`) is part of the fix PR.

### Expected behavior

"Read tracked load → pick → register" should be one atomic step so the k-th decision sees the k-1 registrations made before it. With that, `least_request` keeps per-pod counts within one of each other and the burst above lands exactly 64/64/64/64.

Tokenization / prefix matching (`PrefillScorePolicy.Prepare`) depends only on the request and does not need to be serialized; only the tracker-dependent half of the selection does.

I have a fix ready and will open a PR against this issue.

### Environment

- AIBrix version: `main` at fe7db93e
- Deployment environment: reproduced in a unit test against the gateway plugin (`pkg/plugins/gateway/algorithms`), no cluster needed
- LLM(s) being used: n/a

qidaye:
## Pull Request Description

`filterPrefillDecodePods` reads `PrefillRequestTracker` and `PendingDecodeTracker` to score candidates, but the winning pods were only registered back in `Route` after `Select` returned. Concurrent requests therefore all scored the same pre-burst snapshot: none of them saw the others' picks, and a burst at idle pods piled onto whichever pods looked emptiest at read time. With four idle prefill pods and 256 concurrent requests under `least_request`, one run on `main` landed `59/62/76/59` instead of `64` each.

#2389 / #2394 moved registration to right after the pair is selected, which narrows the window but does not close it. This PR makes "read tracked load → pick → register" one atomic step:

- **`pdRouter.selectMu`.** The load-dependent part of `filterPrefillDecodePods` (combined-pod imbalance check, prefill/decode imbalance fast paths, per-pod scoring, `finalPDScore`) runs under it, and the chosen decode and prefill pods are registered with the trackers before the lock is released. `Route` no longer registers; it still owns the matching removals (`RemovePendingDecode` deferred, `RemovePrefillRequest` on prefill failure / in the executor).
- **`scorePrefillPods` split into `preparePrefillScorer` + `scorePreparedPrefillPods`.** `PrefillScorePolicy.Prepare` (tokenization and prefix matching) depends only on the request, so it runs before the lock is taken; only the tracker-dependent half is serialized. `scorePrefillPods` keeps its signature as the composition of the two, so existing tests and callers are unchanged.
- Combined-pod early returns register the pending decode too, matching what `Route` did for them before.
- `PodSelector.Select` doc now states that a successful `Select` registers its picks; `DefaultSelector` delegates to `filterPrefillDecodePods`, which does so.

Lock scope: `selectMu` is router-scoped, `countersMu` is taken inside it (in `finalPDScore`) and never the other way round, and nothing inside performs I/O (the prefix-index update is a non-blocking channel send; metric reads are in-memory cache lookups).

### Test

`TestPDRouter_ConcurrentBurstSeesPriorSelections` (new, `pd_select_register_test.go`): 256 concurrent `Route` calls at four idle prefill pods with every prefill HTTP call parked on a gate until the whole burst is routed. It asserts that the tracker ledger observed while the burst is held and the `prefill-target-pod` response headers both show exactly 64 per pod, that the pending-decode count is 256, and that every registration is released afterwards.

- On `main` the test fails (`59/62/76/59` in the run above).
- With this PR: `go test -race -count=3 -run TestPDRouter_ConcurrentBurstSeesPriorSelections ./pkg/plugins/gateway/algorithms/` passes.
- `go test -race ./pkg/plugins/gateway/algorithms/... ./pkg/types/...` and `go test ./pkg/plugins/gateway/` pass.

## Related Issues
Resolves: #2677

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>

### DIFF
--- a/pkg/plugins/gateway/algorithms/pd/selector/selector.go
+++ b/pkg/plugins/gateway/algorithms/pd/selector/selector.go
@@ -30,6 +30,10 @@ import (
 
 // PodSelector picks a prefill/decode pod pair from the full ready-pod list.
 // A nil prefill pod signals combined-role routing (decode pod handles both phases).
+//
+// A successful Select also registers the chosen pods with the request trackers
+// it consulted, atomically with the decision, so that concurrent selections
+// observe each other; the caller (Route) owns the matching removals.
 type PodSelector interface {
 	Select(routingCtx *types.RoutingContext, readyPods []*v1.Pod) (prefill *v1.Pod, decode *v1.Pod, err error)
 }
--- a/pkg/plugins/gateway/algorithms/pd/trackers.go
+++ b/pkg/plugins/gateway/algorithms/pd/trackers.go
@@ -43,9 +43,10 @@ func NewPrefillRequestTracker() *PrefillRequestTracker {
 }
 
 // AddPrefillRequest records that requestID has been assigned to podName and
-// increments that pod's active-request counter. Called in Route before
-// doPrefillRequest so concurrent routing sees in-flight assignments. Must be
-// paired with RemovePrefillRequest when prefill completes or the assignment is abandoned.
+// increments that pod's active-request counter. Called by the pod selector as
+// soon as the prefill pod is chosen, under the same lock as the selection, so
+// concurrent routing sees in-flight assignments. Must be paired with
+// RemovePrefillRequest when prefill completes or the assignment is abandoned.
 func (t *PrefillRequestTracker) AddPrefillRequest(requestID, podName string) {
 	countInterface, _ := t.podRequestCounts.LoadOrStore(podName, &atomic.Int32{})
 	count := countInterface.(*atomic.Int32)
@@ -148,8 +149,9 @@ func NewPendingDecodeTracker() *PendingDecodeTracker {
 }
 
 // AddPendingDecode records that requestID has been assigned to podName and
-// increments that pod's pending-decode counter. Called at the start of Route
-// after pod selection. RemovePendingDecode is deferred in Route.
+// increments that pod's pending-decode counter. Called by the pod selector as
+// soon as the decode pod is chosen, under the same lock as the selection.
+// RemovePendingDecode is deferred in Route.
 func (t *PendingDecodeTracker) AddPendingDecode(requestID, podName string) {
 	if t == nil {
 		return
--- a/pkg/plugins/gateway/algorithms/pd_disaggregation.go
+++ b/pkg/plugins/gateway/algorithms/pd_disaggregation.go
@@ -206,6 +206,20 @@ type pdRouter struct {
 	selectionCounts       map[string]int64
 	podSelector           selector.PodSelector
 	prefillExecutor       prefill.PrefillExecutor
+
+	// selectMu makes "read every candidate's tracked load, pick the best,
+	// register the pick" one atomic step in filterPrefillDecodePods. Without
+	// it, concurrent requests in a burst all read the same pre-burst tracker
+	// snapshot and pile onto whichever pods looked idle at read time, because
+	// the registration that would have made them visible to each other only
+	// happened after selection returned.
+	//
+	// Only load-dependent work runs under it: the imbalance fast paths, the
+	// per-pod scoring, finalPDScore and the two tracker registrations.
+	// Tokenization and prefix matching (PrefillScorePolicy.Prepare) depend
+	// only on the request and run before the lock is taken. countersMu is
+	// taken inside selectMu (in finalPDScore) and never the other way round.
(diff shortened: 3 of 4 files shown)


varungup90 (approved):
Reviewed the diff (`pd/selector/selector.go`, `pd/trackers.go`, `pd_disaggregation.go`, and the new test) and verified it end-to-end rather than just reading it:

**Correctness of the fix**
- Traced every return path of `filterPrefillDecodePods`: all three success paths (both combined-pod early returns and the final scored pair) now register with the trackers before returning; all error paths return `nil, nil, err` without registering, so there's nothing to leak.
- `Route` no longer double-registers — it only owns the `defer RemovePendingDecode` / error-path `RemovePrefillRequest`, and `RemovePrefillRequest`/`RemovePendingDecode` are already idempotent (`LoadAndDelete`-based), so the pre-existing "remove in executor + remove in Route on error" double-call stays safe.
- `preparePrefillScorer` is correctly hoisted outside `selectMu`: I checked `ScorePod` for all three prefill policies (prefix_cache, least_request, conductor) — each does a per-pod map/param lookup with no dependency on the size of the pod slice passed to `Prepare`, so scoring a narrower post-imbalance-fast-path slice with a scorer prepared over the wider slice (as the comment claims) is sound.
- `countersMu` is only ever locked inside `finalPDScore`, which only runs under `selectMu`, and never the reverse — no lock-order inversion.
- Confirmed `cache.GetMetricValueByPod`/`GetMetricValueByPodModel` are plain in-memory `sync.Map` lookups and `enqueuePrefixUpdate` is a non-blocking channel send, so the "no I/O under the lock" claim in the comment holds.

**Ran it, not just read it**
- Built `./pkg/plugins/gateway/...` on the PR branch — clean.
- `go test -race -count=3 -run TestPDRouter_ConcurrentBurstSeesPriorSelections` — passes every time.
- To make sure the new test is actually load-bearing (not a false positive), I cherry-picked it onto the merge-base commit (`fe7db93`, pre-fix) and ran it there: it fails reproducibly (e.g. `38/101/34/83`, `57/67/64/68`, `43/94/43/76` across three runs), confirming the race is real and this PR closes it.
- Full `go test -race ./pkg/plugins/gateway/algorithms/... ./pkg/types/...` — all green.
- `golangci-lint run ./pkg/plugins/gateway/algorithms/...` — 0 issues.

**One non-blocking observation**
`selectMu` now serializes the entire load-dependent scoring pass (imbalance checks + per-pod scoring + `finalPDScore`) across *every* concurrent PD-routed request on the router, not just the tracker read/register step. That's an inherent trade-off of making the decision atomic (you can't parallelize reads of mutable state that the decision itself needs to be consistent), and the PR is explicit about it, but it's worth watching under very high QPS or very large pod counts, since it turns a previously-parallel (if racy) hot path into a fully serialized one. Not a blocker — correctness over the previous silent load-skew bug is the right call, and a finer-grained (e.g. per-roleset) lock could be a later optimization if profiling shows contention.

Nice, focused fix with a genuinely convincing regression test.


---

Issue #2680, opened 2026-09-08
[RFC]: Token-weighted prefill load policies for the PD router (token_load, hybrid_cache_load)


qidaye:
> **Status (2026-09-09):** implemented and merged. PR 1 #2681 (`token_load`, tracker, metrics, wiring) and PR 2 #2682 (`hybrid_cache_load`, `AIBRIX_MIN_MATCH_PCT`, `new_tokens` charge) are both on `main`. The text below is the original proposal; the places where the merged code differs from it (session header name, session-table bounds, env variable ranges, metric names) are marked **[as merged]**.

### Summary

Add two prefill score policies to the PD router that measure a prefill pod's load in **estimated prompt tokens** instead of in requests:

- `token_load`: pick the pod with the lowest token-weighted load.
- `hybrid_cache_load`: same load figure, discounted by the pod's prefix-cache match so that cache affinity can override *small* load differences but never large ones.

Both read one shared in-memory `TokenLoadTracker` that the router keeps in step with its own routing decisions:

```
priority(pod) = active_prefill_tokens(pod) + kv_weight * resident_kv_tokens(pod)

token_load:         score = priority
hybrid_cache_load:  score = priority * (1 - (match_pct/100)^2 * cache_factor)
```

Lower is better. `active_prefill_tokens` is charged when the pod is selected and released when the prefill HTTP call returns. `resident_kv_tokens` is charged at the same time and released when the whole request finishes, because the prefill pod keeps that request's KV cache until the decoder has pulled it.

No new dependencies, no engine-side changes, no change to the `PrefillScorePolicy` interface. Selection is `AIBRIX_PREFILL_SCORE_POLICY=token_load|hybrid_cache_load` or `routingConfig.prefillScorePolicy`, like the existing policies.

The design was developed by my colleague @zhutong196 and me on a downstream fork of AIBrix; PRs will carry `Co-authored-by` for both of us.

### Motivation

The three existing prefill policies all count **requests**, not tokens, when they look at load:

| Policy | Load term |
| --- | --- |
| `prefix_cache` | `req_count / max_req_count` (tie-break after prefix match) |
| `least_request` | `req_count` |
| `conductor` | `queue = req_count * avg_prefill_seconds` |

[...]

PR #2682, opened 2026-09-08
[Gateway] Add hybrid_cache_load prefill policy and new-tokens charge to the PD router

qidaye:
## Pull Request Description

Second of the two PRs proposed in #2680. #2681 (`token_load`) has merged (a374e4c1) and this branch is rebased onto main, so the diff is this PR alone.

### Why

`prefix_cache` always prefers the pod that holds the prompt's prefix, even when that pod is buried under long prompts and an idle neighbour would answer sooner. `token_load` (#2681) ignores the cache and recomputes prefixes another pod already holds. Multi-turn traffic wants both: cache affinity between idle pods, load between busy ones. #2680 has the benchmark and the reasoning behind the cost model.

### What

**`hybrid_cache_load` policy** (`pd/prefill_scorer.go`): `Prepare` tokenizes and calls `MatchPrefix` exactly as `prefix_cache` does; `ScorePod` reads the same ledger as `token_load` and discounts it by the match:

```
r        = match_percent / 100   (0 when below AIBRIX_MIN_MATCH_PCT)
discount = 1 − r² × factor
score    = load × discount       when load ≥ 1, else discount
```

With no load anywhere every score is a bare discount below 1, so the best match wins outright. On busy pods a full match halves the load (default factor `0.5`), so a pod holding the prefix loses once it carries more than twice the load of an idle neighbour. The square keeps small matches from moving the decision.

**`AIBRIX_MIN_MATCH_PCT`** (default `0`, off): matches below the threshold count as no match, in the score and in the charge, so prompts that share only a system prompt are not all pulled to one pod. Applied only by `hybrid_cache_load` in this PR; adding it to `prefix_cache` is a one-liner if maintainers want it there too.

**`new_tokens` charge** (`TokenLoadTracker.NewTokens`): the token-load charge now covers only what the pod has to compute, by the first rule that applies:

1. session delta: the request carries the caller-owned `x-aibrix-session-key` and the gateway saw a shorter prompt for that (model, key) within `AIBRIX_TOKEN_LOAD_SESSION_TTL_SECONDS` (default `1800`): charge the growth. Each turn resends the history but an engine that still holds the conversation's KV computes only the new turn, so the rule assumes the earlier turns are reachable from the selected prefill pod (shared or tiered KV store, sticky routing, or the pod's own prefix cache); deployments without that should not send the key, or set the TTL to `0` to turn the rule off.
2. prefix match: the scorer looked the prompt up in the prefix cache: charge `prompt_tokens × (1 − match/100)`.
3. whole prompt.

Scorers expose the match through a small optional interface (`pd.PrefixMatchScorer`); `chargeTokenLoad` takes the prepared scorer it already had in scope. `token_load` behaviour is unchanged except that it now benefits from the session rule. The session table lives in the tracker and is swept by the janitor from #2681.

**Config** (all positive; invalid values keep the default):

| Variable | Default | |
|---|---|---|
| `AIBRIX_HYBRID_CACHE_LOAD_FACTOR` | `0.5` | discount of a full prefix match, `0` to `1` |
| `AIBRIX_MIN_MATCH_PCT` | `0` | ignore matches below this percentage, `0` to `100` |
| `AIBRIX_TOKEN_LOAD_SESSION_TTL_SECONDS` | `1800` | how long a session's last prompt size is remembered; `0` turns the session rule off (not in the RFC table; it exists because the session table needs a bound) |
| `AIBRIX_TOKEN_LOAD_MAX_SESSIONS` | `100000` | maximum live sessions; beyond it new sessions are charged without a delta until the janitor frees slots |

**Docs**: "Hybrid Cache-Load Scoring Policy" section, the `new_tokens` rules in the token_load section, policy lists and env table in `pd-disaggregation.rst`.

### Notes for reviewers

- `prompt_tokens` stays `len(ReqBody)/4` for both policies. The RFC said `hybrid_cache_load` would use the tokenizer's count; I dropped that because the byte encoding of `TokenizeInputText` differs per tokenizer (4 bytes per tiktoken token, 1 per character) and only the ranking matters. Easy to add later behind the same interface if wanted.
- The session key is `(model, x-aibrix-session-key)`. A shorter or equal prompt under a known session falls back to rules 2/3 and becomes the new baseline, so a reused ID does not produce a negative or stale delta. Keys over 256 bytes are ignored and the table is capped at `AIBRIX_TOKEN_LOAD_MAX_SESSIONS`, so the client-controlled header cannot grow gateway memory. The RFC named `x-session-id`; the PR switched to the existing caller-owned key after review, since `x-session-id` is the gateway-issued token of `session-affinity` routing.
- No `PrefillScorePolicy` interface change; `chargeTokenLoad` gains a parameter inside the router package only.

### Testing

- `pd/prefill_scorer_hybrid_cache_load_test.go`: discount math at 0/50/100 % with factor 0.5, load outweighing a full match, sub-token load treated as idle, min-match clamp (score and reported match agree), `ClampMinMatch` table, nil tracker, `PrefixMatchPercent` on non-prefix scorers, env config.
- `pd/token_load_tracker_test.go`: `NewTokens` without a session (no info / partial / full / clamped / negative), session delta across turns (first seen, growth, repeat, shrink, other model), session tracking off with TTL 0, janitor forgets idle sessions without touching charges, over-long IDs ignored, session cap and its release, swept session not refreshed, session count under concurrent sweeps, env config (including TTL 0 and the max-sessions bound).
- `pd_hybrid_cache_load_test.go` (router level): idle pods follow the prefix and are charged only the uncached half; a loaded pod loses despite a full match; five requests across two sessions and an anonymous one produce the expected cumulative charges; min-match threshold turns a 30 % match into a full charge; `routingConfig.prefillScorePolicy=hybrid_cache_load` on a token_load router.
- `test/e2e/gateway/pd/pd_hybrid_cache_load_test.go` (`TestPDDisaggregationVLLMHybridCacheLoad`): a new `hybrid-cache-load` config profile on the mock PD StormService selects the policy; a fresh prompt is sent once, then repeated, and the repeat must route back to the prefill pod that served the warm-up (the policy scores the pod holding the prefix strictly below an idle cold pod). The `pd_token_load_*` gauges are scraped from both gateway replicas and must have been charged for every prefill pod used and be back at zero, reusing the helper from the `token_load` e2e. The new_tokens charge itself is not observable through metrics there (the mock answers in ms), so it stays in unit tests; the gateway log at `-v=4` on a local Kind cluster built like the Installation E2E workflow shows the expected pair, warm-up `new_tokens=361 prefix_match_percent=0 cost=3861` and repeat `new_tokens=0 prefix_match_percent=100 cost=3500`, with the repeat winning at prefill score 0.5 against 1.
- `go test -race ./pkg/plugins/gateway/algorithms/pd/... ./pkg/plugins/gateway/algorithms/ ./pkg/metrics/...` and `golangci-lint` (v2.13.1) pass locally, re-run after the rebase onto main.

## Related Issues

Part of #2680. Builds on #2681.

Co-authored with @zhutong196.

### DIFF
--- a/pkg/plugins/gateway/algorithms/pd/prefill_scorer.go
+++ b/pkg/plugins/gateway/algorithms/pd/prefill_scorer.go
@@ -20,14 +20,14 @@ limitations under the License.
 //
 //   - prefill_scorer.go  — PrefillScorePolicy / PrefillScorer interfaces and
 //     their built-in implementations (prefix_cache, least_request, conductor,
-//     token_load).
+//     token_load, hybrid_cache_load).
 //   - decode_scorer.go   — DecodeScorePolicy / DecodeScorer interfaces and
 //     their built-in implementations (load_balancing, least_request), plus the
 //     policy registry used by AIBRIX_DECODE_SCORE_POLICY.
 //   - trackers.go        — PrefillRequestTracker and PendingDecodeTracker,
 //     which bridge the gap between pod selection and actual request start.
 //   - token_load_tracker.go — TokenLoadTracker, the token-weighted prefill
-//     load ledger behind the token_load policy.
+//     load ledger behind the token_load and hybrid_cache_load policies.
 package pd
 
 import (
@@ -37,6 +37,7 @@ import (
 	"github.com/vllm-project/aibrix/pkg/cache"
 	"github.com/vllm-project/aibrix/pkg/metrics"
 	"github.com/vllm-project/aibrix/pkg/types"
+	"github.com/vllm-project/aibrix/pkg/utils"
 	"github.com/vllm-project/aibrix/pkg/utils/prefixcacheindexer"
 	"github.com/vllm-project/aibrix/pkg/utils/tokenizer"
 	v1 "k8s.io/api/core/v1"
@@ -62,6 +63,20 @@ const (
 	// routes prefill requests to the pod with the lowest token-weighted load
 	// as tracked by TokenLoadTracker, without consulting the prefix cache.
 	PrefillScorePolicyTokenLoad = "token_load"
+
+	// PrefillScorePolicyHybridCacheLoad selects the hybrid scoring policy,
+	// which discounts a pod's token-weighted load by its prefix-cache match so
+	// that cache affinity wins on idle pods and load wins on busy ones.
+	PrefillScorePolicyHybridCacheLoad = "hybrid_cache_load"
+
+	// DefaultHybridCacheLoadFactor is how much a full prefix match discounts a
+	// pod's token load under hybrid_cache_load: score = load × (1 − factor)
+	// at 100 % match.
+	DefaultHybridCacheLoadFactor = 0.5
+
+	// DefaultMinMatchPct is the prefix-match percentage below which
+	// hybrid_cache_load treats a match as no match. 0 keeps every match.
+	DefaultMinMatchPct = 0.0
 
 	// Default TTFT estimation coefficients for conductor policy.
 	// Units: time in milliseconds, tokens are unitless.
@@ -404,7 +419,14 @@ func (p *tokenLoadPrefillPolicy) Name() string { return PrefillScorePolicyTokenL
 // UsesTokenLoad reports whether policy scores from a TokenLoadTracker, i.e.
 // whether the router must charge the tracker for requests scored by it.
 func UsesTokenLoad(policy PrefillScorePolicy) bool {
-	return policy != nil && policy.Name() == PrefillScorePolicyTokenLoad
+	if policy == nil {
+		return false
+	}
+	switch policy.Name() {
+	case PrefillScorePolicyTokenLoad, PrefillScorePolicyHybridCacheLoad:
+		return true
+	}
+	return false
 }
 
 // tokenLoadScorer is the request-scoped scorer produced by tokenLoadPrefillPolicy.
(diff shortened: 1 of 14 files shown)


pkg/plugins/gateway/algorithms/pd/token_load_tracker.go
```diff
@@ -212,6 +235,71 @@ func EstimatePromptTokens(reqBody []byte) int {
...
+//     matchPct/100). Pass a negative matchPct when no match information is
+//     available.
+//  3. Otherwise the whole prompt.
+//
+// The result is never negative or larger than promptTokens.
+func (t *TokenLoadTracker) NewTokens(model, sessionID string, promptTokens, matchPct int) (int, string) {
+	if promptTokens < 0 {
+		promptTokens = 0
+	}
+	if sessionID != "" && t.cfg.SessionTTL > 0 {
```
varungup90: `sessionID` here is the raw, unvalidated value of the client-supplied `x-session-id` header, and it becomes a key in `t.sessions` (a `sync.Map`) that persists for up to `SessionTTL` (default 1800s) before the janitor sweeps it — independent of request concurrency. Unlike `activeTokens`/`kvTokens`/`entries`, which are bounded by the number of in-flight requests, this map's size is bounded only by the number of *distinct* session-ID values seen in the last `SessionTTL` window.

Since there's no length cap or cardinality limit on the header, a client that sends many requests each with a unique (or large) `x-session-id` value can grow this map roughly proportional to `QPS × SessionTTL` before anything gets swept — e.g. at 1000 req/s with the default 1800s TTL that's up to ~1.8M live entries. Worth at least capping the key length (`simple_session_affinity.go` already has a `maxSessionKeyLen = 256` precedent) or otherwise bounding the table size, given this header is fully attacker-controlled and reaches an unauthenticated code path.

---

pkg/plugins/gateway/algorithms/pd_disaggregation.go
```diff
@@ -340,11 +350,25 @@ func (r *pdRouter) DoneRequestTrace(_ *types.RoutingContext, requestID string, _
...
+// header, otherwise the part of the prompt the pod's prefix cache does not
+// hold when scorer looked it up, otherwise the whole prompt. See
+// TokenLoadTracker.NewTokens.
+func (r *pdRouter) chargeTokenLoad(routingCtx *types.RoutingContext, pod *v1.Pod, policy pd.PrefillScorePolicy, scorer pd.PrefillScorer) {
 	if r.tokenLoadTracker == nil || !pd.UsesTokenLoad(policy) {
 		return
 	}
-	cost := r.tokenLoadTracker.PrefillCost(pd.EstimatePromptTokens(routingCtx.ReqBody))
+	promptTokens := pd.EstimatePromptTokens(routingCtx.ReqBody)
+	sessionID := routingCtx.ReqHeaders[constants.HeaderSessionID]
```
varungup90: This reuses `constants.HeaderSessionID` (`x-session-id`), which already has an established, unrelated meaning elsewhere in this gateway: `simple_session_affinity.go` treats it as an opaque, *gateway-generated*, base64-encoded `IP:Port` string that the client is expected to echo back verbatim to stick to a backend pod (see `sessionAffinityRouter.Route` and the `RespHeaders[constants.HeaderSessionID]` write in that file). Here it's being repurposed as a client-chosen, semantically-meaningful conversation identifier for prefill cost accounting.

Since `pd` and `session-affinity` are different routing strategies, they likely aren't both active for the same model, so this probably doesn't cause a functional collision today. But it's the same header name carrying two very different contracts in the same codebase (gateway-assigned backend token vs. client-assigned conversation ID), which is easy to trip over later — e.g. if `session-affinity` semantics are ever combined with PD routing, or if a client blindly forwards whatever `x-session-id` it last received. Might be worth a distinct header (e.g. `x-conversation-id`) for the new_tokens session-delta feature, or at least a docs callout of the ambiguity.

---

varungup90 (commented):
Reviewed the diff (`prefill_scorer.go`, `token_load_tracker.go`, `pd_disaggregation.go`, docs, and unit/e2e tests). The discount math (`score = load × (1 − r²×factor)`), the `ClampMinMatch` threshold, and the `NewTokens` rule ordering (session delta → prefix match → whole prompt) all check out against their test cases, including the boundary conditions (min-match clamp, sub-token load treated as idle, nil tracker, session TTL disabled/expired). Nothing here looks like a correctness bug in the scoring or charging logic itself.

Two things worth a look before merge, left as inline comments:
1. The new `TokenLoadTracker.sessions` map is keyed by the client-supplied `x-session-id` header with no length or cardinality bound, and entries live for `SessionTTL` (default 1800s) regardless of concurrency — a potential memory-growth vector since that header is fully attacker-controlled.
2. `x-session-id` already has an established, different meaning elsewhere in the gateway (`simple_session_affinity.go`'s gateway-assigned backend-affinity token), and this PR reuses the same header name for a client-assigned conversation ID. Probably not a functional collision today given `pd` and `session-affinity` are separate routing strategies, but worth flagging as a naming/design ambiguity.

Nice test coverage otherwise — the router-level tests (`pd_hybrid_cache_load_test.go`) and tracker tests cover the interesting interactions (load outweighing a full match, min-match clamp, session delta across turns) well.

---

pkg/plugins/gateway/algorithms/pd/token_load_tracker.go
qidaye: Good catch, thanks. Fixed in e86afff9:

- session IDs longer than 256 bytes are ignored, the same bound `simple_session_affinity.go` applies to its key;
- the table is capped at `AIBRIX_TOKEN_LOAD_MAX_SESSIONS` live sessions (default 100000, tens of MB worst case). When it is full, new sessions are not recorded and their requests fall through to the prefix-match / whole-prompt rules, so a flood of distinct IDs costs the flooder its own delta, never gateway memory. The janitor sweep releases the slots.

Unit tests cover the length limit, the cap and its release, and the count staying consistent under concurrent admissions and sweeps. While in there I also closed the small window between the expiry check and the delete in `sweepExpired` (tombstone under the session lock), which cursor flagged inline.

pkg/plugins/gateway/algorithms/pd_disaggregation.go
qidaye: Agreed, reusing `x-session-id` mixed two contracts. Rather than add a new header, e86afff9 reads `x-aibrix-session-key`: it already exists as the caller-owned opaque key (captured in `gateway_req_headers.go`, listed in the header table and in `multi_router_readme.md`), and a client-chosen conversation identifier is exactly its semantics. `x-session-id` is no longer consulted by the PD router, so the gateway-issued backend token stays what it is. The docs and the header table now say which header does what. The RFC (#2680) named `x-session-id`; I will update it to match once this lands.

pkg/plugins/gateway/algorithms/pd/prefill_scorer.go
```diff
@@ -363,3 +386,206 @@ func (s *conductorScorer) ScorePod(pod *v1.Pod, reqCnt, _ float64) float64 {
...
+
+// DefaultHybridCacheLoadConfig returns the defaults, overridden by the
+// AIBRIX_HYBRID_CACHE_LOAD_FACTOR and AIBRIX_MIN_MATCH_PCT environment
+// variables. Each must be positive; an unset, empty or invalid value keeps the
+// default.
+func DefaultHybridCacheLoadConfig() HybridCacheLoadConfig {
+	return HybridCacheLoadConfig{
+		Factor:      utils.LoadEnvFloat("AIBRIX_HYBRID_CACHE_LOAD_FACTOR", DefaultHybridCacheLoadFactor),
+		MinMatchPct: utils.LoadEnvFloat("AIBRIX_MIN_MATCH_PCT", DefaultMinMatchPct),
+	}
```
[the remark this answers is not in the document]
qidaye: `utils.LoadEnvFloat` already rejects parse errors and non-positive values with a warning and falls back to the default, so negative values were never accepted. That said, "positive" was the wrong contract for these two knobs: 0 is a valid value of both, and the factor has an upper bound. e86afff9 parses `AIBRIX_HYBRID_CACHE_LOAD_FACTOR` as 0 to 1 and `AIBRIX_MIN_MATCH_PCT` as 0 to 100, warning and falling back outside those ranges.

pkg/plugins/gateway/algorithms/pd/token_load_tracker.go
```diff
@@ -0,0 +1,414 @@
...
+// AIBRIX_TOKEN_LOAD_TTL_SECONDS and AIBRIX_TOKEN_LOAD_SESSION_TTL_SECONDS
+// environment variables. Each must be positive; an unset, empty or invalid
+// value keeps the default.
+func DefaultTokenLoadConfig() TokenLoadConfig {
+	return TokenLoadConfig{
+		KVWeight:    utils.LoadEnvFloat("AIBRIX_TOKEN_LOAD_KV_WEIGHT", DefaultTokenLoadKVWeight),
+		RequestCost: utils.LoadEnvFloat("AIBRIX_TOKEN_LOAD_REQUEST_COST", DefaultTokenLoadRequestCost),
+		TTL:         time.Duration(utils.LoadEnvInt("AIBRIX_TOKEN_LOAD_TTL_SECONDS", DefaultTokenLoadTTLSeconds)) * time.Second,
+		SessionTTL:  time.Duration(utils.LoadEnvInt("AIBRIX_TOKEN_LOAD_SESSION_TTL_SECONDS", DefaultTokenLoadSessionTTLSeconds)) * time.Second,
+	}
```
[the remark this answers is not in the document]
qidaye: `utils.LoadEnvFloat` / `utils.LoadEnvInt` already reject parse errors and non-positive values with a warning and fall back to the default, so these four cannot go negative. The one deliberate change in e86afff9 is that `AIBRIX_TOKEN_LOAD_SESSION_TTL_SECONDS=0` now turns session tracking off, which the field documentation promised but the loader did not allow.

pkg/plugins/gateway/algorithms/pd/token_load_tracker.go
```diff
@@ -212,6 +235,71 @@ func EstimatePromptTokens(reqBody []byte) int {
...
+		}
+	}
+	if matchPct >= 0 {
+		if matchPct > 100 {
+			matchPct = 100
+		}
+		return promptTokens * (100 - matchPct) / 100, NewTokensSourcePrefixMatch
+	}
+	return promptTokens, NewTokensSourcePrompt
+}
```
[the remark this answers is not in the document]
qidaye: This ordering is deliberate; the part that was wrong was the documentation, so e86afff9 fixes the docs and adds an off switch rather than changing the rule.

The session-delta rule exists for deployments where a conversation's earlier KV is reachable from any prefill pod: a shared or tiered KV store (AIBrix's distributed KV cache, hierarchical caches on the engine side) or sticky routing. There the gateway's prefix table does not know that a "cold" pod can fetch the history, and `max(delta, uncached)` would charge the whole prompt every time hybrid rebalances a conversation, making the ledger pessimistic exactly where the policy is meant to move load. Delta charging is the configuration we benchmarked this policy with.

Where prefill pods only have a local prefix cache, the caller should not send the key, or the operator sets `AIBRIX_TOKEN_LOAD_SESSION_TTL_SECONDS=0` (now a real off switch; `LoadEnvInt` used to reject 0) and rule 2 applies. The docs now state the assumption and the escape hatch explicitly.

qidaye:
Thanks for the review. e86afff9 addresses both points, plus the overlapping ones from the bots:

- **Session table bounded**: IDs over 256 bytes ignored; at most `AIBRIX_TOKEN_LOAD_MAX_SESSIONS` (default 100000) live sessions, new ones beyond that charged by the prefix-match / whole-prompt rules until the janitor frees slots. Tests for the limit, the cap and concurrent sweeps.
- **Header**: the charge now reads the existing caller-owned `x-aibrix-session-key` instead of `x-session-id`, so the gateway-issued affinity token keeps its single meaning. Docs and header table updated.
- **Sweep race**: the janitor tombstones a session under its lock, so a refresh racing the delete re-inserts instead of being lost.
- **Env parsing**: `AIBRIX_HYBRID_CACHE_LOAD_FACTOR` is parsed as 0 to 1 and `AIBRIX_MIN_MATCH_PCT` as 0 to 100 (0 no longer logs as invalid, factor cannot go above 1); `AIBRIX_TOKEN_LOAD_SESSION_TTL_SECONDS=0` turns the session rule off.
- **Session delta on a cold pod** (cursor): kept as designed, since the rule targets deployments with a shared or tiered KV store; the docs now state that assumption and point to the off switch.

The earlier Go Race Tests failure was the SLO queue timing spec (`slo_test.go`), unrelated to this diff; the new push reruns it.

varungup90 (approved):

PR #2709, opened 2026-09-11
[Misc] Add successful PD protocol contract E2E coverage

### LINKED ISSUE #2675: [TEST] Restructure E2E tests and add PD protocol contract coverage

googs1025:
## Summary

- Add recorder-backed successful PD contract E2E coverage for vLLM SHFS, vLLM NIXL, SGLang HTTP/raw JSON, and TRT-LLM OpenAI handoff.
- Add Kubernetes Pod proxy recorder queries, bounded polling, unique request IDs, and raw gateway request support.
- Validate complete handoff propagation, bootstrap fields, nested JSON preservation, and exact TRT integer IDs.

Closes #2675

## Tests

- `go test ./test/e2e/framework/... -count=1`
- Focused PD helper tests: passed
- `go vet ./test/e2e/framework/... ./test/e2e/gateway/pd`
- PR2 Python regression tests: 131 passed
- `git diff --check`

Live Kind E2E was not run because this environment has no configured gateway/Kind cluster.


test/e2e/framework/recorder.go
```diff
@@ -0,0 +1,230 @@
...
+	var successfulPrefill, successfulDecode []MockRequestRecord
+	err := wait.PollUntilContextTimeout(
+		context.Background(),
+		time.Second,
+		2*time.Minute,
+		true,
+		func(ctx context.Context) (bool, error) {
+			prefillRecords, err := QueryMockRequests(ctx, client, namespace, prefillPod, requestID)
+			if err != nil {
+				if !apierrors.IsNotFound(err) {
```
varungup90, line 180: `WaitForSuccessfulPDLegs`'s poll callback treats any pod-proxy error other than `NotFound` as immediately fatal (`return false, err`), which makes `wait.PollUntilContextTimeout` abort the whole wait and `t.Fatalf`. This is inconsistent with every other PD polling helper in this package (`PollPDChatCompletion`, `waitForConsecutivePDSuccesses` in `client.go`), which log and keep retrying on any request error instead of failing fast.

A transient API-server hiccup (timeout, connection reset, a 5xx from the proxy) during any of the up-to-2-minute poll window would fail the whole contract test immediately, even though the PD legs would likely show up on the next poll — this looks like a likely source of E2E flakiness. Consider treating non-NotFound errors the same way as the other helpers (log and retry) rather than aborting.

---

test/e2e/framework/recorder.go
```diff
@@ -0,0 +1,230 @@
...
+			"query mock requests for pod %q request ID %q: %w",
+			podName,
+			requestID,
+			err,
+		)
+	}
+	return records, nil
+}
+
+func classifyPDRecords(records []MockRequestRecord, requestID, contract, engine, role string) (bool, error) {
```
varungup90, line 132: The `contract` parameter threaded through `classifyPDRecords`, `WaitForSuccessfulPDLegs`, and `SelectSuccessfulPDLegs` is never used to filter or validate anything (as the comments above these functions acknowledge) — every PD contract test's distinct contract string (`vllm-aibrix-shfs`, `vllm-aibrix-nixl`, `sglang-http`, `trtllm-openai`) is silently ignored here.

This is documented as intentional, but it's dead weight that could mislead a future reader into thinking contract-level filtering happens in this shared helper, when matching is actually only on requestID/engine/outcome/status. Worth considering whether to drop the parameter or wire it up once the mock recorder emits contract metadata.

---

test/e2e/gateway/pd/pd_contract_test.go
```diff
@@ -0,0 +1,412 @@
...
+	Port float64
+	Room float64
+}
+
+func requireBootstrapFields(t *testing.T, body map[string]any) bootstrapFields {
+	t.Helper()
+	host, ok := body["bootstrap_host"].(string)
+	require.True(t, ok, "bootstrap_host must be a string, got %T", body["bootstrap_host"])
+	require.NotEmpty(t, host)
+	port, ok := body["bootstrap_port"].(float64)
```
varungup90, line 395: `requireBootstrapFields` decodes `bootstrap_room` (and `bootstrap_port`) via plain `float64` instead of the exact-integer (`json.Number`) path this same file uses for TRT-LLM's `disagg_request_id` (`decodeMockJSONNumber`/`requireExactIntegerJSONNumbers`). SGLang's real `bootstrap_room` is a random int63 (see `pkg/plugins/gateway/algorithms/pd/engine/sglang.go`), which routinely exceeds float64's 52-bit mantissa.

Today this doesn't cause a false pass since only prefill-vs-decode equality is checked (both go through the same lossy rounding), but it's inconsistent with the "exact integer" care this PR takes elsewhere, and would silently lose precision if a future assertion checked the room value itself.

---

test/e2e/gateway/pd/pd_contract_test.go
```diff
@@ -0,0 +1,412 @@
...
+		"messages":[{"role":"user","content":"Say this is a test for vLLM SHFS PD contract"}],
+		"max_tokens":8,
+		"stream":false
+	}`)
+
+	result, err := sendPDRequest(context.Background(), e2eConfig, "pd", requestID, body)
+	require.NoError(t, err)
+	requireSuccessfulCompletion(t, result, modelNameVLLM)
+	gatewayRequestID := requireGatewayRequestID(t, result)
+	k8sClient, _ := initializeClient(context.Background(), t)
```
varungup90, line 165: Each of the four new PD contract tests calls `initializeClient` (→ `framework.InitializeClient`) just to get a client for a single raw pod-proxy GET request. That helper spins up full Pod and ModelAdapter shared informers and blocks on cache sync, which is unnecessary overhead here — a plain `kubernetes.NewForConfig(config)` would suffice for `QueryMockRequests`. It also passes `context.Background()`, which is never canceled, so the informer goroutines it starts leak for the life of the test binary. This pattern repeats in `TestPDContractVLLMNIXL`, `TestPDContractSGLangRawJSON`, and `TestPDContractTRTLLM`.

---

test/e2e/gateway/pd/pd_contract_test.go
```diff
@@ -0,0 +1,412 @@
...
+	require.NoError(t, json.Unmarshal(body, &originalBody))
+	require.NotContains(t, prefillBody, "kv_transfer_params")
+	require.Equal(t, modelNameVLLMNIXL, prefillBody["model"])
+	require.Equal(t, false, prefillBody["stream"])
+	require.Equal(t, originalBody["messages"], prefillBody["messages"])
+
+	decodeBody := decodeMockRequestBody(t, decode)
+	var expectedHandoff map[string]any
+	require.NoError(t, json.Unmarshal(prefill.Response, &expectedHandoff), "decode prefill recorder response")
+	require.Equal(t, vllmSHFSOpaqueSentinel, expectedHandoff["opaque"])
```
varungup90, line 252: Nit: `vllmSHFSOpaqueSentinel`, named for the vLLM SHFS contract, is reused verbatim here in the NIXL test's assertion. A reader could reasonably mistake this for a copy-paste bug pulling in SHFS-specific behavior rather than the shared mock sentinel it actually is. Consider a contract-neutral name (e.g. `vllmMockOpaqueSentinel`) to remove the ambiguity.

---

varungup90 (commented):
Reviewed the new PD contract E2E coverage and framework helpers. Nice, thorough addition — field names and headers checked out against `development/app/mock_recorder.py`/`app.py` and the gateway code, and all CI checks are green. Left 5 inline notes, most significant being the `WaitForSuccessfulPDLegs` poll-abort-on-transient-error inconsistency (likely flakiness source); the rest are minor precision/efficiency/naming nits. Nothing here blocks merge.

---

googs1025:
> All review feedback has been addressed.                                                                                                                                             
  >                                                                                                                                                                                     
  > - Added nil-client protection and transient Pod-proxy retry handling.                                                                                                               
  > - Removed the unused contract parameter.                                                                                                                                            
  > - Replaced informer-based client initialization with a lightweight Kubernetes client.                                                                                               
  > - Added exact integer handling for SGLang bootstrap fields.                                                                                                                         
  > - Renamed the opaque sentinel to be contract-neutral.                                                                                                                               
  >                                                                                                                                                                                     
  > The raw messages and tools comparisons intentionally remain byte-level because this E2E is specifically validating raw JSON preservation.                                           
  >                                                                                                                                                                                     
  > Current CI is green  I’m going to merge this PR now. Any follow-up issues will be addressed in a separate PR.

PR #2710, opened 2026-09-11
fix(modeladapter): drop max loading retries cap

jiangxiaobin96:
MaxLoadingRetries may not be needed here for the following three reasons:
1. If one pod's retryCount is greater than MaxLoadingRetries, it will never be attempted to load the adapter.
2. Because of backoffDuration, the more times a pod has failed, the longer it will be until the next attempt to load the adapter.
3. Since the multiplier is at most 300/5 = 60, the maximum value of retryCount can be adjusted to 6.


varungup90:
@jiangxiaobin96 can you sign the PR

jiangxiaobin96:
> @jiangxiaobin96 can you sign the PR

done.

googs1025 (approved):
Thanks for the quick fix and look forward to more contributions. 😄

PR #2715, opened 2026-09-12
[Misc] Use explicit Area field as primary source for area/* labels

### LINKED ISSUE #2714: Improve issue and PR label automation
## Background

AIBrix currently uses automation to label Issues and Pull Requests.

Related work:

- #2696 [Add local AIBrix workflow bot](https://github.com/vllm-project/aibrix/pull/2696)
- #2711 [Add path-based PR labeler](https://github.com/vllm-project/aibrix/pull/2711)

These two PRs improved the labeling workflow, but the problem is not fully resolved.

## Problem

Issue labeling can still assign unrelated `area/*` labels because the bot matches broad keywords across the entire Issue body.

PR and Issue labeling also need a clearer separation. PR area labels should reflect the files changed by the PR, rather than keywords in the PR title or description.

## Proposed direction

Use repository templates as the primary source of label intent:

- Add default `kind/*` labels to Issue Forms where appropriate.
- Add an explicit `Area` field to Issue Forms.
- Add an `Area` selection section to the Pull Request template.
- Use changed file paths as the fallback for PR area labels.
- Use keyword matching only as a fallback for Issues.
- Avoid assigning multiple unrelated area labels automatically.
- Keep PR title/description validation separate from area labeling.
- Handle labels produced by previous bot versions consistently.

The previous two PRs made useful improvements, but did not fully solve the classification and ownership problem.

TheNetherWatcher:
## Pull Request Description
Adds an explicit "Area" field as the primary source for `area/*` labels, instead of relying on keyword matching across the whole issue body (which could tag unrelated areas at once).

- Issue Forms (bug/feature/RFC): new "Area" dropdown. If set, it's used directly; keyword matching is now only a fallback, and it returns at most one label instead of every rule that happens to match.
- PR template: new "Area" checklist. If exactly one box is checked, the bot applies that label; otherwise the existing path-based labeler (`labeler.yml`) output stands untouched.
- Fixed the Issue Forms' stale default labels (`bug`, `triage`, `RFC`) to the labels the bot actually manages (`kind/bug`, `kind/feature`).
- Extended the bot's `--self-test` to cover the new parsing/precedence.

## Related Issues
Resolves: #2714

## Area
- [x] cicd


.github/scripts/aibrix-bot.py
```diff
@@ -281,7 +329,8 @@ def handle_issue(event: dict, github: GitHub):
 
 def handle_pull_request(event: dict, github: GitHub):
     pull_request = event["pull_request"]
-    errors = validate_pr(pull_request.get("title", ""), pull_request.get("body") or "")
+    body = pull_request.get("body") or ""
+    errors = validate_pr(pull_request.get("title", ""), body)
```
googs1025, line 333: Please separate blocking information errors from advisory findings here. A title-format warning or a template placeholder should not add `triage/needs-information`; only an empty/default PR description should. This is why PRs such as #2726 can be mislabeled even when their context is otherwise sufficient.

.github/scripts/aibrix-bot.py
```diff
@@ -291,9 +340,24 @@ def handle_pull_request(event: dict, github: GitHub):
...
+        lines.append(f"Area (explicit): {explicit_area}")
+    elif checked:
+        lines.append("Area: multiple boxes checked, ignoring - falling back to changed-file paths")
+    else:
+        lines.append("Area: none checked, using changed-file paths (path-based labeler)")
     _summary(lines)
     try:
-        github.sync_labels(pull_request["number"], labels, PR_MANAGED_LABELS)
+        if explicit_area:
+            github.sync_labels(pull_request["number"], labels + [explicit_area], PR_MANAGED_LABELS | AREA_LABELS)
```
googs1025, line 358: Please preserve existing path-based `area/*` labels when adding an explicit Area. The workflow runs `actions/labeler` first, so this managed set can delete valid labels such as `area/testing` when the PR selects `area/runtime`. The desired result should be the union of path labels and the explicit label.

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -4,6 +4,20 @@
 ## Related Issues
 Resolves: #[Insert issue number(s)]
 
+## Area
+<!-- Check the single area this PR belongs to, if you know it. Leave everything
```
googs1025, line 8: Please document the exact Area behavior and title examples here: `[API] Add per-replica request limits`, `[Bug] Fix Gateway routing regression`, `[CI] Update integration workflow`, `[Docs] Document Gateway configuration`, and `[Misc] Add integration coverage`. Also state that zero/multiple Area selections fall back to path labels, and title/placeholder warnings do not add triage.

.github/ISSUE_TEMPLATE/100_feature_request.yml
```diff
@@ -36,6 +37,24 @@ body:
     placeholder: |
       One possible approach could be to [implementation idea]. 
       This would integrate with AIBrix's existing [component] by [method].
+- type: dropdown
+  id: area
+  attributes:
+    label: Area
+    description: |
+      Which part of AIBrix does this affect? Leave as "Not sure" if you don't know - a maintainer will triage it. This is used to apply an `area/*` label instead of guessing from keywords in this report.
```
googs1025, line 45: Please keep this Area dropdown synchronized with the bug/RFC forms and the PR template. `Not sure` must mean an explicit opt-out from keyword fallback, while legacy issues without an Area section may still use the fallback. Add a consistency check so future option changes cannot drift across templates.

.github/scripts/aibrix-bot.py
```diff
@@ -303,6 +367,10 @@ def handle_pull_request(event: dict, github: GitHub):
 
 
 def self_test() -> None:
+    # Every AREA_RULES label must map onto a real AREA_OPTIONS/AREA_LABELS
+    # entry, so the two lists can't silently drift apart.
+    assert {label for label, _ in AREA_RULES} == AREA_LABELS
```
googs1025, line 372: Please extend `self_test` with a fake GitHub label-sync case. Parser-only assertions will not catch the important regressions: a single explicit Area deleting path labels, multiple/no selections preserving them, and title/placeholder findings incorrectly adding `triage/needs-information`.

PR #2718, opened 2026-09-13
[Bug] Guard SLOQueue.queueRank against zero throughput

stywzn:
## What this PR does

`SLOQueue.queueRank` divided by the throughput read from the GPU profile without checking for zero, and `SLOQueue` had no unit tests at all. This adds the guard and the first specs for the function.

## The defect

`ModelGPUProfile.ThroughputRPS` returns an error only when the whole `Tputs` table is absent. A zero cell *inside* the table is returned as a valid `0.0`, so `queueRank` computed `float64(sub.Len()-1) / 0` and returned `NaN` when the sub-queue held only the head request, or `+Inf` when requests were queued behind it — in both cases alongside a nil error, leaving `Peek` with no signal to skip the profile.

That matters because `Peek` orders profiles with `q.higherRank(...) < 0`. Every comparison involving `NaN` is false, which leaves `sort.Slice` with an inconsistent less function, and `Inf - Inf` is `NaN` too, so two zero-throughput heads would hit the same problem.

This is reachable with the project's own data: `simulator-llama2-7b-a100.json` has zero entries in `tputs`, including the whole top output-length row, which `GetSignature` clamps any longer request onto.

## The fix

`PendingLoadProvider.GetConsumption` already performs the same division and already guards it with `ErrorSLOFailureRequest`, and `Peek` already skips a profile when `queueRank` returns an error. This applies the established pattern rather than inventing one — two lines:

```go
} else if throughput == 0.0 {
	return 0.0, cache.ErrorSLOFailureRequest
}
```

## Tests

Three specs, on a hand-built `ModelGPUProfile`:

1. The test request maps onto the intended cell — pins the fixture, so a change to `GetSignature` or the index layout fails here rather than confusing the specs below.
2. A zero-throughput cell returns `ErrorSLOFailureRequest` and a zero rank.
3. A non-zero-throughput profile with two queued requests returns a finite rank, verifying the `(len-1)/throughput` arithmetic on the success path.

Spec 3 passes the head request's own `RequestTime` as `currentTime`, so the elapsed term is exactly zero and the assertion does not depend on wall-clock timing.

## Note on scope

The original PR was test-only and characterised the NaN and `+Inf` values instead of fixing them. Following review, it now fixes the defect, since the path is gated behind `queueOverallSLO` (a package-level `bool` fixed at `false`) and carries no user-facing risk, and since the codebase already had the guard elsewhere.

## Test plan

```
go test ./pkg/plugins/gateway/queue/ -run TestQueue -v
golangci-lint run ./pkg/plugins/gateway/queue/...
```

Both pass.

### DIFF
--- a/pkg/plugins/gateway/queue/slo_queue_test.go
+++ b/pkg/plugins/gateway/queue/slo_queue_test.go
@@ -57,9 +57,9 @@ var _ = Describe("SLOQueue", func() {
 	It("should map the test request onto the zero-throughput cell", func() {
 		req := newTestRequest("req-1", g)
 		features, err := req.Features()
-		Expect(err).To(BeNil(), "expected no error, got %v", err)
+		Expect(err).NotTo(HaveOccurred())
 		signature, err := profile.GetSignature(features...)
-		Expect(err).To(BeNil(), "expected no error, got %v", err)
+		Expect(err).NotTo(HaveOccurred())
 		Expect(signature).To(Equal([]int{1, 1}))
 	})
 
@@ -67,11 +67,10 @@ var _ = Describe("SLOQueue", func() {
 		q := &SLOQueue{}
 		sub := NewSimpleQueue[*types.RoutingContext](4)
 		req := newTestRequest("req-1", g)
-		// nolint:errcheck
-		sub.Enqueue(req, time.Now())
+		Expect(sub.Enqueue(req, time.Now())).To(Succeed())
 		rank, err := q.queueRank(time.Now(), req, sub, profile)
 		Expect(math.IsNaN(rank)).To(BeTrue(), "expected rank to be NaN, got %v", rank)
-		Expect(err).To(BeNil(), "expected no error, got %v", err)
+		Expect(err).NotTo(HaveOccurred())
 	})
 
 	It("should return +Inf when throughput is zero and other requests are queued", func() {
@@ -79,14 +78,12 @@ var _ = Describe("SLOQueue", func() {
 		sub := NewSimpleQueue[*types.RoutingContext](4)
 
 		req := newTestRequest("req-1", g)
-		// nolint:errcheck
-		sub.Enqueue(req, time.Now())
+		Expect(sub.Enqueue(req, time.Now())).To(Succeed())
 
 		req1 := newTestRequest("req-2", g)
-		// nolint:errcheck
-		sub.Enqueue(req1, time.Now())
+		Expect(sub.Enqueue(req1, time.Now())).To(Succeed())
 		rank1, err := q.queueRank(time.Now(), req, sub, profile)
 		Expect(math.IsInf(rank1, 1)).To(BeTrue(), "expected rank to be +Inf, got %v", rank1)
-		Expect(err).To(BeNil(), "expected no error, got %v", err)
+		Expect(err).NotTo(HaveOccurred())
 	})
 })
--- a/pkg/plugins/gateway/queue/slo_queue_test.go
+++ b/pkg/plugins/gateway/queue/slo_queue_test.go
@@ -1,0 +1,92 @@
+/*
+Copyright 2024 The Aibrix Team.
+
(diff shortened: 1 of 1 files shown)


stywzn:
Thanks for the review — all six suggestions are applied in 1c09290e.

- `Expect(err).NotTo(HaveOccurred())` now replaces `Expect(err).To(BeNil())` throughout. `HaveOccurred` already renders the error in the failure output, so the manual messages were redundant.
- Both `sub.Enqueue` calls are asserted with `Expect(...).To(Succeed())` and the `nolint:errcheck` directives are gone. Asserting the setup means an enqueue failure surfaces at the line that caused it, rather than as a confusing rank assertion further down.

I kept the explicit failure messages on the `math.IsNaN` / `math.IsInf` assertions. Those matchers reduce to a bare `Expected <bool>: false to be true`, which does not say what the rank actually was — the message carries the observed value so a failure is diagnosable without re-running.

One note: this now diverges from `simple_queue_test.go` in the same package, which still uses `// nolint:errcheck` for its enqueues. Happy to align that file in a follow-up if you would prefer the package consistent.

stywzn:
The `Installation E2E Tests` failure looks unrelated to this PR.

The failing case is `TestPDDisaggregationVLLMTokenLoad` in `test/e2e/gateway/pd`:

```
POST "http://localhost:8888/chat/completions": 503 Service Unavailable
token_load PD chat completion request 0 failed
```

The same case failed the same way on `main` about ninety minutes earlier — run [34742143076](https://github.com/vllm-project/aibrix/actions/runs/34742143076), same 503, same `request 0 failed`, with no PR involved. This branch also passed the same suite earlier today, and the only change since then was assertion style inside `pkg/plugins/gateway/queue/slo_queue_test.go`, which is not compiled into any deployed image.

Failing on the very first request points at the PD backends not being ready when the case starts, rather than at a routing regression.

Could someone re-run the job? I do not have permission to on this repo. I could not find an existing issue for this case — happy to open one and look into the readiness gate separately if it is not already tracked.

stywzn:
Thanks — this is a much better outcome than the characterisation tests I opened with. Taken the first option in 4137db89, and retitled the PR since it is no longer test-only.

**The guard.** `GetConsumption` was the deciding argument: it performs the same division on the same table and already returns `ErrorSLOFailureRequest` on a zero denominator, so `queueRank` was simply the one place the pattern was missing. Two lines, matching that shape:

```go
} else if throughput == 0.0 {
	return 0.0, cache.ErrorSLOFailureRequest
}
```

Your point about `Inf - Inf` is what settled it for me. I had been thinking of `+Inf` as a tolerable "most violating" sentinel, but two zero-throughput heads collapse it straight back to `NaN` and the comparator is inconsistent again — so `+Inf` was never a safe value to lock in.

**Finite-rank counterpart.** Added, against a second profile whose cell is non-zero: two queued requests, asserting the rank lands at the expected value so the `(len-1)/throughput` arithmetic is covered rather than just the divide-by-zero edge. It passes the head request's own `RequestTime` as `currentTime`, so the elapsed term is exactly zero and the assertion carries no wall-clock dependency.

**log2 comment.** Added above `Indexes`, noting that `Unmarshal` converts a JSON profile's indexes into log2 space and that a hand-built profile has to supply them already converted. That is genuinely the thing that confused me while building the fixture, so it should save the next reader the same detour.

**`g` → `predictor`.** Done, along with naming the two profiles `zeroTputProfile` and `nonZeroTputProfile`.

**`BeforeEach`.** I kept the mapping check as its own spec rather than moving it. My reasoning: as a spec it names the failure in the report — a broken mapping shows up as "should map the test request onto the zero-throughput cell" failing — whereas in `BeforeEach` it would surface as all three specs failing without that label. Happy to move it if you would rather have the earlier failure point; it is a one-line change either way.

You are right that the live `rank()` path is still untested — I would like to cover it, along with the SLO branch precedence in `rankImpl`, in a follow-up rather than growing this PR further.

varungup90 (approved):

PR #2719, opened 2026-09-13
[Bug] Enforce tokenizer pool capacity during concurrent creation

lllleolin-max:
## Pull Request Description

Two concurrent requests for different models can both pass `TokenizerPool`'s capacity check, perform their health checks outside the lock, and then insert into the last available slot. A pool configured for one tokenizer can therefore retain two. The initial capacity check also prevents replacing an unhealthy tokenizer when the pool is full, even though replacing that model needs no additional slot.

Recheck capacity while holding the insertion lock and allow same-model replacements. Close clients discarded after a failed health check or a lost capacity race, and close the unhealthy client when replacing it.

The regression tests use a local HTTP server and a channel barrier to force both creations past their initial checks before allowing either to finish. They cover different-model capacity contention, same-model deduplication, recovery at capacity, and idle-connection cleanup.

## Related Issues

No existing issue. Checked all-state tokenizer/pool PRs and issues for overlapping work.

## Validation

- On unmodified main with the regression tests added: capacity contention, unhealthy replacement, and failed-health-check cleanup fail; same-model deduplication passes.
- `go test -tags=nozmq ./pkg/plugins/gateway/algorithms ./pkg/utils/tokenizer -count=1` — passed.
- `go test -race -tags=nozmq ./pkg/plugins/gateway/algorithms -run '^TestTokenizerPool' -count=1` — passed.
- `go vet -tags=nozmq ./pkg/plugins/gateway/algorithms ./pkg/utils/tokenizer` — passed.
- `gofmt` and `git diff --check` — passed.

Validated with Go 1.26.0 on Windows/amd64, using the existing `nozmq` build tag. No live inference service or Kubernetes cluster was needed; cluster E2E and ZeroMQ integration were not run.

## Submission Checklist

- [x] PR title includes the `[Bug]` prefix.
- [x] Changes are explained above.
- [x] New and existing tests in the affected packages pass.
- [x] Code follows the existing locking and cleanup patterns.
- [x] No public API or configuration change requires a documentation update.
- [x] Regression, adjacent-package, and race checks completed.

AI assistance: OpenAI GPT-6 assisted with implementation and tests; the change was reviewed and validated locally.


bolubo (commented):
I read through the fix and it looks right to me. The only red check is a flake that main fixed right after this branch was pushed.

**`Installation E2E Tests` fails on the pre-#2721 assertion.** The failing test is `TestModelAdapterDeletionCleansUpOwnedResources` (ModelAdapter controller suite, not the gateway): chat completions returned 500 in the step right after "running inference to confirm the adapter is servable before deletion" ([job log](https://github.com/vllm-project/aibrix/actions/runs/34804437883/job/103854696617)). That exact call was replaced with `waitForInference` in #2721, merged at 2026-09-14T04:03:38Z; this branch head is from 03:58:46Z and the E2E run was created at 03:59:25Z, so the run exercised the old assertion and failed at 04:36:10Z. A rebase on main picks up #2721 (and #2761, which later stabilized another ModelAdapter e2e test), and the E2E re-run should go green.

**On the fix itself:** the second check under the lock covers both races: another model filling the last slot, and another caller committing the same tokenizer. Both discard paths close what they drop, so the client leak from the TOCTOU is gone, and the new tests exercise the same-model and different-model cases and assert the discarded client is closed. One small note: the `exists` branch assumes the previous entry is unhealthy. That holds today because the healthy case returns earlier under the same lock hold, but a one-line comment there would make the invariant explicit for the next reader.

PR #2723, opened 2026-09-14
[Bug] Scope BlockRemoved prefix eviction to the pod that sent it

rishabhsai:
## Pull Request Description
With KV event sync enabled and more than one replica behind a model, one pod evicting a KV block drops that prefix for every pod, so `MatchPrefix` reports 0% for replicas that still hold the block and the router falls back to load-based selection. vLLM emits `BlockRemoved` on every eviction, so this fires continuously under memory pressure, which is when KV-cache-aware routing matters most.

The cause is `pkg/utils/syncprefixcacheindexer/sync_hash.go:354-362`: `ProcessBlockRemoved` calls `delete(contextData.prefixStore.prefixMap, aibrixHash)` on the outer map even though `prefixMap[hash]` is a `map[podName]*PodInfo`, and it never reads `event.SourcePod`, which the per-pod ZMQ subscriber already sets (`pkg/kvevent/handler.go:119`) and `pkg/cache/store_providers.go:195` forwards. The sibling `RemovePrefix` in the same file already does the per-pod thing. The `totalPrefixes--` on the next line is equally unconditional: it runs once per resolved engine hash whether or not `prefixMap` held that entry, so a `BlockRemoved` that arrives after `RemovePrefix` already dropped the prefix takes the counter negative. `AddPrefix` compares that counter against `maxPrefixesPerContext`, so a drifted counter silently raises the `AIBRIX_SYNC_MAX_PREFIXES_PER_CONTEXT` memory cap.

`ProcessBlockRemoved` now resolves the engine hashes without mutating, deletes only `event.SourcePod` from each prefix entry, and drops the entry, the `engineToAibrix` mapping, and the reverse block index only once no pod holds the block. `totalPrefixes` is decremented only when an entry is actually removed. Lock order stays `mappingMu` then `prefixMu`, matching `ProcessBlockStored`, and the `updateBlockIndex` calls run after `prefixMu` is released so `blockIndexMu` is never nested under it.

Behaviour change: a prefix now survives on the replicas that still hold the block instead of being cleared fleet-wide. A `BlockRemoved` with an empty `SourcePod` still clears the whole entry, so the existing call sites that omit it, such as `TestReverseIndex` and the stress suite, behave exactly as before.

## Related Issues
No existing issue. Checked all-state PRs and issues for `ProcessBlockRemoved`, `BlockRemoved SourcePod`, `totalPrefixes`, `sync_hash`, and `prefix index eviction`; no overlapping fix was found. #2287 (prefix index not invalidated when an engine clears its cache) and #2285 (KV event decoder drops fields) are the same bug class in this indexer but neither covers the eviction scope of `ProcessBlockRemoved`.

## Important Notes
Two table-driven tests were added to `sync_hash_test.go`. Against `main` they fail, and reverting only the `sync_hash.go` change with the tests kept reproduces exactly this output:

```
$ go test ./pkg/utils/syncprefixcacheindexer/ -run 'BlockRemoved|TotalPrefixes' -count=1
--- FAIL: TestProcessBlockRemovedOnlyEvictsSourcePod (0.00s)
    --- FAIL: TestProcessBlockRemovedOnlyEvictsSourcePod/one_of_two_pods_evicts_the_block (0.00s)
        sync_hash_test.go:902: expected 1 matches, got map[]
--- FAIL: TestTotalPrefixesStaysConsistentAfterRemove (0.00s)
    --- FAIL: TestTotalPrefixesStaysConsistentAfterRemove/one_of_two_pods_evicts_the_block (0.00s)
        sync_hash_test.go:996: expected totalPrefixes 1, got 0
    --- FAIL: TestTotalPrefixesStaysConsistentAfterRemove/removal_event_arrives_after_the_pod_was_unsubscribed (0.00s)
        sync_hash_test.go:996: expected totalPrefixes 0, got -1
        sync_hash_test.go:999: totalPrefixes -1 disagrees with prefix map size 0
FAIL

$ go test ./pkg/utils/syncprefixcacheindexer/ -run 'BlockRemoved|TotalPrefixes' -count=1   # with the fix
ok      github.com/vllm-project/aibrix/pkg/utils/syncprefixcacheindexer 0.547s
```

`go test -race ./pkg/utils/syncprefixcacheindexer/... ./pkg/cache/... ./pkg/plugins/gateway/...` passes on all 14 packages with tests, including the indexer's concurrent store/remove stress suite. `go build ./...`, `make lint` (golangci-lint v2.13.1, 0 issues), and `make licensecheck` are clean.

<details>
<summary><strong>Submission Checklist</strong></summary>

- [ ] Documentation updated to reflect changes (if applicable), not applicable, `pkg/utils/syncprefixcacheindexer/README.md` already documents the second level as a prefix hash to pod mapping

</details>


rishabhsai:
Pushed bd9ec8b. The BlockRemoved tests now assert the engine-to-aibrix mapping and the reverse block index directly: both still hold the block after one of two pods evicts it and are gone only once the last pod does, and a new mixed-event case sends one BlockRemoved with two hashes where the reporting pod is the last holder of only the second, checking each hash separately. I also added a klog.Warningf on the empty-SourcePod branch naming the model, LoRA id, and block hash count, so an unscoped caller shows up in the logs instead of silently wiping the fleet. Took the capacity nit since len(event.BlockHashes) is the real upper bound and it was a one-token change.

varungup90 (approved):

PR #884, opened 2025-03-19
[API] Cache and Router refactoring for concurrent performance, concurrent safety and stateful routing.

zhangjyr:
## Pull Request Description

Refactoring for cache:
1. Merge multiple pod, model, and metric mapping by adding Pod metadata and Model metadata and using two main thread-safe registries for metadata.
2. Eliminate the global cache mutex lock on reading and replace it with multiple layers of locks in the metadata.
3. Cache serves in different contexts: control manager, metadata, and gateway. Added cache initialization helpers for each context for later customization.
4. Added more unit test cases to cover the pod and model adapter changes.

Refactoring for router
1. Eliminate thread-unsafe map access in the Router interface.
2. Merge two contexts, context.Context and routing.RoutingContext, as RoutingContext
5. Abstract away the Router interface and the RoutingContext for shared access from both the routing and cache package.
6. RoutingContext now supports both sync and async routing resolution by following the Promise paradigm.

## Related Issues
Resolves: #868 

**Important: Before submitting, please complete the description above and review the checklist below.**

---

<details>
<summary><strong>Contribution Guidelines (Expand for Details)</strong></summary>

<p>We appreciate your contribution to aibrix! To ensure a smooth review process and maintain high code quality, please adhere to the following guidelines:</p>

<h3>Pull Request Title Format</h3>
<p>Your PR title should start with one of these prefixes to indicate the nature of the change:</p>
<ul>
    <li><code>[Bug]</code>: Corrections to existing functionality</li>
    <li><code>[CI]</code>: Changes to build process or CI pipeline</li>
    <li><code>[Docs]</code>: Updates or additions to documentation</li>
    <li><code>[API]</code>: Modifications to aibrix's API or interface</li>
    <li><code>[CLI]</code>: Changes or additions to the Command Line Interface</li>
    <li><code>[Misc]</code>: For changes not covered above (use sparingly)</li>
</ul>
<p><em>Note: For changes spanning multiple categories, use multiple prefixes in order of importance.</em></p>

<h3>Submission Checklist</h3>
<ul>
    <li>[ ] PR title includes appropriate prefix(es)</li>
    <li>[ ] Changes are clearly explained in the PR description</li>
    <li>[ ] New and existing tests pass successfully</li>
    <li>[ ] Code adheres to project style and best practices</li>
    <li>[ ] Documentation updated to reflect changes (if applicable)</li>
    <li>[ ] Thorough testing completed, no regressions introduced</li>
</ul>

[...]

PR #2728, opened 2026-09-14
[Bug] Fix data race and stale snapshots in utils.Registry

phakeandy:
## Pull Request Description

`utils.Registry` caches a slice of its values so `Array()` and `Len()` can serve
readers without taking the lock. The cached slice and its validity flag are two
separate fields, and the lock-free readers load them separately while the writers
replace them under the lock. Three defects follow from that:

**1. The lock-free read is not atomic.** `Array()` (registry.go:110) and `Len()`
(registry.go:145) do `arr, valid := reg.values, reg.valid`. The `// atomic`
comment does not hold: these are two ordinary loads of two independent fields,
racing against the two-value assignments in `Store()` (registry.go:92 and :95)
and `Delete()` (registry.go:63). A reader can observe a truncated `values`
together with a not-yet-cleared `valid`, and return an empty slice for a registry
that was never empty.

**2. A slice already returned to a caller gets rewritten.** The invalidation
branches truncate with `reg.values[:0]` to reuse the backing array, and
`updateArrayLocked()` (registry.go:164) rebuilds into that same array whenever
`cap` is sufficient. `Array()` therefore does not return a snapshot: its contents
change under the caller after any later `Store`/`Delete`. This one reproduces
single-threaded, no race detector involved.

**3. `CustomizedRegistry` has no synchronization at all.** `Store`
(registry.go:101) and `Delete` (registry.go:69) write `reg.values` with no lock
while `Array()` (registry.go:128) reads it with no lock. Worse,
`updateArrayLocked()` publishes `reg.values` while holding the lock and an
unlocked `Store` can clobber it immediately afterwards, so `Array()` can return a
nil `A`. This is the type behind `cache.Model.Pods`.

### Where this happens in production

The writer is the informer callback and the reader is a controller reconcile, in
the same process:

```
write  registry.go:92   <- pkg/cache/informers.go:415  addPodAndModelMappingLocked
                        <- client-go processorListener.run

read   registry.go:110  <- pkg/cache/cache_impl.go:121 ListModelsByPod
                        <- pkg/controller/modeladapter/scheduling/least_adapters.go:43
```

`ListPodsByModel` (cache_impl.go:79) is the same pattern on the gateway request
path, through `CustomizedRegistry`.

The consequences are not only theoretical. `ListModelsByPod` returning an empty
list makes `leastAdapters.SelectPod` believe the pod carries the fewest adapters
and schedule onto it. A nil `*PodArray` from `ListPodsByModel` passes the
`podsArr == nil` guard in gateway_req_body.go:302 (a typed nil inside a
`types.PodList` interface is not `== nil`), then `CountRoutablePods(nil) == 0`
turns a model that does have pods into a 503 `no ready pod available`.

### Why CI did not catch it

- `test/integration/` is where the informer and the controllers actually run
  concurrently, and those ginkgo suites do not run with `-race`.
- `make test-race-condition` does cover `pkg/...`, but `pkg/utils/utils_test.go`
  carries `//go:build !race`, so the ginkgo suite of this package is excluded
  from race builds entirely (`go test -race ./pkg/utils/` reports
  `no tests to run`).
- The existing `Registry` unit tests are single-threaded, so the interleaving is
  never produced in the first place.

### Evidence

Running the controller integration suite under the race detector, on main:

```
KUBEBUILDER_ASSETS=<abs> go test -race -count=1 ./test/integration/controller/...

Ran 81 of 81 Specs in 179.400 seconds
SUCCESS! -- 81 Passed | 0 Failed
--- FAIL: TestAPIs (179.40s)
    testing.go:1490: race detected during execution of test
```

All 81 specs pass; only the detector fires. Six reports, all on this one cache,
writes at registry.go:92 and :95, reads at registry.go:110.

The race is also observable without the detector. With a writer that only
re-stores keys that already exist -- so the map holds exactly 8 entries for the
whole run, and there is no legitimate state with fewer -- 50,000 `Array()` calls
on main reliably return 1 to 2 **empty** snapshots. The `CustomizedRegistry`
equivalent returns a nil array about once per 300,000 reads.

## Changes

`values []V` + `valid bool` become a single `atomic.Pointer[[]V]` (nil meaning
invalid), and `CustomizedRegistry.values` becomes `atomic.Pointer[A]`:

- Readers do one `Load()`, so the pointer and the length always come from the
  same snapshot. The lock-free read path that #884 introduced is preserved.
- Writers publish by copy-on-write and never mutate a snapshot that has already
  been handed out. `updateArrayLocked` allocates the snapshot at exactly
  `len(registry)` capacity, so a caller cannot grow into it either.
- `CustomizedRegistry.Store`/`Delete` now invalidate *after* updating the
  embedded registry. Invalidating first lets a concurrent `Array()` publish a
  snapshot that misses the update and keep it permanently; invalidating last
  makes the worst case one wasted rebuild.
- `CustomizedRegistry.updateArrayLocked` returns the value it just built instead
  of re-reading the field. That re-read was the lost update behind the nil array.

The public API, the method signatures and the nil-receiver behaviour are
unchanged. No caller distinguishes a nil result from an empty one (checked
across `least_adapters`, `bin_pack`, `store_providers`, `cache_metrics`,
`cache_log` and `informers`), so publishing an empty snapshot for a fresh
registry is not observable.

The read path does not get slower. `Array()` + `Len()` in parallel, ns/op:

| values | before | after |
|--------|--------|-------|
| 8      | 3.7 - 5.9 | 4.3 - 4.7 |
| 64     | 4.6 - 7.1 | 4.0 - 5.5 |
| 256    | 7.2 - 8.5 | 4.4 - 5.6 |

One atomic pointer load replaces copying a three-word slice header plus a bool.

## Tests

Four cases added to the existing `Registry` suite in
`pkg/utils/registry_test.go`. Three of them fail on the current code:

```
[Fail] should keep the array returned by Array() unchanged after later updates
    Expected <[]string | len:2>: ["key3", "key2"]
    to consist of <[]string | len:2>: ["key1", "key2"]
[Fail] should keep Array() and Len() consistent while registry is updated concurrently
    Expected <int>: 6   to equal <int>: 0
[Fail] should keep the array returned by customized Array() unchanged after later updates
FAIL! -- 29 Passed | 3 Failed
```

After the change: `Ran 32 of 32 Specs ... SUCCESS! -- 32 Passed | 0 Failed`.

The fourth case (concurrent `CustomizedRegistry.Array()`) does not fail reliably
on the current code without `-race` -- the lost-update window is two adjacent
instructions wide -- so it is included as the analogue of the `Registry` case
rather than as a before/after proof.

Verified:

```
go build ./pkg/...                                                    clean
go vet ./pkg/utils/                                                   clean
go test -count=1 ./pkg/utils/...                                      ok
go test -race -count=1 ./pkg/utils/... ./pkg/cache/...                ok, 0 races
go test -race -count=1 ./pkg/plugins/gateway/...                      ok, 0 races
go test -race -count=1 ./pkg/...                                      0 races
KUBEBUILDER_ASSETS=<abs> go test -race ./test/integration/controller/...
                                                     ok  183.764s, 0 races
```

The integration suite reported six races before this change and none after.

## Related Issues

None; found by auditing `pkg/utils` and confirmed by running the controller
integration suite under `-race`.

---

## Follow-ups (deliberately not in this PR)

- Adding `-race` to the integration job would have caught this. Not touched here
  because `AGENTS.md` asks before changing `.github/` workflow policy.
- The `//go:build !race` tag on `pkg/utils/utils_test.go` hides this whole
  package's specs from race builds. Removing it currently surfaces unrelated
  pre-existing findings in the package, so it is left for a separate change.

## Checklist

- [ ] Documentation updated to reflect changes (not applicable)


phakeandy:
> Residual, fine as a follow-up: production `valuesProvider` in `informers.go` aliases the snapshot into `PodArray.Pods`, and `initDeployments` sorts that slice in place. Exact capacity stops append into the snapshot; it does not stop `sort.Slice`. `SLOQueue.Peek` calls `Indexes()` on the `ListPodsByModel` result while other readers call `All()`.

Agreed. I opened #2741 to track this as a follow-up.

> Nits: `"should concurrent updateArrayLocked call return cached array"` is still two sequential calls; `NewRegistry` publishes `&[]V{}` while `NewRegistryWithArrayProvider` leaves both pointers nil.

The test name is not correct, i hava renamed it. And the constructor asymmetry is left unchanged because the customized empty value must be produced by the provider, and lazy initialization preserves the existing behavior.

phakeandy:
> * **`CustomizedRegistry.Store`/`Delete` still invalidate the outer cache after releasing the mutex.** Invalidate-after avoids a _permanent_ stale snapshot, but it is not the same as one snapshot. `Array()` is lock-free on the outer pointer; `Len()` reads the inner `Registry` cache. After a new-key `Store`, the inner slice is already COW-published at N+1 while `Array()` can still return the old `*PodArray`. After `Delete`, `Array()` can still list the removed pod. `ListPodsByModel` is `meta.Pods.Array()`, so the gateway can observe that window. Suggested fix: `storeLocked`/`deleteLocked`, and under one `mu` hold invalidate the outer pointer _first_, then update the map/inner cache.
> * **The new concurrent tests never run under `-race`.** They are Ginkgo specs, and `TestCache` lives in `pkg/utils/utils_test.go` with `//go:build !race`. `make test-race-condition` therefore does not execute the tests that exist to prove this fix. A normal `testing` test in `registry_test.go` (no Ginkgo) would cover this type without pulling in the unrelated package findings that block dropping `!race`.

Both issues are addressed in 9ae630fa.

for first issue, i added storeLocked and deleteLocked. CustomizedRegistry.Store and Delete now hold the registry mutex while first invalidating the outer cache and then updating the map and inner cache. This prevents readers from observing outer and inner snapshots from different generations.

for secound issue, I added standalone tests which is under -race.

varungup90 (approved):

PR #2738, opened 2026-09-16
[Misc] Add unit tests for SLOQueue Dequeue guard and Route error propagation

stywzn:
## Pull Request Description

Adds two specs covering the `SLOQueue` dequeue/route lifecycle guards. No production code is changed.

- **`Dequeue` before `Peek`** — calling `Dequeue` without a prior `Peek` must fail with "call SLOQueue.Peek first" and return a nil request, pinning the guard on `lastCandidateSubKey`.
- **`Route` after a failed `Peek`** — when `Peek` recorded an SLO routing conclusion in `lastCandidateError`, `Route` must surface that error and return an empty address instead of re-routing.

Both run against a bare `&SLOQueue{}` — no cache or queue contents needed — keeping them as fast as the existing rank specs.

Verified locally: `go test ./pkg/plugins/gateway/queue/ -run TestQueue -count=1`, `go vet` and `golangci-lint run` on the package all pass.

## Related Issues

N/A — continues the SLOQueue test coverage work from #2718 and #2734.


googs1025 (approved):
LGTM

PR #2742, opened 2026-09-17
[Feat] Add Redis-backed session-key pinning for cross-replica agentic routing

varungup90:
## Pull Request Description

Adds Redis-backed session-key pinning so `session-affinity` can keep concurrent agentic
requests on the same inference pod across gateway replicas, plus follow-up correctness
fixes so that stickiness actually holds after scale-out, cache isolation, and blend
behavior.
### Agentic routing
`x-session-id` only works after the first response comes back, which is too late for
agentic workloads that fan out concurrent tool calls or sub-agent requests. Callers can
now send a stable `x-aibrix-session-key` from the first request:
- Sequential multi-turn chat still uses the gateway-issued `x-session-id`.
- Concurrent/agentic requests use a caller-owned `x-aibrix-session-key` (conversation ID,
  agent-run ID, etc.).
- With `REDIS_HOST` configured, the key→pod pinning is shared across gateway replicas
  (1-hour sliding TTL). Without Redis, mapping stays deterministic via rendezvous hashing
  over the current ready pod set.
- Keys longer than 256 bytes are treated as unset and fall back to normal
  session-affinity resolution.
- The same header can be reused by PD `token_load` / `hybrid_cache_load` policies.
Docs: `docs/source/features/agentic-routing.rst`.

### Correctness Follow-ups

- **Drop the session-affinity load-balance auto-blend:** Session-affinity `ScoreAll` is binary (1 for the resolved pod, 0 elsewhere), so the previous 5:4 blend could never be overridden by load-balance disagreement. Bare `session-affinity` now runs unblended, like `pd` / `slo*` / standalone `load-balance`. Prefix-cache keeps the 5:4 ratio because its scoring is graded.
- **Honor stickiness on the single-ready-pod fast path:** `selectTargetPod` used to return that pod without calling `router.Route()`, so a session started while only one pod was ready never got a Redis pin or `x-session-id` header. After scale-out, rendezvous hashing could move the session. The router's `PostRouteUpdate` hook now runs on that path, including through `multiStrategyRouter`.
- **Isolate and bound the session-key cache:** Redis/local cache keys are scoped by model (`sessionCacheKey`) so two models sharing Redis — or a reused session-key value — don't overwrite each other. The local `sessionKeyPods` cache is capped by `AIBRIX_SESSION_AFFINITY_MAX_LOCAL_KEYS` (default 100k); past the cap, new keys still resolve via Redis or rendezvous hashing.
- **Harden Lookup / local-cache races:** `Lookup` rejects a nil provider instead of dereferencing it. `storeSessionKeyLocal` uses a `LoadOrStore` / `CompareAndSwap` retry loop so concurrent writes can't clobber a newer pinning or un-confirm one whose Redis write already landed.

> **Note / Known Leftover:** `persistSessionKeyToRedis`'s unconditional `SET` is shared by initial-claim, TTL-refresh, and failover-repin writes and can still race across gateway replicas. Called out with a `TODO` in code; not addressed here.

## Related Issues

---

<details>

<ul>
</ul>

<h3>Submission Checklist</h3>
<ul>
</ul>

</details>


pkg/plugins/gateway/algorithms/simple_session_affinity.go
```diff
@@ -35,76 +41,426 @@ const RouterSessionAffinity types.RoutingAlgorithm = "session-affinity"
...
-	// Session ID and session key missing, invalid, or pod not ready → fallback
-	klog.V(4).InfoS("Session affinity failed, falling back", "request_id", ctx.RequestID, "session_id", sessionID)
-	return r.fallbackRoute(ctx, readyPodList)
+	return rendezvousPod(ctx, pods, sessionKey), sessionKey, "rendezvous"
+}
+
+// Route implements session affinity by attempting to route requests to the same pod
+// using a session ID stored in the request header. The session ID encodes the target
+// pod's address as "IP:Port". If no valid session exists, it falls back to a randomly selected ready pod.
+func (r *sessionAffinityRouter) Route(ctx *types.RoutingContext, readyPodList types.PodList) (string, error) {
```
googs1025, line 488: The single-ready-pod fast path appears to bypass this method entirely.

`selectTargetPod` returns directly at `gateway.go:682-685` when only one single-port pod is ready. In that case, `Route` is not called, so a request carrying `x-aibrix-session-key` does not persist a Redis pin, and the response does not receive the `x-session-id` set here.

This can become visible when a session starts with one pod and the model later scales out: because the first request was not pinned, rendezvous hashing may select a newly added pod for the next request.

Could we either let `session-affinity` bypass that fast path, or run a common post-route hook before returning? It would also be useful to add a regression test that starts with one ready pod, expands the pod set, and verifies that the original session remains pinned.

pkg/plugins/gateway/algorithms/simple_session_affinity.go
```diff
@@ -35,76 +41,426 @@ const RouterSessionAffinity types.RoutingAlgorithm = "session-affinity"
...
+// at the routing call sites so a slow/unavailable Redis never adds latency to
+// the request path; a failed write just means this replica's local cache is,
+// for now, the only copy of this pinning. Returns whether Redis now has it.
+func (r *sessionAffinityRouter) persistSessionKeyToRedis(sessionKey, addr string) bool {
+	if r.redisClient == nil {
+		return false
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), sessionKeyRedisWriteTimeout)
+	defer cancel()
+	if err := r.redisClient.Set(ctx, sessionAffinityRedisKey(sessionKey), addr, sessionAffinityTTL).Err(); err != nil {
```
googs1025: Could we avoid using the same unconditional `SET` operation for initial claims, TTL refreshes, and failover repins?

Two gateway replicas that miss Redis while observing different ready-pod sets can select different pods and then overwrite each other’s mapping. There is also a stale-write case: a delayed TTL refresh for the old address may complete after another replica has repinned the session, restoring the old address in Redis.

One possible approach would be:

- use `SET NX EX` for the initial claim, and read the existing winner when the claim is not acquired;
- refresh the TTL only when the stored value still matches the local address;
- use a Lua script or another compare-and-swap operation for `oldAddr -> newAddr` repins.

A test with two router instances, different pod views, and deliberately reordered writes would help cover both concurrent claims and stale refreshes.

pkg/plugins/gateway/algorithms/simple_session_affinity.go
```diff
@@ -35,76 +41,426 @@ const RouterSessionAffinity types.RoutingAlgorithm = "session-affinity"
...
+	// timeout cache.GetPodsRunningRequests already uses on this same hot path.
+	sessionKeyRedisReadTimeout = 100 * time.Millisecond
+
+	// sessionKeyRedisWriteTimeout bounds the background persist call. It runs
+	// in its own goroutine off the request path, so this only guards against
+	// a leaked goroutine piling up on a stuck connection, not request latency.
+	sessionKeyRedisWriteTimeout = 2 * time.Second
+)
+
+func sessionAffinityRedisKey(sessionKey string) string {
```
googs1025: Should the cache identity include the model and a stable deployment or cluster scope?

At the moment, both the local cache and Redis are keyed only by the caller-provided `x-aibrix-session-key`. If the same key is used for model A and model B, the models have different candidate pod sets but still share one mapping. Each model can therefore treat the other model’s address as unavailable and repeatedly replace the same Redis entry. Deployments sharing one Redis instance could encounter a similar collision.

One option would be to derive the cache key from:

```text
deployment-scope | model | session-key
```

The caller-provided part could also be hashed rather than stored directly. If session keys are instead required to be globally unique across every model and deployment, could we document that requirement and add validation or a cross-model test?

googs1025 (commented):
A few suggestions on the session-affinity flow.

pkg/plugins/gateway/algorithms/simple_session_affinity.go
```diff
@@ -35,76 +41,426 @@ const RouterSessionAffinity types.RoutingAlgorithm = "session-affinity"
...
 }
 
-type sessionAffinityRouter struct{}
+type sessionAffinityRouter struct {
+	// sessionKeyPods is a Redis read-through cache of session-key -> pod
+	// address pins. It is only populated when redisClient is non-nil:
+	// without Redis the router stays stateless and rendezvousPod remains
+	// the sole (and cross-replica-deterministic) mapping. Idle entries
+	// are dropped during the sync loop after sessionAffinityTTL.
+	sessionKeyPods sync.Map // sessionKey (string) -> sessionKeyCacheItem
```
googs1025: Could we place a hard capacity bound on this local cache?

The cache key comes directly from the caller, and every distinct valid `x-aibrix-session-key` can remain in `sessionKeyPods` for up to one hour. A client sending continuously changing keys can therefore grow the map in proportion to the unique-key rate. The periodic synchronization also scans the entire map and issues batched `MGET` calls, so both local memory and synchronization work grow without a fixed upper bound.

Would a size-bounded TTL/LRU cache be appropriate here, with a configurable maximum and eviction metrics? Once the limit is reached, evicted or uncached keys could safely fall back to Redis lookup or stateless rendezvous hashing. It would also be useful to test that high-cardinality session keys cannot grow the local cache beyond the configured limit.

pkg/plugins/gateway/algorithms/simple_session_affinity.go
varungup90: for now, added a TODO

varungup90:
@googs1025I’ve addressed your comments and will go ahead and merge the PR. Feel free to continue reviewing and leave any additional comments—I’ll address them in a follow-up PR.

pkg/plugins/gateway/algorithms/simple_session_affinity.go
bolubo: Thanks for the detailed breakdown — I'd like to take this one.

Following the approach you outlined, I'll land it as small focused PRs rather than one change:

1. `SET NX EX` for the initial claim (the route path and the sync-pass retry);
2. adopt the winning value when a claim loses the race, so replicas converge without waiting for the next sync;
3. TTL refresh gated on the stored value still matching (no plain value write);
4. `oldAddr -> newAddr` repin via a Lua compare-and-swap.

Each step gets the two-replica / reordered-write tests you described, on the existing miniredis harness. Starting with the first one shortly.

Issue #2737, opened 2026-09-16
[Testing] Expand integration and E2E coverage for core controller and gateway workflows


googs1025:
## Summary

Expand AIBrix integration and end-to-end test coverage for several core workflows that currently have partial coverage or are only exercised at a lower test layer:

- ModelClaim controller state transitions
- ModelRouter routing-resource lifecycle
- Metrics-driven gateway routing and fallback behavior
- PodSet lifecycle and draining
- RayClusterFleet and RayClusterReplicaSet lifecycle

This is a tracking issue. The checklist may be completed through multiple focused PRs, but all work remains tracked here.

## Motivation

Current test coverage is strong around ModelAdapter, StormService, RoleSet, PD disaggregation, and basic gateway routing. The areas below still lack either deterministic integration coverage or real-cluster validation.

The goal is to verify complete workflows across Kubernetes resources, controllers, Gateway API resources, routing state, and real requests—not simply add more isolated happy-path tests.

### 1. ModelClaim controller integration coverage

Add an envtest-based integration suite under:

- `test/integration/controller/modelclaim_test.go`

Cover the following controller behavior:

- [ ] A newly created ModelClaim progresses through the expected initial status transitions.
- [ ] A ModelClaim with no matching warm pods reports `NoMatchingPods`.
- [ ] Reconciliation recovers when a matching pod becomes available.
- [ ] Invalid engine configuration produces the expected failed phase and condition.
- [ ] Runtime activation failure produces an observable condition/event and is retried according to controller policy.
- [ ] Runtime state is reflected as Activating, Active, Sleeping, and Failed.
- [ ] Deleting or losing an assigned pod causes the ModelClaim status to converge to the correct state.
- [ ] Reconciliation remains idempotent when the same ModelClaim is processed repeatedly.
- [ ] Multiple ModelClaims do not incorrectly overwrite each other's assignments or status.

Use a deterministic fake runtime endpoint instead of depending on an external inference engine.

### 2. ModelRouter end-to-end coverage

Add a live-cluster E2E package under:

- `test/e2e/controller/modelrouter/`

Validate the complete routing-resource lifecycle:

- [ ] Creating a supported model workload causes the expected HTTPRoute to be created.
- [ ] A workload outside `aibrix-system` causes the required ReferenceGrant to be created.
- [ ] The generated route points to the expected backend and model path.
- [ ] A request sent through the gateway reaches the expected model backend.
[...]

PR #2743, opened 2026-09-17
[CI] Add ModelRouter lifecycle E2E coverage

googs1025:
## Pull Request Description

Add a live-cluster ModelRouter E2E package covering the controller behavior that is already supported today.

- exercise Deployment and ModelAdapter discovery and validate generated HTTPRoute and ReferenceGrant configuration;
- send a uniquely identified request through the gateway and confirm it reaches the isolated mock backend;
- verify namespace-level ReferenceGrant sharing and cleanup, plus controller restart idempotency and post-restart add/delete handling;
- use unique namespaces, bounded polling, failure diagnostics, and the existing resource-preservation setting.

This PR does not change ModelRouter controller behavior and does not add update or same-model shared-HTTPRoute scenarios.

## Related Issues

Related to #2737 (scope 2).

## Test Plan

- [x] `make lint` (0 issues)
- [x] `go test ./pkg/controller/modelrouter/... -count=1`
- [x] `go test ./test/e2e/controller/... -run "^$" -count=1`
- [x] `go test ./test/e2e/framework/... -count=1`
- [ ] `go test -p 1 ./test/e2e/controller/modelrouter/... -v -count=1` (not run locally: `KUBECONFIG` is not configured; intended for the installed E2E cluster)

## Submission Checklist

- [x] PR title includes an appropriate prefix
- [x] Changes are clearly explained
- [x] Relevant static, unit, and compile checks pass
- [x] Code follows project style and best practices
- [x] Test documentation is updated
- [x] Commit includes a matching `Signed-off-by` trailer for DCO


test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+		"max_tokens": 1,
+	})
+	if err != nil {
+		t.Fatalf("marshal gateway request: %v", err)
+	}
+
+	var lastErr error
+	err = wait.PollUntilContextTimeout(ctx, 2*time.Second, modelRouterPollTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			_, lastErr = framework.SendPDRequest(ctx, h.config, "random", requestID, body)
```
varungup90: This does not validate the ModelRouter HTTPRoute.

`SendPDRequest` with `"random"` sets `routing-strategy: random`. The gateway plugin then selects a pod from its cache and sets `target-pod`; it never calls `validateHTTPRouteStatus`, and Envoy never uses this test's HTTPRoute `backendRefs`. A successful mock-pod hit only proves that the plugin discovered a uniquely labeled pod. It can pass even if the generated route is wrong or unaccepted.

Issue #2737 asks for a request through the gateway that reaches the backend **via the generated route**.

Suggested fix:
1. Wait until the HTTPRoute is `Accepted` and `ResolvedRefs` (see `waitForHTTPRouteReady` in `test/e2e/installation/smoke_test.go`).
2. Send with an empty routing strategy so the plugin stays on `RouterNotSet`, sets the `model` header, and Envoy matches the ModelRouter HTTPRoute to the test Service.
3. Keep the mock `request_id` assertion.

Passing `""` is enough: `deriveRoutingStrategyFromContext` ignores a blank `routing-strategy` header.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+			current, err := h.gatewayClient.GatewayV1().HTTPRoutes(modelRouterGatewayNamespace).
+				Get(ctx, utils.ModelRouterName(model), metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			if err != nil {
+				return false, err
+			}
+			route = current
+			return true, nil
```
varungup90: This returns as soon as the HTTPRoute object exists. `createHTTPRoute` uses `Create`, so the object appears immediately, before Envoy Gateway has accepted it or resolved the cross-namespace Service via the ReferenceGrant.

If the gateway probe is switched to the HTTPRoute path (`RouterNotSet`), waiting only for existence will race with dataplane programming. Please wait for parent conditions `Accepted` and `ResolvedRefs` to be true, matching `waitForHTTPRouteReady` in `test/e2e/installation/smoke_test.go`.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+		t.Fatalf("list controller pods: %v", err)
+	}
+	if len(pods.Items) == 0 {
+		t.Fatal("controller deployment has no pods to restart")
+	}
+	oldUIDs := make(map[string]struct{}, len(pods.Items))
+	for i := range pods.Items {
+		oldUIDs[string(pods.Items[i].UID)] = struct{}{}
+		if err := podsClient.Delete(ctx, pods.Items[i].Name, metav1.DeleteOptions{
+			GracePeriodSeconds: ptr.To[int64](0),
```
varungup90: Please drop `GracePeriodSeconds: 0`. The manager runs with `--leader-elect`; force-delete skips graceful lease release, so the replacement pod can become Ready while still waiting on the old lease.

Waiting for new pod UIDs / `AvailableReplicas` is not the same as “ModelRouter is reconciling again.” Use a graceful delete, then treat a post-restart create (the probe Deployment below) as the readiness gate rather than pod Ready.

This restart also runs in the default `controller` / `all` suites. `./test/e2e/controller/...` is alphabetical, so `roleset` and `stormservice` run next against a just-restarted manager.

test/e2e/controller/modelrouter/modelrouter_test.go
```diff
@@ -0,0 +1,139 @@
...
+	h.waitForRouteDeleted(t, ctx, adapterModel)
+	_ = h.waitForRoute(t, ctx, primaryModel)
+	_ = h.waitForReferenceGrant(t, ctx)
+
+	primarySpec := primaryRoute.Spec.DeepCopy()
+	h.restartController(t, ctx)
+	restartedRoute := h.waitForRoute(t, ctx, primaryModel)
+	assert.True(t, reflect.DeepEqual(primarySpec, restartedRoute.Spec.DeepCopy()),
+		"HTTPRoute spec changed across controller restart")
+	assert.Equal(t, 1, h.routeCount(t, ctx, primaryModel))
```
varungup90: This is a weak idempotency check. `createHTTPRoute` is create-if-not-exists (`AlreadyExists` is ignored), so comparing the pre-restart spec to the object still sitting in `aibrix-system` does not prove the controller re-applied anything. `routeCount == 1` is useful; spec equality after restart is not.

The probe Deployment on the next lines is the real liveness check. Create it as the restart barrier (after a graceful controller restart), then assert the primary route is still unique and the probe route appears.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+	return h
+}
+
+func (h *modelRouterHarness) cleanup(t *testing.T) {
+	t.Helper()
+	if t.Failed() {
+		h.logDiagnostics(t)
+		if strings.EqualFold(strings.TrimSpace(os.Getenv(modelRouterKeepOnFailureEnv)), "true") {
+			t.Logf("preserving ModelRouter E2E namespace %s after failure", h.namespace)
+			return
```
varungup90: When keep-on-failure is set, this returns before deleting HTTPRoutes in `aibrix-system`. The test namespace is unique; those routes are not. Preserving the namespace for debug is good, but leftover routes in the shared gateway namespace can accumulate across failed runs.

Please still delete (or explicitly document) the `aibrix-system` HTTPRoutes even when the namespace is retained.

varungup90 (changes requested):
## Review

The package shape is good: unique namespace, bounded polling, isolated mock backend, shared ReferenceGrant cleanup, and docs for the existing keep-on-failure setting. This should not change ModelRouter behavior, which is the right split from #2737 scope 2.

The main gap is that the live request does **not** exercise the generated HTTPRoute. `routing-strategy: random` makes the gateway plugin pick a pod and set `target-pod`, so the request can succeed even if the HTTPRoute is unaccepted, points at the wrong Service, or the ReferenceGrant is broken. That is the path this E2E is meant to lock in.

Please also tighten controller restart (graceful delete + wait until the controller is actually reconciling) and keep shared-namespace HTTPRoutes from leaking when resources are preserved on failure.

This package is in the default `controller` / `all` suites and has not been run against a cluster yet. After the request-path fix, please run:

```bash
go test -p 1 ./test/e2e/controller/modelrouter/... -v -count=1
```

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+	}
+	if len(pods.Items) == 0 {
+		t.Fatal("controller deployment has no pods to restart")
+	}
+	oldUIDs := make(map[string]struct{}, len(pods.Items))
+	for i := range pods.Items {
+		oldUIDs[string(pods.Items[i].UID)] = struct{}{}
+		if err := podsClient.Delete(ctx, pods.Items[i].Name, metav1.DeleteOptions{
+			GracePeriodSeconds: ptr.To[int64](0),
+		}); err != nil {
```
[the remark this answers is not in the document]
googs1025: Fixed in f7efe35d: controller pods now use a graceful delete, allowing leader-election lease release.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+	t.Helper()
+	var podName string
+	err := wait.PollUntilContextTimeout(ctx, modelRouterPollInterval, modelRouterPollTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			pods, err := h.kubeClient.CoreV1().Pods(h.namespace).List(ctx, metav1.ListOptions{
+				LabelSelector: "app=" + deployment,
+			})
+			if err != nil || len(pods.Items) != 1 {
+				return false, err
+			}
```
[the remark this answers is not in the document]
googs1025: Fixed in f7efe35d: pod readiness polling retries only recognized transient API and network errors; non-transient errors still fail immediately.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+	err := wait.PollUntilContextTimeout(ctx, modelRouterPollInterval, modelRouterPollTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			current, err := h.gatewayClient.GatewayV1().HTTPRoutes(modelRouterGatewayNamespace).
+				Get(ctx, utils.ModelRouterName(model), metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			if err != nil {
+				return false, err
+			}
```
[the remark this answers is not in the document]
googs1025: Fixed in f7efe35d: HTTPRoute polling retries recognized transient API and network errors and waits for readiness conditions.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+func (h *modelRouterHarness) waitForRouteDeleted(t *testing.T, ctx context.Context, model string) {
+	t.Helper()
+	err := wait.PollUntilContextTimeout(ctx, modelRouterPollInterval, modelRouterPollTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			_, err := h.gatewayClient.GatewayV1().HTTPRoutes(modelRouterGatewayNamespace).
+				Get(ctx, utils.ModelRouterName(model), metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
```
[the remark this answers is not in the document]
googs1025: Fixed in f7efe35d: HTTPRoute deletion polling tolerates recognized transient API and network errors.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+	err := wait.PollUntilContextTimeout(ctx, modelRouterPollInterval, modelRouterPollTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			current, err := h.gatewayClient.GatewayV1beta1().ReferenceGrants(h.namespace).
+				Get(ctx, name, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return false, nil
+			}
+			if err != nil {
+				return false, err
+			}
```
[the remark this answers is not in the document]
googs1025: Fixed in f7efe35d: ReferenceGrant polling retries recognized transient API and network errors.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+	t.Helper()
+	name := fmt.Sprintf("%s-reserved-referencegrant-in-%s", modelRouterGatewayNamespace, h.namespace)
+	err := wait.PollUntilContextTimeout(ctx, modelRouterPollInterval, modelRouterPollTimeout, true,
+		func(ctx context.Context) (bool, error) {
+			_, err := h.gatewayClient.GatewayV1beta1().ReferenceGrants(h.namespace).
+				Get(ctx, name, metav1.GetOptions{})
+			if apierrors.IsNotFound(err) {
+				return true, nil
+			}
+			return false, err
```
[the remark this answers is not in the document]
googs1025: Fixed in f7efe35d: ReferenceGrant deletion polling retries recognized transient API and network errors.

test/e2e/controller/modelrouter/helpers_test.go
```diff
@@ -0,0 +1,588 @@
...
+	if err != nil {
+		return
+	}
+	for i := range pods.Items {
+		logs, err := h.kubeClient.CoreV1().Pods(modelRouterGatewayNamespace).
+			GetLogs(pods.Items[i].Name, &corev1.PodLogOptions{TailLines: ptr.To[int64](200)}).
+			DoRaw(ctx)
+		if err == nil {
+			t.Logf("controller logs from %s:\n%s", pods.Items[i].Name, logs)
+		}
```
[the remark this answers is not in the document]
googs1025: Fixed in f7efe35d: diagnostics request logs from the manager container and report log retrieval failures.

test/e2e/controller/modelrouter/helpers_test.go
googs1025: Fixed in f7efe35d: the gateway request now uses an empty routing strategy, so RouterNotSet injects the model header and Envoy routes through the generated HTTPRoute. The request ID recorder assertion remains.

test/e2e/controller/modelrouter/helpers_test.go
googs1025: Fixed in f7efe35d: waitForRouteReady now requires Accepted=True and ResolvedRefs=True on the same parent before the request is sent.

test/e2e/controller/modelrouter/helpers_test.go
googs1025: Fixed in f7efe35d: restart uses graceful pod deletion. Replacement pod readiness is followed by creation of the probe Deployment; the accepted and resolved probe route is the reconciliation barrier.

test/e2e/controller/modelrouter/modelrouter_test.go
googs1025: Fixed in f7efe35d: removed the pre and post spec comparison. The post-restart probe route is now the liveness gate, followed by the primary-route uniqueness assertion.

test/e2e/controller/modelrouter/helpers_test.go
googs1025: Fixed in f7efe35d: cleanup always deletes the tracked aibrix-system HTTPRoutes first, even when failure preservation keeps the test namespace.

test/e2e/controller/modelrouter/helpers_test.go
googs1025: Follow-up in 3296e52e after the live CI run: an empty routing strategy did enter RouterNotSet, but repeated requests logged request_start without request_end and timed out. The probe now sends the model header on the ingress request and omits routing-strategy entirely, so the first Envoy route decision selects the generated header-specific HTTPRoute directly. A unit test verifies both headers.

googs1025: Second follow-up in 72ae6494 after the next live CI run: the direct HTTPRoute request returned successfully. The remaining recorder mismatch was caused by the prefixed non-UUID x-request-id being replaced by Envoy. The test now generates a canonical UUID so the same ID reaches the mock recorder.

googs1025: Third follow-up in 09266f97: the live run confirmed Envoy replaces external request IDs even when they are valid UUIDs. The test now embeds the UUID in the request body, locates the unique successful record by that immutable marker, and separately requires the recorder request ID to be non-empty.

varungup90 (approved):

PR #2746, opened 2026-09-18
[Gateway] Fail the client request and abort the decode leg when the SGLang PD prefill leg fails

### LINKED ISSUE #2289: SGLang PD: prefill failure is silent and leaves the decode request hanging on bootstrap
### Describe the bug

In disaggregated mode the SGLang prefill is fired asynchronously and its result is never checked, so a failed prefill is invisible to the request path and the decode request is dispatched anyway.

In `doPrefillRequest` (`pkg/plugins/gateway/algorithms/pd_prefill_request.go`), the SGLang branch runs the prefill HTTP call in a goroutine and only logs on error:

```go
case SGLangEngine:
    go func() {
        defer r.prefillRequestTracker.RemovePrefillRequest(routingCtx.RequestID)
        if _, err := r.executeHTTPRequest(apiURL, routingCtx, payload); err != nil {
            klog.ErrorS(err, "prefill_request_failed", ...)
            return // failure swallowed
        }
        ...
    }()
```

`Route()` does not wait for this goroutine — for SGLang `doPrefillRequest` returns `nil` immediately, then `Route()` sets the decode pod and returns. The decode request already carries the `bootstrap_room` generated in `preparePrefillPayload`. If the prefill 5xx'd, timed out, or never brought up its bootstrap server, the decode pod waits on a `bootstrap_room` rendezvous that never completes and only fails on SGLang's own bootstrap timeout. The client sees a long hang followed by an opaque decode-side error, with no gateway-side signal and no fast failure.

Two consequences:

- This is asymmetric with the vLLM path, which runs the prefill synchronously (`handleSyncPrefill`) and fails the request fast when the prefill errors.
- `Route()` emits `GatewayPrefillRequestSuccessTotal` for SGLang unconditionally, because the async call returns `nil` before the prefill has actually completed — so the success metric does not reflect prefill outcome.

### Steps to Reproduce

1. Disaggregated SGLang deployment (prefill + decode rolesets) with the `pd` router.
2. Make the selected prefill pod fail the prefill as a request arrives (e.g. kill/OOM it, or block its bootstrap port).
3. Send a request → it is not failed fast; the decode pod hangs until SGLang's bootstrap timeout and then returns an opaque error. The gateway logs `prefill_request_failed` but still counts prefill success and proceeds.

### Expected behavior

A failed SGLang prefill should fail (or retry) the request rather than dispatching a decode that will hang. Options:

- gate decode dispatch on the prefill bootstrap being acknowledged / the prefill request succeeding,
- retry on another prefill pod before giving up, and/or
- surface a fast, typed error to the client.

At minimum, prefill failure should be reflected in the request outcome and in metrics (`GatewayPrefillRequestSuccessTotal` should not be emitted for SGLang when the prefill has not been confirmed).

### Environment

AIBrix: main · SGLang prefill/decode (disaggregated, bootstrap mode).

qidaye:
## Pull Request Description

In SGLang PD disaggregation the gateway fires the prefill leg asynchronously and lets Envoy forward the decode leg. When the prefill leg fails (dead prefill pod, non-2xx, timeout, client cancel) nothing tells the decode pod or the client: the decode pod sits in its bootstrap / prealloc queue for that `bootstrap_room` until the engine's own bootstrap timeout (minutes), holding a decode slot, and the client sees a long hang followed by an opaque decode-side error. This is #2289.

This PR keeps the prefill dispatch asynchronous (the constraint raised in #2289) and only stops discarding its result.

### What changes

1. **`utils.LoadEnvNonNegativeInt`** — `LoadEnvInt` treats every value ≤ 0 as invalid and silently falls back, so a knob whose documented value `0` means "off" cannot be switched off from the environment. The new helper accepts 0 and still refuses negative/unparseable values. `LoadEnvInt` is unchanged for its callers.

2. **Gateway-owned `rid` + decode abort on prefill failure.** SGLang's `/abort_request` matches on `rid` only (`AbortReq` has no room field) and reads `rid` from the request body, not from `X-Request-Id`. The gateway now writes `rid = <request id>-<16-hex nonce>` into both legs' bodies (fixed width because SGLang's abort match is prefix based; the nonce keeps retries with a reused request id distinct). When the async prefill leg fails, the failure is classified (`http_status`, `timeout`, `canceled`, `transport`, `request_setup`, `bad_response`) and the gateway POSTs `{"rid": ...}` to the decode pod's `/abort_request`. The abort is best-effort on its own goroutine with its own timeout, sent twice (immediately and after a short delay) because a prefill pod that refuses the connection outright lets the first abort overtake the decode leg, and the tokenizer manager silently drops an abort for an unknown rid. It is skipped once the decode pod has started responding. `bad_response` (HTTP 200 with an unparseable body) never aborts because the transfer did complete.

3. **Fail the client request immediately.** `Route()` only produces the header mutation, but the same ext_proc stream stays open for the response and is parked in `srv.Recv()` waiting for decode response headers that will not arrive until the bootstrap timeout. `srv.Recv()` moves into one reader goroutine per stream and the processing loop becomes a `select` over the reader, the stream context and a per-request prefill-failure wakeup. Non-PD streams select on exactly the two cases they had before. When a terminal prefill failure lands before the decode pod has responded, the gateway answers with an `ImmediateResponse` (503 for timeout/transport/request_setup/canceled, the upstream status for `http_status`, 502 otherwise) carrying an OpenAI-shaped error body and `x-error-pd-prefill: true`, then closes the stream with a gRPC error so `failure_mode_allow: false` makes Envoy reset the decode connection. If the decode pod has already responded, the failure is only recorded. The per-request leg state lives in a per-incarnation object captured by the prefill goroutine before launch, so a late prefill failure can never land on a recycled `RoutingContext` that another request has since taken from the pool (there is a test seam that reproduces the aliasing deterministically).

4. **Observability.** `gateway_pd_prefill_failure_total{class,stage}` (`stage` = `before_response` / `after_response`), `gateway_pd_decode_abort_total{prefill_failure_class,result}`, structured log lines `pd_prefill_fail_fast` and `pd_decode_abort`. The generic request-fail counters carry `status="pd_prefill_failure"` so a fail-fast is separable from an upstream 503. A fail-fast runs `finalizeRouterContext` exactly once and is never counted as a successful request.

5. **Env vars** (documented in `pkg/plugins/gateway/ENV_VARS.md`): `AIBRIX_PREFILL_REQUEST_TIMEOUT` (existing), `AIBRIX_DECODE_ABORT_TIMEOUT` (seconds, default 3, `0` disables the abort), `AIBRIX_DECODE_ABORT_RETRY_DELAY` (seconds, default 2, `0` sends a single attempt).

6. **Tests.** Unit tests for the failure classification, the abort scheduling (both attempts, skip-after-response, `0` semantics), the `select` and the fail-fast response, plus a regression test for the pooled-context aliasing. e2e (`test/e2e/gateway/pd/pd_fail_fast_test.go`): `TestPDPrefillFailFastRejectedPrefill` rejects the prefill leg with the mock while holding the decode leg for 10 s (new role-scoped `x-aibrix-mock-delay-role` header on the mock app) and asserts the client is answered with the prefill status + `x-error-pd-prefill: true` before the decode leg could respond, and that a decode mock recorded `/abort_request` with the gateway's rid; `TestPDPrefillFailFastPrefillTimeout` covers the `timeout` class / 503 mapping and skips itself unless the gateway is deployed with a short `AIBRIX_PREFILL_REQUEST_TIMEOUT` (exported as `AIBRIX_E2E_PREFILL_REQUEST_TIMEOUT_SECONDS`), because the shipped 60 s exceeds the mock's delay cap. The mock app gains an `/abort_request` route so the abort is observable.

### Notes for review

- Commits are separable: (1) `LoadEnvNonNegativeInt`, (2) rid + decode abort, (3) client fail-fast, (4) e2e + docs. Each builds and vets on its own.
- `http_status` passes the prefill engine's own 4xx/5xx through to the client (502 if it is not a code Envoy names); the other classes map to 503. Happy to switch to "always 503 with the upstream status in the body" if preferred.
- `AIBRIX_DECODE_ABORT_TIMEOUT=0` is the off switch for the abort. A dedicated boolean would work as well; this keeps the knob count down.
- `TestPDFailureSGLangAsyncPrefillHandoff` now normally sees the gateway's fail-fast response before the decode mock's own 500; its `>= 500` assertion still holds and the decode mock still records its handoff failure. Flagging it in case it flakes on CI.
- Race detector is clean on `pkg/plugins/gateway`, `pkg/plugins/gateway/algorithms/pd/...` and `pkg/types`; `make lint` and `make licensecheck` pass.

### Verification on a real deployment

Fault injection on a 2-prefill / 2-decode SGLang deployment with a gateway built from the equivalent change, prefill request timeout set to 3 s so that a cold ~24k-token prompt times out on the prefill leg:

| scenario | client | decode pod | prefill pod |
|---|---|---|---|
| prefill timeout, abort enabled | 6/6 got 503 at 3.01 s (before: hang until the engine's 300–600 s bootstrap timeout) | `AbortReq` dequeued + ACKed 6/6, prealloc/transfer queues back to baseline | abort notification in the same second, 6/6 released |
| prefill timeout, `AIBRIX_DECODE_ABORT_TIMEOUT=0` | same 503 at 3.01 s | released by the Envoy reset alone ≈1.1 s after the 503 | ≈1.1 s after the 503 |
| `kill -9` a prefill pod under load | requests on the dead pod: 503 in 2.2 s (`transport`); requests on the healthy pod: 200 | abort ACKed 4/4 | n/a |

No collateral under normal load (≈1,800 requests, ≈65k-token prompts, no injected failures): every abort attempt logged was the "client closed the stream after completion" echo, all after the request's own `request_end` (Δ 0.1–0.6 ms), all skipped because the decode pod had already responded; the engines' `num_aborted_requests_total` did not move on any pod. Routing overhead: `routing_time` p50 5.4 ms, p99 81 → 185 ms.

## Related Issues

Resolves: #2289

<details>

<h3>Submission Checklist</h3>
<ul>
    <li>[x] Documentation updated to reflect changes (if applicable)</li>
</ul>

</details>


pkg/plugins/gateway/gateway.go
```diff
@@ -373,26 +389,58 @@ func (s *Server) processOnce(srv extProcPb.ExternalProcessor_ProcessServer, st *
...
 	select {
-	case r := <-ch:
+	case r := <-st.recvCh:
+		st.recvCh = nil
 		if r.err != nil {
 			return s.handleRecvError(st, r.err)
 		}
 		req = r.req
+	case <-prefillFailed:
+		st.prefillFailFastDone = true
```
varungup90, line 430: **Bug: `prefillFailFastDone` is set before the failure is actually handled.**

If this `select` fires on `prefillFailed` and a buffered `Recv` then wins the inner `select`, we permanently disarm the wakeup and never call `handlePrefillFailFast`. Any later terminal prefill failure is ignored.

With the shipped Envoy config (`request_trailer_mode: SKIP`) the next message is usually decode response headers, so this may not hang today — but it does skip fail-fast and drops the `after_response` metric on that tie. If request trailers are enabled, or any other non-response ext_proc message lands in that window, the client waits out the bootstrap timeout again (#2289).

Please only set `st.prefillFailFastDone = true` on the `default` branch that calls `handlePrefillFailFast`. If a Recv wins, leave the wakeup armed. After response headers, the next iteration will fire fail-fast, see `DecodeResponded()`, record `after_response`, and return nil without spinning.

Please also add a unit test where a non-response message is already in `recvCh` when the prefill channel closes, and assert the client still gets an `ImmediateResponse`.

Suggested shape:

```go
case <-prefillFailed:
    // A message that has already arrived wins the tie.
    select {
    case r := <-st.recvCh:
        st.recvCh = nil
        if r.err != nil {
            return s.handleRecvError(st, r.err)
        }
        req = r.req
    default:
        st.prefillFailFastDone = true
        return s.handlePrefillFailFast(srv, st)
    }
```

pkg/plugins/gateway/algorithms/pd/abort.go
```diff
@@ -0,0 +1,478 @@
...
+			if leg.DecodeResponded() {
+				logAbort(attempt, abortResultSkippedStreaming, 0, nil)
+				return
+			}
+			if abortErr := postDecodeAbort(client, decodeAddr, rid, timeout); abortErr != nil {
+				logAbort(attempt, abortResultError, time.Since(start), abortErr)
+				return
+			}
+			logAbort(attempt, abortResultOK, time.Since(start), nil)
+		}
```
varungup90, line 420: **In-flight `/abort_request` is not cancelled when decode starts responding.**

Re-checking `leg.DecodeResponded()` before each POST is the right idea, but `postDecodeAbort` uses its own timeout context that nothing cancels. The window this skip exists to close is still open:

1. Prefill times out after KV has actually landed
2. Abort POST is already on the wire
3. Decode response headers arrive → `MarkDecodeResponded()`
4. Abort still hits the pod and kills a generation the client has started reading

Timeout is the class where "prefill may be in any state"; this is that case.

Please give the abort goroutine a context stored on `PDLegState`, cancel it from `MarkDecodeResponded()`, and pass that context into `postDecodeAbort` so an in-flight POST is abandoned as soon as decode is live. Keep the pre-POST check.

A good regression is to extend `TestAbortRetrySkippedWhenDecodeStartsStreaming` so headers arrive *during* the POST, not only between attempts.

varungup90 (changes requested):
Thanks for the thorough write-up and the separable commits. The direction is right: keep the SGLang prefill dispatch async, stop discarding the failure, own the engine `rid`, abort the decode leg, and fail the client instead of waiting out bootstrap.

Please address the two items below before merge. The first one can silently disable fail-fast; the second can still kill a generation after decode has started responding.

The per-incarnation `PDLegState` (captured before the prefill goroutine, replaced on `reset()`), rid injection into both legs, `bad_response` not aborting, and the `ImmediateResponse` + `codes.Aborted` pairing look sound.

pkg/plugins/gateway/gateway.go
qidaye: Fixed in the updated `feat(gateway): fail the client request …` commit. `prefillFailFastDone` is now set only in the `default` branch, right before `handlePrefillFailFast`; when a buffered Recv wins the tie the wakeup stays armed, and the next iteration fires fail-fast (recording `after_response` if response headers were among the buffered messages). New unit test `TestProcessFailsFastAfterBufferedMessageWinsTheTie` seeds `recvCh` with a non-response message before closing the prefill channel and asserts the client still gets the `ImmediateResponse` on the following iteration; it fails against the previous code with "the wakeup must stay armed until the failure has actually been handled".

pkg/plugins/gateway/algorithms/pd/abort.go
qidaye: Fixed in the updated `feat(gateway): abort the decode leg …` commit. `PDLegState` now owns the abort context (`AbortContext()`, built lazily per incarnation, cancelled by `MarkDecodeResponded()`, and cancelled immediately if decode had already responded by the time it is built). `postDecodeAbort` layers the per-attempt timeout on that context, so a POST already on the wire is abandoned as soon as decode headers arrive; the pre-POST `DecodeResponded()` check stays, and the wait between the two attempts is now a timer/ctx select instead of `time.Sleep`. A POST cut short by that cancellation is recorded as `skipped_streaming`, not `error`. New `TestAbortCancelledWhenDecodeStartsStreamingMidFlight` blocks the decode handler until `MarkDecodeResponded()` is called mid-POST and asserts the request context was cancelled, the outcome is `skipped_streaming` and no second attempt is sent; `TestAbortRetrySkippedWhenDecodeStartsStreaming` now uses a 30 s retry delay to prove the wait is interruptible.

The same commit also fixes the CI race job: the existing `TestPDRouter_RouteRecordsAsyncPrefillFailureWithoutSuccess` returned while the abort goroutine was still emitting its metric, racing the deferred restore of the test counter hook. `PDLegState.AbortDone()` is closed when the abort handling of an incarnation finishes (goroutine exit, or a skip decision), and that test plus our own abort tests join on it. `go test -race` over `./pkg/plugins/gateway/... ./pkg/types/... ./pkg/utils/...` is clean.

pkg/plugins/gateway/gateway.go
```diff
@@ -373,26 +389,58 @@ func (s *Server) processOnce(srv extProcPb.ExternalProcessor_ProcessServer, st *
...
+		select {
+		case r := <-st.recvCh:
+			st.recvCh = nil
+			if r.err != nil {
+				return s.handleRecvError(st, r.err)
+			}
+			req = r.req
+		default:
+			return s.handlePrefillFailFast(srv, st)
+		}
```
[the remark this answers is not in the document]
qidaye: Addressed together with the maintainer comment above: the flag is now set only in the `default` branch, with a regression test.

qidaye:
Pushed 41af6a01 addressing both review items and the race-job failure:

1. `prefillFailFastDone` is set only where the failure is handled; a buffered Recv that wins the tie leaves the wakeup armed. Regression test `TestProcessFailsFastAfterBufferedMessageWinsTheTie`.
2. The abort goroutine runs on a context owned by `PDLegState`, cancelled from `MarkDecodeResponded()`, so an in-flight `/abort_request` is abandoned once decode is live (recorded as `skipped_streaming`); the retry wait is interruptible. Regression test `TestAbortCancelledWhenDecodeStartsStreamingMidFlight`.
3. `Go Race Tests` failed because the existing `TestPDRouter_RouteRecordsAsyncPrefillFailureWithoutSuccess` returned before the abort goroutine emitted its metric. The leg state now exposes `AbortDone()` and the affected tests join on it.

`go build`, `go vet`, `make lint`, `make licensecheck`, gofmt and `go test -race` over `./pkg/plugins/gateway/... ./pkg/types/... ./pkg/utils/...` are clean locally. @varungup90 PTAL.

varungup90 (approved):

PR #2752, opened 2026-09-19
[Misc] Add unit tests for the RoundRobin dispatch policy

stywzn:
## What this PR does

`python/aibrix/aibrix/batch/client/` had no tests at all. This adds the first ones, for the `RoundRobin` dispatch policy in `policy.py`.

## What the tests cover

Three specs against `RoundRobin.pick`:

1. **No reachable channels** — returns `None`.
2. **A single channel** — keeps returning that channel across repeated calls, which is the gateway / ClusterIP / LB case the class docstring calls out as degenerating to identity.
3. **Several channels** — successive picks walk the list in order and the fourth wraps back to the first, so the cursor is shown to be taken modulo the live channel count rather than growing without bound.

The specs pass plain strings as stand-in channels: `pick` only reads `len(channels)` and indexes into the list, so constructing real `Channel` objects would add setup without adding coverage.

## Scope

Test-only; no production code is touched. `FIFO.schedule` is left for a follow-up, since it is an async generator and wants `pytest-asyncio` coverage of its own.

## Test plan

```
pytest tests/batch/client/test_policy.py -v
ruff check tests/batch/client/test_policy.py
```

Both pass.


stywzn:
Both points applied in 72c882ab. One correction on the first, since I checked it before changing anything.

**Type safety.** The annotations were indeed being violated, but CI would not have failed: this repo runs `mypy .` with only `ignore_missing_imports` set, and mypy does not check the bodies of unannotated functions by default. These specs carry no annotations, so `mypy` reports `Success` on the file either way. Running it with `--check-untyped-defs` is what surfaces the problem — 15 errors on the original file.

Worth noting that `MagicMock` alone does not clear them. With `--check-untyped-defs` the suggested form still reports:

```
Argument 2 to "pick" of "RoundRobin" has incompatible type "list[MagicMock]"; expected "list[Channel]"
note: "List" is invariant -- Consider using "Sequence" instead, which is covariant
```

A list literal passed inline is inferred from context and is fine, but one bound to a variable is inferred as `list[MagicMock]`, and `list` is invariant. Annotating that one variable as `list[Channel]` takes the file to zero errors under `--check-untyped-defs`, so that is what I did.

**Dynamic channel coverage.** Agreed, and this was the more valuable half — the docstring promises the cursor tolerates the reachable set growing and shrinking, and nothing covered it. Added as its own spec, offering lists of length 3, 2 and 1 across three calls.

One note on the expectations there: my first instinct for checking the spec had teeth was to vary the list sizes, but that is not a real check — after the first call the cursor is 1, and `1 % 2` and `1 % 3` are both 1, so the assertion holds either way. Breaking the source instead (`index = 0` in `pick`) turns both rotation specs red, which is the evidence I wanted.

PR #2755, opened 2026-09-19
[Bug] Skip profile cache updates when Redis client is nil in standalo…

### LINKED ISSUE #2754: [Bug] Skip profile cache updates when Redis client is nil in standalone mode

ijayer:
The gateway-plugin panics with a nil pointer dereference **~10 seconds after startup** in standalone mode without Redis, which contradicts the documented behavior
  (`deployment/local/README.md`: *"Redis is not required in local mode. Without Redis, rate limiting is disabled but routing works normally."*).

  **Panic:**

  ```
  panic: runtime error: invalid memory address or nil pointer dereference

  goroutine 114 [running]:
  github.com/vllm-project/aibrix/pkg/cache.(*Store).updateDeploymentProfiles(...)
      pkg/cache/cache_profile.go:36
  github.com/vllm-project/aibrix/pkg/cache.initProfileCache.func1()
      pkg/cache/cache_init.go:553
  created by github.com/vllm-project/aibrix/pkg.cache.initProfileCache in goroutine 1
      pkg/cache/cache_init.go:548
  ```

  **Root cause:** `initProfileCache` unconditionally starts the 10s profile-refresh ticker, and on the first tick `updateDeploymentProfiles` calls `c.redisClient.Scan(...)` at
  `pkg/cache/cache_profile.go:36` without a nil check. In standalone mode with no Redis, `main()` deliberately continues with a nil `RedisClient` (warning: *"Running without
  Redis: rate limiting and user auth disabled"*), and `InitOptions.RedisClient` is documented as "Can be nil" — every sibling initializer in `cache_init.go` guards with `if
  opts.RedisClient != nil` (lines 420/425/432) except this one. Regression introduced in #1192 (`cf64e51b`).

  **Fix:** Skip the background refresh loop in `initProfileCache` when `store.redisClient` is nil, logging a warning. Two deliberate choices:

  1. `store.pendingLoadProvider = newPendingLoadProvider(store)` stays **unconditional** — it is consumed by the request-tracing path (`cache_trace.go:355`) independently of
  Redis and must not be disabled by this fix.
  2. The guard is placed inside `initProfileCache` (not at the call site) so the semantics of `enableProfileCaching` ("is the profile feature wanted") remain decoupled from
  Redis availability. Consumers already degrade gracefully when profiles are absent: `GetModelProfileByPod` returns `MissingProfileError`, and `cache_trace.go:372` explicitly
  treats it as non-fatal.

  **Workaround (verified on unpatched binary):** setting `AIBRIX_MODEL_GPU_PROFILE_CACHING_FLAG=false` also avoids the panic, confirming the profile-refresh path is the only
  crash source.

  **Testing:**

  - `go build ./pkg/cache/` and `go vet ./pkg/cache/` pass; `go test ./pkg/cache/` passes.
  - End-to-end on macOS (built `gateway-plugins` with `-tags "nozmq"` from `main` @ `abb2cd20`): standalone mode + `--endpoints-config` + **no Redis running** → unpatched binary
   crashes after the first 10s tick; patched binary stays alive and logs `"Redis client is nil, skipping deployment profile cache updates"` (observed process alive >13s).

  **Open question for maintainers (out of scope here):** in Kubernetes mode, a misconfigured Redis now silently disables profile-based features. A fail-fast at startup (similar
  to `EnableKVSync`'s `klog.Fatalf` at `cache_init.go:439`) could make such misconfiguration visible, but that changes existing behavior and likely deserves its own discussion.

  ## Related Issues

  Resolves: #2754

  ---

  - [x] Documentation updated to reflect changes (if applicable) — N/A, code behavior now matches existing docs


googs1025 (approved):
LGTM

varungup90 (approved):

